PR #4, opened 2015-02-27
MultiChannelShading plugin: converted prefs to profiles

nicost:


nicost:
Did that all look sane and how it is supposed to be done?

ChrisWeisiger:
Yes, that looked reasonable to me.

By the way, one side-effect of the changes I pulled into the primary
repository is that classes that descend from MMDialog or MMFrame can't use
the getPrefsNode() method, and consequently won't build. This method was
used to "share" a preferences node across classes, and is thus obsolete in
the new Profile-driven system. I got all of the uses in the main Java
layer, but missed the plugins. Once I finish with the Java layer migration,
I can start helping with plugins.

-Chris

PR #8, opened 2015-02-28
Mm2

nicost:


ChrisWeisiger:
Looks good. Do we have any other references to GUIUtils.recallPosition? If not, we should remove that method (and the stuff associated with it), since descending from MMFrame/MMDialog is strictly preferable.

Unfortunate that we need the Double[] unboxing / double[] boxing methods; too bad Java doesn't do that for us.

nicost:
> Looks good. Do we have any other references to GUIUtils.recallPosition? If not, we should remove that method (and the stuff associated with it), since descending from MMFrame/MMDialog is strictly preferable.
> 
Good idea.  I took them out.  Only place where they still appears to be in use was in the Projector plugin.  All gone now.

Best,

Nico

PR #9, opened 2015-02-28
Mm2

nicost:


ChrisWeisiger:
Two minor comments:

 * I suspect that setting the version string here is only sufficient to change how it appears in the program; it won't affect e.g. the application version as output by the builder. 

 * ProjectorPlugin.java includes a commented-out call to GUIUtils.recallPosition(), which should be deleted instead.

PR #12, opened 2015-03-03
Mm2

nicost:


ChrisWeisiger:
Any time you find yourself doing "getMaxIndex(axis) + 1" (e.g. in IntelligentAcquisitionFrame.java), you should use "getAxisLength(axis)" instead. I added the latter call specifically because I had millions of the former call in the display code at one point, and my inability to consistently remember to add the 1 had created lots of off-by-one errors.

ShadingTableModel.java has a static string CHANNELS that doesn't appear to be used anywhere. *shrug*

Otherwise looks good, and thanks for fixing up the plugins!

PR #16, opened 2015-03-04
Mm2

nicost:


ChrisWeisiger:
Oh, your emails are talking about closing the Datastore. Sorry for being dense. I should better-document the interaction between Datastore.close() and DisplayManager.track(). Basically, when you ask Micro-Manager to track a Datastore, it will automatically call Datastore.close() when the last display for that Datastore is closed. If you aren't using DisplayManager.track(), though, then Micro-Manager itself won't keep any references to the Datastore around except in the displays themselves, so once all displays are closed, the only references to the Datastore are in your code. Thus you can fuctionally "close" the Datastore simply by removing any references to it that you have.

I think for SplitView, using track() is not the right solution. track() is mostly meant for when a plugin creates some data and then wants to declaim responsibility for the datastore/viewers -- for example, a plugin that generates a Z-projection, but doesn't want to have to worry about prompting the user to save the data when the display is closed. The plugin can create a Datastore, create a display for it, then say "Hey, Micro-Manager, you take care of this", and forget about both of them. This isn't what SplitView is doing.

Clearly the documentation around this needs to be handled better. It's entirely possible that the close() method shouldn't even be exposed in the API, since it's only needed to tell Micro-Manager to clean up a tracked Datastore.

nicost:
> I think for SplitView, using track() is not the right solution. track() is mostly meant for when a plugin creates some data and then wants to declaim responsibility for the datastore/viewers -- for example, a plugin that generates a Z-projection, but doesn't want to have to worry about prompting the user to save the data when the display is closed. The plugin can create a Datastore, create a display for it, then say "Hey, Micro-Manager, you take care of this", and forget about both of them. This isn't what SplitView is doing.

Actually, that is exactly what Splitview is doing (except, it does not want to prompt the user to save the data, is there a flag for that too?).  

Nevertheless, “track()” currently does not work as it immediately closes a datastore when executing this sequence:

      store_ = gui_.data().createRAMDatastore();
      DisplayWindow display = gui_.display().createDisplay(store_);
      gui_.display().track(store_);

The viewer opens and immediately closes again…..

Nico

ChrisWeisiger:
On Wed, Mar 4, 2015 at 9:20 AM, nicost <notifications@github.com> wrote:

> > I think for SplitView, using track() is not the right solution. track()
> is mostly meant for when a plugin creates some data and then wants to
> declaim responsibility for the datastore/viewers -- for example, a plugin
> that generates a Z-projection, but doesn't want to have to worry about
> prompting the user to save the data when the display is closed. The plugin
> can create a Datastore, create a display for it, then say "Hey,
> Micro-Manager, you take care of this", and forget about both of them. This
> isn't what SplitView is doing.
>
> Actually, that is exactly what Splitview is doing (except, it does not
> want to prompt the user to save the data, is there a flag for that too?).
>

I didn't explain my point well enough. When you call track(), you are
implicitly stating that you will never interact with that datastore or its
displays again, and that you don't care at all what happens to them.
SplitView clearly *does* care because it has to be able to insert new
images into the datastore when its snap/live functions are invoked, so it
should not be using track().


>
> Nevertheless, “track()” currently does not work as it immediately closes a
> datastore when executing this sequence:
>
> store_ = gui_.data().createRAMDatastore();
> DisplayWindow display = gui_.display().createDisplay(store_);
> gui_.display().track(store_);
>
> The viewer opens and immediately closes again…..
>

I can't reproduce this. Firstly in that if you create a display for a
datastore that has no images in it, then the display window is not actually
generated (DisplayWindow waits until there is at least one image to display
before it does anything). Second in that even if I do add an image, the
display doesn't vanish. Here's my script in its entirety:

store = mm.data().createRAMDatastore();
display = mm.displays().createDisplay(store);
mm.getCore().snapImage();
store.putImage(mm.data().convertTaggedImage(mm.getCore().getTaggedImage()));
mm.displays.track(store);

-Chris

nicost:
> 
> >
> > Nevertheless, “track()” currently does not work as it immediately closes a
> > datastore when executing this sequence:
> >
> > store_ = gui_.data().createRAMDatastore();
> > DisplayWindow display = gui_.display().createDisplay(store_);
> > gui_.display().track(store_);
> >
> > The viewer opens and immediately closes again…..
> >
> 
> I can't reproduce this. Firstly in that if you create a display for a
> datastore that has no images in it, then the display window is not actually
> generated (DisplayWindow waits until there is at least one image to display
> before it does anything). Second in that even if I do add an image, the
> display doesn't vanish. Here's my script in its entirety:
> 
> store = mm.data().createRAMDatastore();
> display = mm.displays().createDisplay(store);
> mm.getCore().snapImage();
> store.putImage(mm.data().convertTaggedImage(mm.getCore().getTaggedImage()));
> mm.displays.track(store);

Change that script to:

store = mm.data().createRAMDatastore();
display = mm.displays().createDisplay(store);
mm.displays.track(store);
mm.getCore().snapImage();
store.putImage(mm.data().convertTaggedImage(mm.getCore().getTaggedImage()));

and you will likely see the same behavior that I have been seeing.

It is quite complex to ensure that the call to track() is only made after an image is put into the store.  There is also nothing explicit why one should first put an image into the store before calling track().  

However, it sounds like you do not want the track() method to be used by external code at all.  Maybe it should be removed form the api and be replaced with an event that signals closing of the last viewer?  Even then, it would be nice to have methods to close the datastore properly, with the option to prompt the user whether or not to save the data.

Best,

Nico

ChrisWeisiger:
On Wed, Mar 4, 2015 at 9:35 AM, nicost <notifications@github.com> wrote:

>
> Change that script to:
>
> store = mm.data().createRAMDatastore();
> display = mm.displays().createDisplay(store);
> mm.displays.track(store);
> mm.getCore().snapImage();
>
> store.putImage(mm.data().convertTaggedImage(mm.getCore().getTaggedImage()));
>
> and you will likely see the same behavior that I have been seeing.
>

Nope; still works fine for me.


>
> It is quite complex to ensure that the call to track() is only made after
> an image is put into the store. There is also nothing explicit why one
> should first put an image into the store before calling track().
>

You shouldn't need to.


>
> However, it sounds like you do not want the track() method to be used by
> external code at all. Maybe it should be removed form the api and be
> replaced with an event that signals closing of the last viewer? Even then,
> it would be nice to have methods to close the datastore properly, with the
> option to prompt the user whether or not to save the data.
>

I think we'd better talk in person so we can clearly hash out what exactly
track() does, because there's clearly some confusion here. It's a necessary
part of the API, but it doesn't make sense for SplitView to use it. We can
talk about it when I stop by in a bit.

-Chris

nicost:
> On Mar 4, 2015, at 9:50 AM, ChrisWeisiger <notifications@github.com> wrote:
> 
> On Wed, Mar 4, 2015 at 9:35 AM, nicost <notifications@github.com> wrote:
> 
> >
> > Change that script to:
> >
> > store = mm.data().createRAMDatastore();
> > display = mm.displays().createDisplay(store);
> > mm.displays.track(store);
> > mm.getCore().snapImage();
> >
> > store.putImage(mm.data().convertTaggedImage(mm.getCore().getTaggedImage()));
> >
> > and you will likely see the same behavior that I have been seeing.
> >
> 
> Nope; still works fine for me.

Cool.  Works on my end now as well (guess something changed since yesterday, or this only manifests itself on my laptop).

> > However, it sounds like you do not want the track() method to be used by
> > external code at all. Maybe it should be removed form the api and be
> > replaced with an event that signals closing of the last viewer? Even then,
> > it would be nice to have methods to close the datastore properly, with the
> > option to prompt the user whether or not to save the data.
> >
> 
> I think we'd better talk in person so we can clearly hash out what exactly
> track() does, because there's clearly some confusion here. It's a necessary
> part of the API, but it doesn't make sense for SplitView to use it. We can
> talk about it when I stop by in a bit.

Actually, I am using it in the “Live Replay” plugin

PR #23, opened 2015-03-07
Mm2

nicost:
Fixes a serious issue with MMFrame.  Also makes a couple of plugins run with current api of v2.0.


ChrisWeisiger:
Thanks for catching that MMFrame issue!

PR #33, opened 2015-05-13
mmstudio: Added Javadoc, made variables in API final whenever possibl…

nicost:
…e and removed superfluous imports in API


ChrisWeisiger:
Okay, this is in now. Thanks, Nico!

PR #66, opened 2015-09-24
Fix typo 'acquistion' in comments and messages

marktsuchida:


ChrisWeisiger:
Wow, how did I lose sight of this?

Issue #110, opened 2015-11-25
Plugins in MMJ_.jar get loaded multiple times when running from NetBeans


marktsuchida:
Originally reported here: http://micro-manager.3463995.n2.nabble.com/Plugin-is-loading-multiple-times-tp7586337.html
"It looks like every plugin is being loaded once for every jar in the mmplugins directory (and every jar in the mmautofocus directory)." (as summarized by @ChrisWeisiger ).

ChrisWeisiger:
This appears to be down to something to do with class loaders and/or SciJava's plugin discovery mechanism, as we are somehow discovering jars that should be invisible to the current class loader. It's possible that SciJava is doing something unusual to enable discovering plugins when run from within NetBeans (and/or Eclipse). It's also possible that NetBeans is doing something unusual to enable itself to work at all, which interacts poorly with our runtime plugin discovery. Suggest next course of action is to contact the SciJava folks and see if they have any ideas.

marktsuchida:
* SciJava has special code for Eclipse support, but I have checked that it doesn't get executed when running from NetBeans.
* We call `org.scijava.plugin.PluginIndex.discover()`, which in turn calls `org.scijava.plugin.DefaultPluginFinder.findPlugins()`. The latter method uses `org.scijava.annotations.Index<org.scijava.plugin.Plugin>` to iterate over all `META-INF/json/org.scijava.plugin.Plugin` resources (and entries therein) for the given class loader (in this case our `org.micromanager.internal.pluginmanagement.PluginFinder$PluginLoader` instance).
* That enumeration invokes the class loader's `getResources(String)` method, which is implemented in the ancestral `ClassLoader` class. `ClassLoader.getResources()` returns the combined results from the parent's `getResource()` and the current class's `findResource()` methods.
* According to `ClassLoader` javadoc, it is intended that subclass override the `findClass()`, `findResource()`, and `findResources()` methods, rather than `loadClass()`, `getResource()`, and `getResources()`.
* The reason `MMJ_.jar`'s `META-INF` gets searched is because all resources of the matching name in the ancestral class loaders are being enumerated (not sure why this doesn't happen outside of NetBeans).
* So a naïve fix would be to override `getResources()` in `org.micromanager.internal.pluginmanagement.PluginFinder$PluginLoader` to not look at the parent class loader's resources. But that, of course, is incorrect because we need plugins to be able to load and enumerate their regular resources.
[...]

PR #111, opened 2015-11-25
Prevent repeated loading of plugins contained in MMJ_.jar

marktsuchida:
Fixes issue #110 and clarifies the behavior of the class loader used by
plugins (once plugin discovery is finished, it is just a URLClassLoader).


ChrisWeisiger:
My only request is a comment indicating why the logic around `blockInheritedResources_` works the way it does (i.e. why do we call `findResource()` instead of `getResource()`?). Otherwise, looks good, and nice simplification of the class loader.

marktsuchida:
@ChrisWeisiger Not sure what you mean - what part is not obvious from the `ClassLoader` javadoc (and the name `blockInheritedResources_`)?

ChrisWeisiger:
As a personal preference, I like code that doesn't require me to resort to the javadocs, when reasonably possible. I mean, as the author of the code you know what the difference is between `findResource()` and `getResource()`, but anyone else who reads the code will need to spend the extra time to work that out. Or you could just tell them, in a comment. Yes the comment is redundant with existing documentation, but the comment is also much more convenient.

PR #116, opened 2015-12-02
mm2: Fix JavaDoc and make fields final that can be final in display api

nicost:


ChrisWeisiger:
I manually merged these in. Thanks for the contributions, Nico!

PR #177, opened 2016-02-09
diSPIM plugin port to Micro-Manager 2.0

nicost:
This PR should only affect the ASIdiSPIM plugin.  It consists of code very recently copied from the 1.4 branch and changed to work with the 2.0 api.  Not everything is functional yet, but there does not seem to be much point to keeping it in a private branch anymore.  I plan on continue to fix things for the next few days.


marktsuchida:
Merging this as it is would introduce ASIdiSPIM.jar into the build, which we shouldn't do if it is not yet ready. I will work to merge a version excluding the change to `Makefile.am` (hopefully also simplifying the history).

nicost:
I don't think that building the diSPIM plugin is not a problem, since it is functional (although not everything works as expected yet, but that is something that can be said about most software code;)

marktsuchida:
Sounds good. Looks like your `build.xml` got removed (probably at one of the merges from 'mm2'), so resurrected that on your behalf. #188.

PR #210, opened 2016-02-19
mm2:  Upgraded OughtaFocus plugin to use math commons 3 rather than 2 library

nicost:
There should be no functional consequences and this should help towards getting rid of dependency on (the very outdated) commons math2 library.

### DIFF
--- a/autofocus/OughtaFocus.java
+++ b/autofocus/OughtaFocus.java
@@ -1,13 +1,16 @@
 ///////////////////////////////////////////////////////////////////////////////
 //FILE:           OughtaFocus.java
 //PROJECT:        Micro-Manager
-//SUBSYSTEM:      Autofocusing plug-in for mciro-manager and ImageJ
+//SUBSYSTEM:      Autofocusing plug-in for micro-manager and ImageJ
 //-----------------------------------------------------------------------------
 //
 //AUTHOR:         Arthur Edelstein, October 2010
 //                Based on SimpleAutofocus by Karl Hoover
 //                and the Autofocus "H&P" plugin
 //                by Pakpoom Subsoontorn & Hernan Garcia
+//                Contributions by Jon Daniels (ASI): FFTBandpass and MedianEdges
+//                Chris Weisiger: 2.0 port
+//                Nico Stuurman: 2.0 port and Math3 port
 //
 //COPYRIGHT:      University of California San Francisco
 //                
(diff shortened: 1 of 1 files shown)


ChrisWeisiger:
Would it make sense to add "minimum Z step size" to the parameters for configuring OughtaFocus?

In any case, merging.

marktsuchida:
@ChrisWeisiger The minimum Z step size _is_ the Tolerance_um parameter.

@nicost The relative tolerance _rel_ is a factor, not an absolute value, so 0.05 is not 50 nm. If I understand correctly, setting _rel_ to 0.05 and _abs_ to (say) 0.1 µm means that, if the result of the optimization is 56.0 µm, the tolerance is approximately 0.05 x 56.0 + 0.1 = 2.9 µm (probably not what was intended).

Forgetting Z stages for a moment, the reason `BrentOptimizer` needs both a relative and absolute tolerance is that it is designed for use with arbitrary floating-point numbers. As you know, a given floating point type (e.g. `double`) always has a certain relative precision, but larger values have less absolute precision (e.g. `1.0e100 + 1.0 == 1.0e100`).

So if the result of the optimization is a large FP number, having just an absolute tolerance (say, `0.01`) would result in the algorithm never terminating (or failing when it reaches the max iterations constraint, if any). Using a reasonable relative tolerance can prevent this and allow the optimization to finish when it reaches the nearly-best-achievable optimum given the FP type.

Z stage positions, of course, don't have this property and behave like a fixed-point data type, so only the absolute tolerance is important. As long as we have a value of _abs_ that is greater than the stage's minimum step size, the optimizer should always converge (barring other issues, such as a pathological target function (=focus score)), as long as we don't run into the FP data type limitations.

So for this case I think we should select _rel_ to be large enough for the `double` type and small enough to be negligible in terms of stage position values. Since we don't expect huge values for the stage position, we can be relatively conservative and use a value that is much larger than the recommended minimum (2 x epsilon).

In fact, in the old code for commons-math 2, we were setting only the absolute tolerance ("accuracy") and leaving the relative tolerance at the default, which appears to be `1.0e-9` (= `new BrentOptimizer().getRelativeAccuracy()`). This seems like a reasonable value, since it translates to 1 nm when the stage position is 1 m (1e6 µm). Thinking of piezo stages, a generous position of 1000 µm would give relative tolerance of 1 pm, again small enough to be negligible.

I will create a patch for this.

nicost:
Thanks!

PR #217, opened 2016-02-23
Reduced dependency of OughtaFocus on internal components and reduced use of math commons 2

nicost:
Uncoupled OughtaFocus from internal.MathFunctions.  Also ported MathFunctions from math commons 2 to 3.  Tested with build-in test function.  This class is possibly only used in Projector plugin, in which case it would be moved there. Reason:  I need to use OughtaFocus with the diSPIM and MMClearVolume plugins and want to reduce the dependencies I need to bring into the MMClearVolume project to get it to run under Netbeans. Also, generally useful to get rid of math commons 2 and to reduce dependencies on internal classes.


ChrisWeisiger:
Looks good to me.

PR #228, opened 2016-03-11
diSPIM: adds stepsize to metadata and post an AcquisitionStartedEvent

nicost:
Small additions to the diSPIM code that has been lingering in on e of my forks for a while.


ChrisWeisiger:
It wasn't anticipated that third-party code would implement any of the events in org.micromanager.events. I know that the documentation didn't state this, however. Is there a particular reason that you need to post an AcquisitionStartedEvent? I think all it will do is confuse our controls that show "stop acquisition" buttons, which I'm guessing won't work to stop a diSPIM acquisition anyway.

Everything else looks fine, and you can "fix" the above problem by just making your SPIMAcquisitionStartedEvent not inherit from AcquisitionStartedEvent, which will cause MM's internal objects to not pay attention to it (I'm assuming that you consume this event elsewhere in the ASIdiSPIM plugin).

nicost:
I am using this event to start the ClearVolume viewer whenever an acquisition (with more than 1 z) starts (when the viewer is so configured).  This event is perfect for that, and if I am not supposed to use it, I would need the ClearVolume plugin to become depended on the diSPIM plugin, which would be very ugly.   I do understand now that the events api was written such that plugins are supposed to only listen to these kind of events, not generate them, but that seems very restrictive.  Would it not be better to have internal events (i.e. events that plugin never see and can never generate), and public events (i.e. events that plugins can see and generate)?  Without something like that,  things will become crooked real fast.

ChrisWeisiger:
Thanks for the explanation. The problem (as we discussed in person) is that this ends up confusing µManager internal code, which does not expect this event to be published by anyone except itself. In commit 1d33b124f74ef0cbac2b37f16eb355d56f8f3362 I've fixed this by adding a `getSource()` method to `AcquisitionStarted/EndedEvent`, which can be used to distinguish between different originators of the event. I've also created a subclass, `AcquisitionSequenceStartedEvent`, that is the only one that provides access to a `SequenceSettings` object. This is so that we can in the future use a different event for when SequenceSettings eventually gets deprecated, without breaking the API.

Thus, for your purposes, you should publish an AcquisitionSequenceStartedEvent with a source that is some internally-controlled object (for example, in µManager's internal code, I used the AcquisitionWrapperEngine object -- don't use that same object, of course!). The ClearVolume code will need to be updated as well to listen for AcquisitionSequenceStartedEvent instead of just AcquisitionStartedEvent.

I hope this satisfies your needs?

ChrisWeisiger:
Incidentally, if you do want to determine if an acquisition originates from within µManager, then d509aa52b9142eddb150faa570cd4028c9451cc5 adds AcquisitionManager.isOurAcquisition() for precisely that test.

nicost:
Th ClearVolume code needs the SequenceSettings (since it needs to know 
how many z planes to expect in this acquisition).  I will therefore 
inherit from AcquisitionSequenceStartedEvent rather than from 
AcquisitionStartedEvent in the diSPIM plugin.  Is there any problem with 
that?

ChrisWeisiger:
No, that's perfectly fine. We added the AcquisitionSequenceStartedEvent solely so that we aren't stuck with all AcquisitionStartedEvents being required to produce SequenceSettings objects.

nicost:
Perfect!  Can this Pull Request now be accepted?

ChrisWeisiger:
Two comments:

1) I would not recommend having the event return itself as the "source". I mean, yes, technically this will work given the current usage of that method in practice, but the intent of the `getSource()` method is that it consistently indicate the "owner" of the acquisition. In other words, you could potentially cause problems down the road if we start using that source object for something more meaningful, and in your case the source is changing every time you run an acquisition.

2) (super minor) The comments in AcquisitionSettings at line 120 are misleading; µManager won't pay any attention to those values because it doesn't care about your acquisition (IMO it's perfectly fine to refer to ClearVolume by name here). 

I don't really care about the second one, to be honest, but the first should probably be fixed.

nicost:
> Two comments:
>
> 1) I would not recommend having the event return itself as the 
> "source". I mean, yes, technically this will work given the current 
> usage of that method in practice, but the intent of the |getSource()| 
> method is that it consistently indicate the "owner" of the 
> acquisition. In other words, you could potentially cause problems down 
> the road if we start using that source object for something more 
> meaningful, and in your case the source is changing every time you run 
> an acquisition.
>

Thing is, I really do not care about that field (nor any other code that 
I am aware of) and creating a source class for this seems just busy 
work.  If anyone would ever want to use that source, they would need to 
become dependent on the ASIdiSPIM internal classes in any case, so I 
would rather leave this for whomever feels that need.

> 2) (super minor) The comments in AcquisitionSettings at line 120 are 
> misleading; µManager won't pay any attention to those values because 
> it doesn't care about your acquisition (IMO it's perfectly fine to 
> refer to ClearVolume by name here).
>

Changed.

ChrisWeisiger:
You shouldn't need to create anything new; just pass in whatever class the plugin uses to run its own acquisitions. For example, in µManager's internal code, I use the AcquisitionWrapperEngine object.

nicost:
So, what is better about passing an instance of private 
runAcquisitionPrivate rather than an instance of 
SPIMAcquisitionStartedEvent?

ChrisWeisiger:
The only difference is that the former is (presumably) an entity that persists for the duration of the acquisition, rather than being discarded as soon as the event is done being consumed. The API requires this per its documentation (which in 274d6ae8445b9725d4781d6b9a2f5023cfb1d605 I have more explicitly clarified).

Imagine that we have some entity that listens for AcquisitionStartedEvent and creates a new object to track that acquisition, based on the source in the event. That object will persist until it gets the corresponding AcquisitionEndedEvent (which you should also be publishing). If the source objects don't match up, then it will persist indefinitely.

nicost:
It seems dangerous to rely on 3rdparty code to generate the start and 
end event with the same object, especially when that 3rdparty code 
itself has no need for it (as in this case). Nevertheless, I added the 
code that you desire.

ChrisWeisiger:
If third-party code doesn't implement things to spec per the documentation, then we can't guarantee that the program will function properly. Kind of like how Swing can't force you to do GUI manipulation in the EDT, but if you do it outside the EDT then bad things may happen.

In any event, this looks fine now.

PR #241, opened 2016-04-04
Fix Flat field plugin in live mode

hadim:
The plugin try to add metadata to UserData even in Live mode, which result in null pointer exception because this metadata does not exist in live mode.

This PR should fix it.


hadim:
Oops I made the PR against `master`... I close for `mm2`.

PR #242, opened 2016-04-04
Fix shading plugin live mode

hadim:
The plugin try to add metadata to UserData even in Live mode, which result in null pointer exception because this metadata does not exist in live mode.

This PR should fix it.

### DIFF
--- a/plugins/MultiChannelShading/src/main/java/org/micromanager/multichannelshading/ShadingProcessor.java
+++ b/plugins/MultiChannelShading/src/main/java/org/micromanager/multichannelshading/ShadingProcessor.java
@@ -18,8 +18,6 @@
 //               IN NO EVENT SHALL THE COPYRIGHT OWNER OR
 //               CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
 //               INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES.
-
-
 package org.micromanager.multichannelshading;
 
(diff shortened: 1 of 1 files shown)


ChrisWeisiger:
I think your commit to "correct" the indentation is not actually desirable. We prefer to use the same indentation level the file currently uses, rather than replace it. Among other things, changing the indentation level means that `git blame` will assign "ownership" of every line to your commit. Also, the MM Java programming style guide is to use 3-space indents. Which is unusual, I know, but if you wanted to switch to 4-space indents then you'd have to modify pretty much every file in the program. So, please remove that commit, and I'll be happy to pull this change in.

hadim:
I tought 4-space was standard... I will revert the commit and ping you when it's done (tonight or tomorrow).

ChrisWeisiger:
Thanks for your understanding. 4-space is certainly the most common standard I'm aware of, but I've also seen 2-space, 6-space, and tabs only. But when in doubt, just imitate what's already there...

hadim:
Done.

I used Alt+Shift+F in NetBeans to auto indent the code and some others lines have been modified. Hope it's fine for you...

hadim:
Tell me if the few indented lines are not desirable.

ChrisWeisiger:
Sorry, lost track of this one...

The additional whitespace changes aren't exactly desirable, but they're not a huge deal either. Just something to watch out for in your future pull requests. Anyway, thanks for the fix!

hadim:
Ok.

PR #244, opened 2016-04-05
Fix user profile not saved on linux machine

hadim:
I have a constant log error : `unable to move user profile to ~/.config/Micro-manager`. The directory can be written and the user profile file in `/tmp` exists.

The bug appears on java 1.6 and java 1.8.

Replacing `renameTo` by `Files.move` makes it work.

What do you think ?

### DIFF
--- a/mmstudio/src/main/java/org/micromanager/internal/utils/DefaultUserProfile.java
+++ b/mmstudio/src/main/java/org/micromanager/internal/utils/DefaultUserProfile.java
@@ -601,9 +601,14 @@ public class DefaultUserProfile implements UserProfile {
             }
             destination.renameTo(backup);
          }
-         if (!tempFile.renameTo(destination)) {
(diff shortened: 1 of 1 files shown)


hadim:
My system is `Operating System: Linux (amd64) 4.3.0-040300-generic`

ChrisWeisiger:
I believe what was actually happening here is that Linux may disallow using `File.renameTo()` if the destination path already exists. The block above the code you modified contained special logic to handle that issue on Windows specifically. I've modified the block to occur on all operating systems. However, I also see no reason not to use `Files.move` instead of `File.renameTo`; I prefer the exception-throwing method over the one that just returns a success/failure boolean.

I've manually incorporated your change, as well as the one I outlined above, and fixed the whitespace "incompatibility". Thanks for the contribution!

hadim:
Nice.

hadim:
I confirm the fix !

PR #245, opened 2016-04-05
Remove error window message in Flat Field plugin 

hadim:
I am not sure if you will accept this PR but I found very annoying to have an additional window with an error message opening when the dark image has been removed from the hard drive in the Flat field plugin ...

MM should be silent about it in my opinion.

Did you consider adding a "Log Window" ? The same way Fiji did recently ? So users without a dev environment would be able to see the log in real time instead of having to open `CoreLogs/*.txt` files.

### DIFF
--- a/plugins/MultiChannelShading/src/main/java/org/micromanager/multichannelshading/MultiChannelShadingMigForm.java
+++ b/plugins/MultiChannelShading/src/main/java/org/micromanager/multichannelshading/MultiChannelShadingMigForm.java
@@ -335,7 +335,7 @@ public class MultiChannelShadingMigForm extends MMDialog implements ProcessorCon
                  DARKFIELDFILENAME, backgroundFileName_);
          studio_.data().notifyPipelineChanged();
       } catch (MMException ex) {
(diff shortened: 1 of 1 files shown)


ChrisWeisiger:
We have some long-term plans to have a more "low-key" way of notifying the user of things they ought to know about, without creating a dialog that they have to immediately deal with, but those probably won't be fully-implemented anytime soon. Having a window that provides a live tail of the CoreLog is also an interesting idea, but not something that users should be required to use just to ensure that the program is working correctly -- 99% of our users don't read log files, nor should they be expected to (and the CoreLog includes a great deal of information that can be safely ignored in most cases, anyway). Users need to be able to trust µManager to tell them in a highly visible fashion when something important has gone wrong; otherwise they may end up performing experiments that are not what they thought they were, which can introduce additional sources of error into their data.

In this case, the user should either intentionally disable dark field correction, or they should provide a valid file; the failure to do either is definitely an error and something the user should be made aware of in an obtrusive manner.

hadim:
Ok I understand your point.

PR #249, opened 2016-04-09
Projector plugin and Generic SLM

nicost:
Fixes a bug in the Generic SLM Device Adapter that causes MM to crash hard.
Adds Checkerboard option to the Projector plugin (very helpful for alignment).
Improves documentation somewhat and speeds up calibration of SLM devices.

### DIFF
--- a/DeviceAdapters/GenericSLM/GenericSLM.cpp
+++ b/DeviceAdapters/GenericSLM/GenericSLM.cpp
@@ -98,7 +98,8 @@ GenericSLM::GenericSLM(const char* name) :
    monochromeColorNum_(0),
    inversionStr_("Off"),
    monochromeColorStr_("White"),
-   chosenDisplayIndex_(-1)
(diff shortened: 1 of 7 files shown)


ChrisWeisiger:
The Java side looks fine to me, but I'll give Mark a chance to comment on the additional delay in the device adapter (commit 18542b6) if he sees fit.

nicost:
Do you want me to back out those C++ changes and submit them straight into the 1.4 branch instead?  If these pull request take more than a few days, I start to forget what is all about (plus I may want to work more on the Projector stuff, but is a bit hard with a pull request sitting there that I need to continue).  B.t.w., the not-initialized pointer in the Generic SLM results often in a hard crash for everyone using the Generic SLM...

ChrisWeisiger:
Sorry about that. I wanted to get Mark's input, but forgot to ask him yesterday. Let's just assume the delay is fine; we can always back it out later.

PR #259, opened 2016-05-02
Add FrameProcessor plugin

hadim:
Comments ?


hadim:
It also has a wiki page now : https://micro-manager.org/wiki/FrameProcessor

:-D

ChrisWeisiger:
Thanks for the contribution. I have a few comments, which are mostly pretty nitpicky:

 * Generally we'd rather not have binary files that aren't necessary to the operation of the program be stored in the source code repository. Specifically here I'm thinking of the screenshot. GitHub can't do diffs of binaries, so any time the file is changed, another copy of the file must be stored...and then everyone who downloads the source code gets every copy of the file, which can rapidly get out of control. Would it suffice to point users to the wiki page, which has the screenshot?

 * We may want to consider having an explicit pipeline that's just for Live mode, in which case your two checkboxes (enable in live mode / enable in acquisitions) wouldn't be needed. I've created issue #260 for this. If/when #260 is implemented, your checkboxes and associated logic should be removed. Of course for the time being they're just fine.

 * Stylistic concern: please put an "_" (single underscore) at the end of the names of class member variables. More generally, please adhere to the [style guide](https://micro-manager.org/wiki/Micro-Manager_Coding_Style_and_Conventions).

 * In `processSummaryMetadata()`, I would recommend omitting the `isAcquisitionRunning()` call and instead checking to see if `intendedDimensions()` is null. That should correctly handle both live mode and MDA and also be more future-proof. People sometimes want to run Live mode in the middle of an acquisition.

 * I must commend you on your carefully-commented regexes. :)

 * I think the log calls in `FrameProcessorFactory` and `SingleCombinationProcessor` can probably be removed now. The use of `Logger` in `SingleCombinationProcessor` should be replaced by `studio_.logs().logError()`. 

 * Remember that the inputs to your Processor may be output by other Processors, and thus may not be images that could physically be generated by any real camera. Also remember that RGB cameras exist (detectable because `Image.getNumComponents() > 1`). Your current `*ProcessImages` methods silently fail when handed an image that doesn't have an expected format; you should log an error in these situations.

Otherwise, looks good!

hadim:
I think all your comments have been addressed. Except for your last point. How do you want me to check whether the image should be processed or not ? In case I don't process it should I display an error message with `logError()` ?

ChrisWeisiger:
You should check the number of components and the number of bytes per pixel in the Image. If there is more than 1 component, or more than 2 bytes per pixel, then you should log an error and either blindly propagate the image or refuse to process it; my vote would be to refuse to process it. Ideally the first time you encounter this scenario you should also _show_ an error to the user (e.g. using `LogManager.showError()`; however, you should be careful to only do this once so the user is not spammed with an error message for every image they acquire.

In the future we plan to have a less-intrusive method to inform the user that something has gone wrong, but I don't have a timeline for that feature yet.

hadim:
Done !

ChrisWeisiger:
Great! Looks good, and thanks for the contribution!

Issue #262, opened 2016-05-04
Autostretch feature takes into account ROI display on the image window


hadim:
I have the feeling the Autostretch feature takes into account the pixel drawn when using the ROI tools for example and so the contrast scale become totally wrong when using such a tool. It's harder to reproduce with the demo config but I totally see it with all the microscopes I use.

Step to reproduce : 

- launch a live window
- check the Autostretch checkbox 
- select the rectangle tool from the ImageJ window
- draw a rectangle on the live window
- the autostretch should start scaling incorrectly the image

(I noticed that using mm2, I didn't try but I am pretty sure I used to saw that also with mm1).

ChrisWeisiger:
This is working as intended. If you have an ROI on the image, then only the pixels in the ROI are a) shown in the histogram, and b) used to calculate autostretch. Or am I misunderstanding what you're saying?

hadim:
Sorry I wasn't clear I think. I am saying that the pixel **of the ROI** (the yellow pixel) used to draw the ROI are taken into account and so the scale become very weird when adding a rectangle or whatever ROI it is on the image window (I am not talking here about the use of a ROI for acquisition on the camera only ROI drawn on the image window).

hadim:
Actually you can use the demo config to show something similar as I have on the microscope by pressing the left button of the mouse when selecting an area with a ROI (whatever the shape is). When the left button is pressed you can notice the min and max of the histogram being set to -1.

hadim:
Oh no sorry I get it now... the area covered by the ROI drawn is the one used for the calculation of the min/max... I didn't know that.... That is why sometime when selecting a very specific region of my sample I had weird histogram not scaling with my whole sample.

hadim:
The thing is sometime I draw a ROI only to put a marker over a spot or a feature of my sample (to see how it moves actually) and I don't want necessarily to use the pixels of the ROI to compute the histogram. 

Would you integrate it if I add a checkbox on the histogram panel to enable/disable this feature ? I you do, could you point me out to the right class to modify, it will save me sometime.

ChrisWeisiger:
Okay, I'm glad we're clear on what's going on. :)

A toggle option to enable/disable this sounds reasonable. I would suggest adding it to the "gear menu" for the histograms panel, which is where we've put a number of other rarely-accessed configurable settings. Said menu is generated by `getGearMenu` in

[...]

PR #263, opened 2016-05-05
Add an option to disable the use of the ROI pixels 

hadim:
Add an option to disable the use of the ROI pixels for scaling the intensity of the image displayed. See #262 for discussion.

@ChrisWeisiger I added the option to the histogram panel (not in the gear menu) because I already did it when I saw your message. I like it here but if you want me to move it to the gear menu, I'll do it.

### DIFF
--- a/mmstudio/src/main/java/org/micromanager/display/DisplaySettings.java
+++ b/mmstudio/src/main/java/org/micromanager/display/DisplaySettings.java
@@ -182,6 +182,7 @@ public interface DisplaySettings {
       DisplaySettingsBuilder histogramUpdateRate(Double histogramUpdateRate);
       DisplaySettingsBuilder shouldSyncChannels(Boolean shouldSyncChannels);
       DisplaySettingsBuilder shouldAutostretch(Boolean shouldAutostretch);
+      DisplaySettingsBuilder shouldScaleWithROI(Boolean shouldScaleWithROI);
       DisplaySettingsBuilder extremaPercentage(Double extremaPercentage);
       DisplaySettingsBuilder bitDepthIndices(Integer[] bitDepthIndices);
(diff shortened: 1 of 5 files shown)


ChrisWeisiger:
I do think it really belongs in the gear menu. This is not an option that I would anticipate users wanting to change frequently, so dedicating space in such a high-visibility location is wasteful. 

Other comments:

 * The description of the shouldScaleWithROI property in the API documentation is misleading. It's not just newly-displayed images; it's all histogram calculations. Something like "Whether histogram calculations should use only the pixels in the current ROI" is an accurate description.

 * You are at risk of getting NullPointerExceptions when casting the shouldScaleWithROI property to a `boolean`. `ContrastCalculator.calculateHistogramWithSettings()` should check for a null value and assume `true` if the value is null; similar logic is needed in `DefaultDisplayManager.calculateHistogram()`.

Otherwise, this looks good.

hadim:
Done.

ChrisWeisiger:
Looks good, except that there's still a chance of NullPointerExceptions in `DefaultDisplayManager.calculateHistogram()`. Fix that and we're golden.

hadim:
Should be ok now. (I used the new Github feature to commit files online, pretty awesome !).

ChrisWeisiger:
Unfortunately, your check is invalid -- a `boolean` (like the function takes as a parameter) can never be null, but a `Boolean` (like the property of the `DisplaySettings`) can be. Since the function takes a `boolean` (chosen because the parameter should not be optional, so we want to coerce its value to be either true or false), any null checks have to happen before the function is called, which is why I specifically mentioned making the fix in `DefaultDisplayManager`.

hadim:
Sorry I didn't pay attention. I'll do that tonight.

hadim:
Hopefully it's ok now !

ChrisWeisiger:
Great, looks good!

PR #277, opened 2016-05-19
Add support for multi-ROI mode

ChrisWeisiger:
This adds functions to Camera devices, the Core, and the Java layer to support the ability to set multiple simultaneous ROIs.

Note that µManager will use the bounding box of each selection component to define the list of ROIs -- thus, if you have a non-rectangular selection like an ellipse or polygon as part of your multi-selection, µManager will use its bounding box instead of erroring out. This is due to a limitation of ImageJ which makes extracting the original true ROI difficult. See this thread:

http://forum.imagej.net/t/extract-sub-rois-from-composite-roi/1645

PR #284, opened 2016-06-08
Add support for setting multiple simultaneous ROIs.

marktsuchida:
Separated out C++ portion of #277. Based on master; to be committed to SVN.


marktsuchida:
In 1.4 now.

marktsuchida:
In mm2 now, as a result of #285.

PR #289, opened 2016-06-09
Update the Hardware Configuration Wizard

ChrisWeisiger:
This makes substantial updates to the Hardware Configuration Wizard's user interface. It also adds a supportsDeviceDetection() method to the Core and a SupportsDeviceDetection() method to the Device interface (and bumps their version numbers accordingly). This method is used to indicate if the device supports automatic detection via port scans.


marktsuchida:
C++ changes have been committed to 1.4 and merged into mm2.
Will create updated pull request.

PR #310, opened 2016-06-25
fix for warnings C4456 in VS2015

jondaniels:
Avoid using the name of an existing variable when creating a new
variable.  The behavior is not incorrect but it generates warning C4456
in VS2015 and can cause confusion when inspecting code.


marktsuchida:
This is now in both 1.4 SVN and mm2. Thanks, Jon.

Issue #324, opened 2016-07-07
Script panel does not "see" classes loaded as plugins


nicost:
For instance, the following Beanshell script:
```
import org.micromanager.projector.ProjectorControlForm;
pcf = ProjectorControlForm.showSingleton(core, mm);
```
Generates an error when run in Micro-Manager 2.0, even when the Projector plugin is open.  It does not generate an error when run in Netbeans (when the Projector plugin is part of the project).  I tried adding the Projector.jar directly to the classpath, but that leads to an error on startup.  I guess this currently means that scripts can not call plugins, which is significant problem for us.

ChrisWeisiger:
This was intentional: we did not want to pollute the main program's namespace with plugin classes, as a matter of principle (otherwise a plugin could insert items into the `org.micromanager` namespace or do other harmful things).

You should still be able to access the plugin in a more general fashion by using the `PluginManager`: `mm.plugins().getMenuPlugins().get("org.micromanager.projector.ProjectorPlugin")` will get you the `MenuPlugin` instance corresponding to the Projector plugin, for example. 

What error do you get on startup if you add Projector.jar to the classpath?

nicost:
Is this documented somewhere?

Putting the Projector.jar on the classpath leads to a bunch of errors.  
Easy enough to try if you want to trouble shoot.

ChrisWeisiger:
The PluginManager is documented in our JavaDocs:
http://valelab.ucsf.edu/~MM/doc-2.0.0-beta/mmstudio/org/micromanager/PluginManager.html
which should be readily discoverable when examining the documentation for the `Studio` class.

The change to how plugins are loaded is not documented so far as I'm aware, because it was expected to be an implementation detail.

As for the classpath issue, I modified the mmimagej launcher by adding

`-classpath "mmplugins/Projector.jar" \

immediately prior to the `-eval` line, and was able to import and instantiate the ProjectorPlugin with no difficulties.

nicost:
> The PluginManager is documented in our JavaDocs:
> http://valelab.ucsf.edu/~MM/doc-2.0.0-beta/mmstudio/org/micromanager/PluginManager.html 
> <http://valelab.ucsf.edu/%7EMM/doc-2.0.0-beta/mmstudio/org/micromanager/PluginManager.html>
> which should be readily discoverable when examining the documentation 
> for the |Studio| class.
>

That is hard/impossible to find out for any non-full-time programmer.  
[...]

PR #326, opened 2016-07-08
Projector plugin changes to make it usable with new plugin loading policies.

nicost:
Projector plugin: Added accessor function so that scripts and plugins can get hold of the Projector plugin
logic and use it.  Necessitated by changes in the plugin loading policies.

### DIFF
--- a/plugins/Projector/src/main/java/org/micromanager/projector/ProjectorPlugin.java
+++ b/plugins/Projector/src/main/java/org/micromanager/projector/ProjectorPlugin.java
@@ -28,9 +28,11 @@ import org.micromanager.Studio;
 import org.scijava.plugin.Plugin;
 import org.scijava.plugin.SciJavaPlugin;
 
-// The Projector plugin provides a user interface for calibration and control
-// of SLM- and Galvo-based phototargeting devices. Phototargeting can be
-// ad-hoc or planned as part of a multi-dimensional acquisition.
+/**
+ * The Projector plugin provides a user interface for calibration and control
+ * of SLM- and Galvo-based phototargeting devices. Phototargeting can be
+ * ad-hoc or planned as part of a multi-dimensional acquisition.
+*/
 @Plugin(type = MenuPlugin.class)
 public class ProjectorPlugin implements MenuPlugin, SciJavaPlugin {
    public static final String MENUNAME = "Projector";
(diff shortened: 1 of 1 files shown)


marktsuchida:
1. Can we please not have a method named `getControlForm()` whose behavior is to _show_ the form? (Maybe name it `showControlForm()` instead?)

2. Please see my comment on #324. Do you think we could do a compromise: make the added method `private`? BeanShell scripts will still be able to call it (after calling `setAccessibility(true);`), and that way we make it explicit that it's not a stable API.

nicost:
Guys, I think this is a very clear example of programming ideals getting 
in the way of work done.

Yes, I have been using the Projector plugin from other code for more 
than a year now, and needs access to it.  To "improve" things this 
access has been taken away and I have no way of getting my work done 
other than trying to fix the problem.  I have no desire to create a 
wonderful super-well designed programming environment, I want to get my 
work done.

If that means switching all our systems back to 1.4, so be it. If you 
want me to work on 2.0, please do let me focus on my work rather than 
creating obstructions.  So, please modify the Projector plugin so that 
it can be accessed programmatically from other scripts and plugins, and 
raise awareness that this is going to be desirable for many other 
plugins as well.


> 1.
>
>     Can we please not have a method named |getControlForm()| whose
>     behavior is to /show/ the form? (Maybe name it |showControlForm()|
>     instead?)
>

No, the goal of that function is to return the singleton instance of the 
form.  If there is no instance yet, it will be shown.  Hence, that 
function returns the controlForm and 'getControlForm" is the appropriate 
name.

> 1.
>
>     Please see my comment on #324
>     <https://github.com/micro-manager/micro-manager/issues/324>. Do
>     you think we could do a compromise: make the added method
>     |private|? BeanShell scripts will still be able to call it (after
>     calling |setAccessibility(true);|), and that way we make it
>     explicit that it's not a stable API.
>

No, since my micro-nuclei plugin code will need the same access. There 
are plenty of examples in the plugins of code accessing non-API parts of 
MM (search for "internal" imports).  IMHO this is fine, as long as the 
plugin writer is aware that this can break things at any point in time.  
Not allowing that kind of access (as was now done for all the plugins) 
will drastically increase the burden on the plugin writers, hence you 
are dramatically reducing contributions from the community, which imho 
is a very bad thing.

marktsuchida:
Discouraging access to internal code is not "programming ideals getting in the way of work done", because code that does that will quickly become unmaintainable, vastly increasing the total amount of work in the long term, only for short-term gain. The fact that your 1-year-old code stopped working is nothing but proof that, once we expose internals, people will complain when we take them away.

However, I'm all for having ways to access internals for experimental/prototyping purposes, as long as that stance is made clear. Chris is now working on creating a separate "library" JAR for Projector. All classes in it will be in an "internal" package, for now, but they will be on the shared classpath (so will be accessible from Java as well as BeanShell).

marktsuchida:
Issue addressed by #330.

PR #328, opened 2016-07-08
Added USB3VisionCamera device adapter

derinsevenler:
This is a device adapter for USB3Vision-compliant cameras. This adapter is heavily based on the previously existing GigECamera adapter, so it unfortunately doesn't conform to the code-formatting guidelines right now (because apparently that adapter doesn't either). It's been working for us for about six months, but only tested with a single camera. It only compiles for 32-bit architectures right now, I'm not sure how much work it would be to make it also compile for x64. These codes are also hosted on their own Github repo here, where you can find a bit more info: https://github.com/derinsevenler/usb3visioncamera 

This is my first time contributing to an open-source project so I am looking for your guidance in how I should format & document my contribution. Thanks everyone!

### DIFF
--- a/DeviceAdapters/USB3VisionCamera/JAISDK.h
+++ b/DeviceAdapters/USB3VisionCamera/JAISDK.h
@@ -1,0 +1,11 @@
+#pragma once
+
+#include <cstdint> // Must be included before JAI headers to prevent redefinition of integer types.
+
+#include "jai_factory.h"
+
+// Trivial conversion between char* and int8_t* for working with JAI's API.
+inline int8_t* str2jai(char* s) { return reinterpret_cast<int8_t*>(s); }
+inline int8_t* cstr2jai(const char* s) { return const_cast<int8_t*>(reinterpret_cast<const int8_t*>(s)); }
+inline const int8_t* cstr2cjai(const char* s) { return reinterpret_cast<const int8_t*>(s); }
+inline const char* cjai2cstr(const int8_t* s) { return reinterpret_cast<const char*>(s); }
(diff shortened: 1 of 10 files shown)


marktsuchida:
Thanks, I'll take a look at this. Would you be okay with renaming it to "GenICam" instead of "USB3VisionCamera" -- or is there code specific to USB3 (e.g. in camera selection?)?

derinsevenler:
I don't believe there's anything specific to USB3 in the device adapter. I would be happy to change the name and increase the scope of the adapter slightly, I also think it makes logical sense. But, I might need help testing the adapter with other GenICam (eg. GigE) cameras because I only have access to a few USB3 cameras right now.

derinsevenler:
I've made some largely cosmetic changes to the adapter so it's renamed GenICam now. It compiles ok for 32-bit Micro-Manager 1.4.23, and works fine with at least one USB3 camera from Point Grey. I don't have access to a GigE camera, so I won't be able to verify that this driver supports them- but I am fairly certain there is nothing USB3-specific.

brdfrnk:
4 years later, was this added in some other aspect?

nicost:
Thanks for bringing this up!  That prompted me to have a go at incorporating it in the build.  I discovered that it needs the JAI SDK to build.  That SDK is no longer supported (see: https://www.jai.com/support-software/jai-software).  Since you are the first to ask and since the SDK is no longer supported, I am wondering if it is worth my effort to add this to the automated builds.  You could try building it from https://github.com/derinsevenler/usb3visioncamera, and if you feel that it is important to add to the MM automated builds, please let us know here.

JiangXL:
I think Aravis will be good alternate for USB3Vision camera.
 
https://github.com/AravisProject/aravis

marktsuchida:
@JiangXL Thanks, good to know.

PR #330, opened 2016-07-08
Add "libraries" build section; split Projector plugin into plugin and separate library

ChrisWeisiger:
This creates a new section of the repository for storing libraries that MM will build and include in the plugins/Micro-Manager directory of the install. It also splits the Projector plugin into the plugin itself (a very thin wrapper) and the library, which handles all of the actual logic of the plugin. This should address the issues discussed in pull request #326.


marktsuchida:
This doesn't yet update the Windows installer. I'll fix that separately.

PR #336, opened 2016-07-12
Script Panel: Better error reporting.

nicost:
An EvalError was previously degraded to a - general - error, and
the getErrorText() function was used to show feedback to the user.
For whatever reasons, an EvalError has much more useful information
in the getMessage() function.  Therefore the formatBeanshellError
function now uses the getMessage() function to provide feedback to
the user for errors of type EvalError.  This greatly helps scripts writers.


ChrisWeisiger:
Oh, excellent! Thanks for tracking this down; it's been bugging me for awhile but I could never figure out where the useful error information was getting swallowed.

PR #341, opened 2016-07-14
Changed separator in ROI String.

nicost:
Micro-Manager internally uses a String to represent an ROI.
Up to now, that string consisted of - separated numbers.  Even
though negative numbers should never show up, they do, and
cause problems because the negative sign is seen as a separator
by the code.  This commit changes the separator to an underscore
and tries to handle old strings with the dash as separator
gracefully.  Note that this is a work-around the separate issue
of ROIs with negative numbers.

### DIFF
--- a/MMCoreJ_wrap/MMCoreJ.i
+++ b/MMCoreJ_wrap/MMCoreJ.i
@@ -633,7 +633,7 @@
       int [] xSize = new int[1];
       int [] ySize = new int[1];
       getROI(x, y, xSize, ySize);
-      roi += x[0] + "-" + y[0] + "-" + xSize[0] + "-" + ySize[0];
+      roi += x[0] + "_" + y[0] + "_" + xSize[0] + "_" + ySize[0];
       return roi;
    }
 
--- a/mmstudio/src/main/java/org/micromanager/internal/utils/MDUtils.java
+++ b/mmstudio/src/main/java/org/micromanager/internal/utils/MDUtils.java
@@ -539,7 +539,13 @@ public class MDUtils {
    public static Rectangle getROI(JSONObject tags)
       throws IllegalArgumentException, JSONException {
       String roiString = tags.getString("ROI");
-      String[] xywh = roiString.split("-");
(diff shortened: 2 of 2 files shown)


marktsuchida:
Although I've long been aware of the problem, this proposed fix will break third-party code and scripts (specifically, those depending on `mmcorej` and those that parse file metadata, the latter of which there are likely many). We also need to ensure that it doesn't affect file I/O (including the reading of existing files) -- though I think that will not be an issue. I don't know what other problems it may cause, but I guess that itself is a problem.

Given that we never professed to support negative numbers in ROI strings, I think the best practical change would be to add the error checking without changing the separators. Is there a case where that wouldn't be sufficient?

It is, of course, also possible to parse `-`-separated strings that also contain negative numbers, even though it is more work.

There may also be other things we can attempt on the camera interface side -- is there a particular test case that generates negative numbers?

nicost:
I was just following Chris' suggestion.  I am happy to take another 
route, but do need a solution fast (as in today).

The negative numbers are easily generated with the Andor Zyla.

ChrisWeisiger:
Good call Mark, I'd forgotten that the ROI spec being in MMCoreJ meant that we can't just change it willy-nilly. Sorry for the bad suggestion, Nico.

marktsuchida:
@nicost Does the Zyla return negative values after you set positive values, or do you only see it when you set an ROI with negative parameters in the first place?
(If the latter, we could just refuse negative values in `CMMCore::setROI()`, which we probably should anyway.)

If you get negative numbers without having set negative numbers (which I guess is what it sounds like), yes, of course the device adapter should be fixed.

But I guess parsing negative numbers correctly is a good idea anyway, just to be safe. I'll take a crack at it now.

nicost:
Nice, thanks!  

Yes, the Zyla returns ```-1``` quite regularly (without setting it).  Not sure what the pattern is, but I have surely seen after changing binning.  The Device Adapter fix will need some understanding of the Andor SDK3 and how it behaves with real cameras.

PR #344, opened 2016-07-14
 MultiChannelShadingPlugin refactoring

nicost:
Removed most dependencies on Micro-Manager internal code.
Deleted code paths that were no longer used.  Added several file headers,
removed unused import statements, and added several comments.
There should be no functional differences.

### DIFF
--- a/plugins/MultiChannelShading/src/main/java/org/micromanager/multichannelshading/ImageCollection.java
+++ b/plugins/MultiChannelShading/src/main/java/org/micromanager/multichannelshading/ImageCollection.java
@@ -30,9 +30,6 @@ import java.awt.Rectangle;
 import java.util.HashMap;
 import org.micromanager.Studio;
 
-import org.micromanager.internal.utils.ImageUtils;
-import org.micromanager.internal.utils.MMException;
-
 /**
  * Utility class for Shading plugin that holds the background and flatfield
  * images.  The images are stored as ImagePlusInfo objects.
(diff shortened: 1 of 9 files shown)


ChrisWeisiger:
While relying on internal code isn't great, I have to think that making your own copy of it is not the ideal solution. Personally I don't really mind third-party code making use of stuff in `org.micromanager.internal.utils`, especially stuff like `ImageUtils` which is very stand-alone and could easily (conceptually) be moved into a different jar for general access.

In fact, now that we have the "libraries" section of the source code, we really ought to start thinking about moving such general-purpose files into separate jars. I'll create issues for that.

nicost:
Once there is a library available, that will be used.  Up till then, pull in as much as possible to avoid issues.  If it is OK to use ```org.micromanager.internal.utils```, why not move it to ```org.micromanager.utils```?

marktsuchida:
The only problem is that not everything in `org.micromanager.internal.utils` is good for general use. Some of it is general enough but not packaged as a safe API; a lot is actually closely tied to mmstudio internals or conventions; others are poor reimplementations of things that are available in better forms in third-party libraries (such "limited reinventions of wheels" aren't always bad -- sometimes it is simpler to write a quick implementation than to depend on a massive and complex external library -- but things written for such reasons clearly shouldn't be shared with other projects).

With regard to `ImageUtils` specifically, aside from some horrendous methods it includes (like `makeProcessor()` whose behavior depends on MMCore state), it's a really good question whether we should have our own implementations of these functions at all. Ideally we would use ImgLib2 or some modern ImageJ2 interface (OPS?) to do things like image subtractions. It seems a little odd to start providing an API whose sole purpose is to make it convenient to use old-fashioned classes like `ImageProcessor` and `ImagePlus` (whereas having internal utility methods for that purpose is fine). I guess things will become easier once we have a good example of a plugin doing fairly simple image processing using modern APIs.

So I _do_ mind if third-party code calls things in `org.micromanager.internal.utils`. However, I don't think it's a big issue if _plugins included in our source tree_ call such classes (for now), since we can ensure that they continue to build. New code (even if in our repo) should avoid this, but I don't think that, in general, we should introduce duplicate code into existing plugins for the sole purpose of eliminating use of `internal` MMStudio classes. I think it is better to just use `internal` classes until we have a cleaner way of doing things.

Despite the reservations I have about adding duplicated code, `subtractImageProcessors()` is actually only used by Projector and MultiChannelShading, meaning that it is a candidate for deletion from MMStudio (Projector only performs float image subtraction, which is easy to implement without `ImageUtils`). So I'm going to say that the minimal duplication introduced here is likely temporary and acceptable.

PR #349, opened 2016-07-15
MultiShading plugin: Add missing code and fix negative ROIs.

nicost:
Previous refactoring left out processing of Floatprocessors, which
is needed at runtime.  Also, negative ROIs that now make it though to
this plugin cause problems with image dimensions due to the way they
are handles by the ImageJ setROI function.  Therefore, negative x and y
positions are moved to 0, which works around this porblem, possible at
cost of a slightly offset flatfield (not sure if that is really the case).

### DIFF
--- a/plugins/MultiChannelShading/src/main/java/org/micromanager/multichannelshading/ImageCollection.java
+++ b/plugins/MultiChannelShading/src/main/java/org/micromanager/multichannelshading/ImageCollection.java
@@ -250,6 +250,16 @@ public class ImageCollection {
       } else {
          resultProcessor = ipi.getProcessor().duplicate();
       }
+      // HACK/Fix: The Andor Zyla often returns an ROI with roi.x ==-1 pr roi.y == -1
+      // That creates problems because the image after setRoi will be one pixel
+      // to small (i.e., the image should always have the correct height and width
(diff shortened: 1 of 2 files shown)


ChrisWeisiger:
Stylistically, it would be better to do your bounds checking as `x = Math.max(0, Math.min(image size, x))`, instead of just checking for negative numbers. That way you catch numbers that are too big as well as too small.

However, I don't see that as a reason to block this change; you can always add that improvement later.

nicost:
Not sure where the parameter ```image size``` in your code would come from.  I assume that it represents the size of the image sensor for the given binning (possibly minus the width of the ROI).  That information is not easily available (the only approach I can think of is using the ```core.clearROI()``` function, asking the camera for the image width (hoping that the camera will give the correct value even before an image is snapped), and res-setting the original ROI.  That procedure will certainly cause significant slow down for some cameras and the only reason to do it is to exclude a possibility we have not seen in the wild yet (a camera lying about its ROI in such a way that it places the starting coordinate outside of the sensor).  

Whether to use ```a = Math.max(a, b);``` or ```if (a < b) a = b;``` I would leave to the code writer.  I would not be surprised if the latter is faster (although that is of little concern here).

PR #354, opened 2016-07-18
Beanshell startup script, fix for issue with inspect function.

nicost:
The inspect function declared in scriptpanel_startup.bsh no longer
worked and barfed in line 216 in which a null value for the String
value was used.  By setting the variable value to a String "null", this
error is avoided.  It is a bit unclear why the problem occurred to begin
with, but I do not think that warrants the time to figure out an explanation.


ChrisWeisiger:
If I had to guess, this may have occurred when we updated the Beanshell jar. Presumably the old jar allowed variables to persist outside of their block scopes (the `try/catch` in that method declares the `value` variable for the first time), and the new jar causes variables to vanish once they're out of scope. Thanks for tracking this down.

PR #360, opened 2016-07-20
Generic SLM: avoid exception

nicost:
Occasional crashes were caused by an exception caused by an
uninitialized variable.  Since it passed the test for NULL, it
was not properly initialized.  Setting it to NULL explicitly
in the constructor avoids that situation (the short title of this
commit is misleading).
Also added some more error testing, error reporting and error
propagation.


marktsuchida:
Merged into 1.4 SVN (r16104).

PR #365, opened 2016-07-22
Added Magellan to MM2.0 repository and updated to new MM2.0 APIs

henrypinkard:


marktsuchida:
@henrypinkard Great to see this coming to MM2!

Upon initial review it looks like the source directories will need to be changed to match the mm2 convention (which is the Maven directory layout): `src/org/micromanager` needs to be moved to `src/main/java/org/micromanager`. The `.png` files (but not the `.form` files) will need to go into `src/main/resources/org/micromanager`.

There are also changes needed in the build system, to include some extra libraries, but I should be able to handle those.

marktsuchida:
See #390; merged.

Issue #274, opened 2016-05-12
Inspector window is always on top


nicost:
Maybe this is heretic, but now working with 2.0 more regularly, I start to dislike the always on top behavior on the Inspector window.  It is difficult to arrange windows around it, so that their content becomes visible.  I do appreciate our consideration that it is an important window, and that it should stand out, but currently, it is more bothersome then helpful.  Not super important, but thought it useful to bring this up.

ChrisWeisiger:
One of the common complaints about 1.4 was that when you get more than a few image windows open, it becomes impossible to keep easy access to the main window where your histograms and metadata are. Making the inspector be always-on-top is our solution to that problem. Can you propose an alternative solution?

hadim:
Maybe a big visible button on the main MM window that will bring the Inspector window on top once clicked ?

nicost:
Hey, I am not the UI expert, just a user;)  I can think of various approaches.  You could bring the Inspector to the front whenever a new image window is opened.  There could be a button on the image window to bring the inspector to the front.  The problem with the Inspector always-on-top is that the Inspector window can get quite large and leave little space on the screen for plugins, images, etc.. Also, its behavior to disappear whenever Micro-Manager goes out of focus (i.e. when switching to another application) makes it harder to work with it.

ChrisWeisiger:
@hadim Keep in mind that the main window also does not stay on top, so you're still stuck with trying to find a specific window (or with your suggestion, one of two specific windows) in your massive pile.

We might be able to rig up something on the display window to let you raise _an_ inspector, but keep in mind there is no "_the_" inspector window. We create one automatically when you first create an image window, but the user is free to close that, create their own, etc.

What kind of display are you working with? I'll freely agree that on my 13" Macbook, the inspector frequently gets in the way, but I'm not trying to drive a microscope on this system. I'd expect most of our users will be working with a 21" display at minimum.

I'd prefer if the inspector stay visible when switching applications, but unfortunately the "disappear when focus switches" behavior seems to be something at the operating system level, so we're unlikely to be able to fix that.

nicost:
I am using it on 21" displays on microscopes, and find it to be as hard to work with as on smaller displays.  I completely agree that the idea is good, but in practice it does not work so well.

hadim:
I know but (at least for me) I always try to keep the main window at the same place without any other window on top of this one. So it's always visible.

marktsuchida:
[...]

PR #367, opened 2016-07-23
Provide option to set/unset the Inspector Window to be always on top

nicost:
This addresses issue #274 by creating an option to have the inspector window always on top.  
The option is set to true by default, and the user's choice will be remembered in the profile.
This will keep the current behavior for novel users, but users in the know who are bothered by
an inspector window that is always in the way, can change it.
Currently, the choice will not propagate to inspector windows that are currently open. This could
be added, but does not seem very important.


nicost:
Oops, should have been against mm2.  Closing...

PR #368, opened 2016-07-23
Provide option to set Inspector Window always on top choice

nicost:
This addresses issue #274 by creating an option to have the inspector window always on top.

The option is set to true by default, and the user's choice will be remembered in the profile.
This will keep the current behavior for novel users, but users in the know who are bothered by
an inspector window that is always in the way, can change it.
Currently, the choice will not propagate to inspector windows that are currently open. This could
be added, but does not seem very important.

### DIFF
--- a/mmstudio/src/main/java/org/micromanager/display/internal/inspector/InspectorFrame.java
+++ b/mmstudio/src/main/java/org/micromanager/display/internal/inspector/InspectorFrame.java
@@ -29,7 +29,6 @@ import java.awt.Cursor;
 import java.awt.Dimension;
 import java.awt.GraphicsConfiguration;
 import java.awt.Insets;
-import java.awt.Point;
 import java.awt.Rectangle;
 import java.awt.event.ActionEvent;
 import java.awt.event.ActionListener;
@@ -44,9 +43,7 @@ import java.io.File;
 
 import java.util.ArrayList;
 import java.util.Collections;
-import java.util.HashMap;
 import java.util.List;
-import java.util.Stack;
 
 import javax.swing.BorderFactory;
 import javax.swing.JButton;
@@ -122,14 +119,14 @@ public class InspectorFrame extends MMFrame implements Inspector {
     * the panel itself.
     */
    private class WrapperPanel extends JPanel {
-      private JPanel header_;
+      private final JPanel header_;
       private InspectorPanel panel_;
       private boolean shouldOverrideSize_ = false;
       private Dimension overrideSize_ = null;
-      private String title_;
+      private final String title_;
       public WrapperPanel(String title, InspectorPanel panel) {
          super(new MigLayout("flowy, insets 0, fill"));
-         setBorder(BorderFactory.createRaisedBevelBorder());
+         super.setBorder(BorderFactory.createRaisedBevelBorder());
          panel_ = panel;
          title_ = title;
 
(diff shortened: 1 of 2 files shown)


ChrisWeisiger:
For propagating the status to existing windows: this can easily be done:

 * Add an InspectorWindowAlwaysOnTopEvent (or something similar) class to `org.micromanager.events.internal`.
 * Publish the event on the application event bus when the status is changed (`DefaultEventManager.getInstance().post(event)`)
 * Add an event subscriber to `InspectorFrame`. This class is already registered with the application event bus, so you don't need to register it.

Events are cheap, and this provides a convenient, easy mechanism for inspectors to be notified about the change in status without needing to track all of the inspector instances or do unpleasant examination of all open windows. Indeed, this kind of thing is exactly what the event mechanism is for!

I would prefer that we always go to the profile when we need to access this value, rather than storing a second copy of the state in the `alwaysOnTop_` variable. Accessing the profile is cheap, and this eliminates the possibility of the variable getting out of sync with the profile. Plus static variables are generally icky. :)

Please always use curly braces even for one-liner conditional statements (c.f. lines 317-318 of `InspectorFrame.java`). Omitting the curly braces makes it easy to introduce bugs later where someone adds another line to the conditional without realizing they also need to add the curly braces, so it's better to just always use them. Actually that line can be better written as `setAlwaysOnTop(DefaultUserProfile.getInstance().getBoolean(...))`. 

Does NetBeans really complain if you don't specific `super.` in front of all of those method calls? That looks super-ugly to me; when I see `super.` in front of a method invocation, it implies to me that this class normally overrides the specified behavior, but wants to specifically use the super's implementation in this case instead. If you're trying to eliminate NetBeans warnings, in this case I would prefer to modify what NetBeans considers to be warning-worthy.

nicost:
On 7/25/16 8:23 AM, ChrisWeisiger wrote:
>
> For propagating the status to existing windows: this can easily be done:
>
>   * Add an InspectorWindowAlwaysOnTopEvent (or something similar)
>     class to |org.micromanager.events.internal|.
>   * Publish the event on the application event bus when the status is
>     changed (|DefaultEventManager.getInstance().post(event)|)
>   * Add an event subscriber to |InspectorFrame|. This class is already
>     registered with the application event bus, so you don't need to
>     register it.
>

Could you add that?

> Events are cheap, and this provides a convenient, easy mechanism for 
> inspectors to be notified about the change in status without needing 
> to track all of the inspector instances or do unpleasant examination 
> of all open windows. Indeed, this kind of thing is exactly what the 
> event mechanism is for!
>
> I would prefer that we always go to the profile when we need to access 
> this value, rather than storing a second copy of the state in the 
> |alwaysOnTop_| variable. Accessing the profile is cheap, and this 
> eliminates the possibility of the variable getting out of sync with 
> the profile. Plus static variables are generally icky. :)
>

Could you add that too?
>
> Please always use curly braces even for one-liner conditional 
> statements (c.f. lines 317-318 of |InspectorFrame.java|). Omitting the 
> curly braces makes it easy to introduce bugs later where someone adds 
> another line to the conditional without realizing they also need to 
> add the curly braces, so it's better to just always use them. Actually 
> that line can be better written as 
> |setAlwaysOnTop(DefaultUserProfile.getInstance().getBoolean(...))|.
>

Please feel free to change.

> Does NetBeans really complain if you don't specific |super.| in front 
> of all of those method calls?
>

Yes.

> That looks super-ugly to me; when I see |super.| in front of a method 
> invocation, it implies to me that this class normally overrides the 
> specified behavior, but wants to specifically use the super's 
> implementation in this case instead. If you're trying to eliminate 
> NetBeans warnings, in this case I would prefer to modify what NetBeans 
> considers to be warning-worthy.
>

I agree with Netbeans and like to be specific.  Feel free to change if 
you do not like it.


In the mean time, I really want to see the option to have the Inspector 
not always on top.  Should I fork and distribute my own 2.0?

marktsuchida:
Merged manually with some additional commits. Setting change is now reflected immediately.

For use of `super.` when calling non-overrided methods in ctor, see http://stackoverflow.com/a/3404369.

PR #376, opened 2016-07-26
Projector plugin: Fix bug that caused arrays of OvalRois to throw a null pointer exception

nicost:


ChrisWeisiger:
For reference, the array parameter to `ArrayList.toArray()` doesn't need to have any size. E.g. `ArrayList.toArray(new Roi[] {})` is perfectly valid. Not a big deal though.

nicost:
https://docs.oracle.com/javase/6/docs/api/java/util/ArrayList.html#toArray(T[]) 

toArray

public <T> T[] toArray(T[] a)

    Returns an array containing all of the elements in this list in proper sequence (from first to last element); the runtime type of the returned array is that of the specified array. If the list fits in the specified array, it is returned therein. Otherwise, a new array is allocated with the runtime type of the specified array and the size of this list. 


@ChrisWeisiger : If you don't like it, just change it.  This kind of lecturing is quite off-putting.  It really is that I need this to work, or I would stop contributing with maintainers who act like this.

PR #398, opened 2016-08-11
DayNightTime: Add TitledBorders to the Day/Nighttime control.

nicost:
Also added Separator and TableHeader, but did not check if those work.


marktsuchida:
Can I ask what component (or part thereof) this is intended to fix, and on which platform? Just want to be sure I'm looking at the right case.

nicost:
This fixes the TitledBorder that is used in several plugins (ASIdiSPIM 
being one of them).  While trying out what was actually needed, I did 
add a few more (Separator and TableHeader if I remember correctly), that 
could possibly be removed (but then will look bad if any code will ever 
use them). It seems to me that it would be better to set all occurrences 
of Color Keys in UIManager that contain ".foreground", rather than an 
almost complete subset as is done now.  TitleBorder is different since 
it uses the extension ".titleColor"

marktsuchida:
@nicost I think it makes sense, thanks.

PR #400, opened 2016-08-11
mmStudio: File Open command now runs on its own thread,

nicost:
so that the progress dialog actually becomes visible.  The same strategy
as correctly used in the implementation of the OpenRecentMenu was applied
to the promptToOpenFile method.  Also removed some superfluous imports and
other minor fixes.

### DIFF
--- a/mmstudio/src/main/java/org/micromanager/display/internal/DefaultDisplayManager.java
+++ b/mmstudio/src/main/java/org/micromanager/display/internal/DefaultDisplayManager.java
@@ -29,10 +29,10 @@ import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Stack;
+import java.util.concurrent.ConcurrentHashMap;
 
 import javax.swing.JOptionPane;
 
-import org.micromanager.data.Coords;
 import org.micromanager.data.Datastore;
 import org.micromanager.data.DatastoreFrozenException;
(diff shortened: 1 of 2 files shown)


marktsuchida:
I suspect `DisplayManager.manage()` is not currently thread-safe, so moving the call to a background thread could cause races (which probably means the "open recent" implementation is already problematic). Rather than calling it on the EDT, it might make sense to make the "manage" API methods thread-safe.
I'm happy to look into that if you can give me a few days. It will need to be determined whether the "manage" APIs will need to run on the EDT or if they can safely be synchronized to run on an arbitrary thread. Please also feel free to propose other solutions.

nicost:
I think that synchronization of DefaultDisplayManager::manage(Datastore) is harmless (but can not completely oversee).  I can either add the synchronization or put it in a SwingUtilities.invokeLater.  It would really be nice to fix these, as the current blocking behavior is clearly wrong (and especially notable when opening large files).

marktsuchida:
Thanks, I think synchronization of the list of managed datasets is the way to go.

nicost:
I changed the HashMap storeToDisplays_ into a ConcurrentHashMap.  I think that should do the trick.

marktsuchida:
@nicost Actually this doesn't make it thread safe (there are still races between multiple calls on `storeToDisplays_`, for example). Thanks, anyway; I will patch this up after merging.

Issue #419, opened 2016-08-18
Not closing the app when quitting doesn't clean up windows


marktsuchida:
When "Close app when quitting MM" is unchecked in Tools > Options, the following windows stay open (but not functional):
- Image Inspector
- Messages
- Quick Access Panel
- Stage Control
- Probably, various plugins

PR #432, opened 2016-08-23
mmstudio and plugins: Windows now listen for the ShutdownCommencing Event.

nicost:
By listening to this event, these windows will close when shutdown starts and do not rely on a 
System.ext(0) to close.  The plugins that I changed (not all yet), listen to the external event.  It 
may be good to add this function to the MMPlugin interface so that plugin writers realize that they
may want to implements this.

Addresses issue #419


nicost:
Addressed in Pull Request #439.

Issue #423, opened 2016-08-19
Quick Access Panel should enforce unique names


marktsuchida:
When creating, suffix " 2", " 3", etc.
When renaming, require unique name.

nicost:
Addressed in Pull Request #436.

PR #436, opened 2016-08-27
QuickAccess Panels: Fixes naming bugs

nicost:
Addresses issue #423.  Makes sure that new Panels have unique names, and that changed names are unique.  Also removes the use of com.bulenkov.iconloader.IconLoader (which should be removed everywhere since it is trivial code), makes variables final whenever possible and add some Javadoc.

### DIFF
--- a/mmstudio/src/main/java/org/micromanager/quickaccess/internal/DefaultQuickAccessManager.java
+++ b/mmstudio/src/main/java/org/micromanager/quickaccess/internal/DefaultQuickAccessManager.java
@@ -19,7 +19,6 @@
 
 package org.micromanager.quickaccess.internal;
 
-import com.bulenkov.iconloader.IconLoader;
 import com.google.common.base.Charsets;
 import com.google.common.eventbus.Subscribe;
 import com.google.common.io.Files;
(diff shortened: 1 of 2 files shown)


ChrisWeisiger:
Have you verified that creating a new ImageIcon instead of using IconLoader correctly uses high-DPI images when they are available? That is the reason why IconLoader is used: it identifies when the icon being drawn is on a high-DPI display, and then draws the "@2x" version of the icon if available (e.g. instead of drawing foo.png, it draws foo@2x.png). Not doing this results in a very ugly GUI on high-DPI displays.

nicost:
Ah, Sorry!  Completely forgot about that and though we were still using the old IconLoader.  Will revert those changes.

ChrisWeisiger:
Oh, you're thinking of that thing that we replaced with `getClass().getResource(path_to_file)`, aren't you? Yeah, as far as I'm aware, that module has been completely excised from the code and isn't even in the build any more.

nicost:
Perfect!  Sorry for my mis-understanding/poor memory.  Just reverted back the icon loading code (and I'll remember to use the same in the future).

marktsuchida:
Merged manually as commit 4436785874f98fbcfd1ac7fa19e66efc89b3e465.

Issue #434, opened 2016-08-24
Userdata in Metadata are unreadable


nicost:
The "userData" field in the metadata browser is presented as a comma separated string of key:value pairs in the metadata browser.  If there are more than one or two key:value pairs, it becomes impossible to read them (I tried making the windows wide, but that makes only one or two more visible).   

Possible solutions:
1. List Userdata as all other metadata, but at the bottom of the frame, separated by a line indicating these are "Userdata".
2. Make the field clickable to bring up a frame that lists them in a readable way.

I think that I like option 1 best, but any other solution will do.

PR #440, opened 2016-08-29
MetadataPanel: Better display of UserData

nicost:
For both per-image and summaryMetadata, now shows all userData as key-value pairs where the key is preceded with "userData-".  Note that the capitalization is different for image and summarymetadata, not sure if I caused that and/or that should be normalized.  

This PR include PR #435, which introduced the userdata in the summaryMetadata.

Fixes issue #434


marktsuchida:
The capitalization is indeed different internally ('userData' for plane metadata; 'UserData' for summary metadata), and it has been so long enough that we shouldn't change it in the file format. But for display we can do whatever we like.

PR #448, opened 2016-09-14
Change Timeout to "Core-TimeoutMs"

impact27:
This changes allows users to change the timeout value of the camera

### DIFF
--- a/DeviceAdapters/QCam/QICamera.cpp
+++ b/DeviceAdapters/QCam/QICamera.cpp
@@ -4203,7 +4203,9 @@ int QICamera::QISequenceThread::svc()
    }
 
    // set a reasonable timeout (in msec) for waiting on frames
-   timeout = (long)(5 * m_pCam->m_dExposure + 50000);
+   char ctimeout[MM::MaxStrLength];
+   m_pCam->GetCoreCallback()->GetDeviceProperty("Core", "TimeoutMs", ctimeout);
+   timeout = (long)(5 * m_pCam->m_dExposure + atof(ctimeout));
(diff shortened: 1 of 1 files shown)


marktsuchida:
Thanks, @impact27. Can I ask if the motivation is to be able to set longer timeouts (because you are using external trigger?), or shorter ones (why?)?

I'm wondering in part because your change would change the default from (5 * exposure + 50) sec to (5 * exposure + 5) sec, which is rather drastic when the exposure is short. This will break existing configurations where people are expecting the camera to wait at least 50 seconds for an external trigger. So we might want to think about a different way to make the timeout configurable.

impact27:
I am trying to use a logarithmically spaced external trigger. The reason I made this change is that as there is already a setting for a timeOut, I would expect it to change all the time outs? As a user, the Core-TimeOutMs was the first think I changed when I had this problem. The alternative would be to add another setting but there is already two. (The second one being only accessible with BeanShell, but I am unsure how to access this one from the cpp file)

impact27:
For reference: http://micro-manager.3463995.n2.nabble.com/No-way-to-change-timeout-time-td7585801.html#a7585832

impact27:
Or the default value of Core-TimeoutMs could be changed to 50000

marktsuchida:
If this is still needed, we need a different solution, because `Core-TimeoutMs` is for waiting for devices to exit their "busy" state (it is already too restrictive that we only have one timeout for all devices). The only way to allow control of a camera's external trigger timeout is to add a camera-specific property.

PR #449, opened 2016-09-15
Add the possibility to change Sync B

impact27:
Add a setting to change the output pin "Sync-B" from micro Manager
This is basically copied from "TriggerType"


nicost:
Committed to subversion in r17220.  Sorry this took so long.

PR #467, opened 2016-10-13
Two fixes for the PrecisExcite adapter (master)

zindy:
These are the changes pushed to master instead of MM2.

To recap:

**First fix** concerns the erroneous behaviour (when shutter open) of turning on any selected channel via the PrecisExcite.ChannelLabel property without switching off the deselected channels. With shutter closed however, single LED channels were selected as expected.

**Second fix** concerns the population of PreciseExcite.ChannelLabel with invalid channel names. This has the effect of delaying any PreciseExcite property read-out, since the adapter needs to wait for the serial communication to timeout for invalid channels.

I have tested this with my pe-300 light source. I will check this with the pe-2 later.

According to my logs, there are still two timeouts left during the adapter initialisation, but I think we can live with those for now.

Cheers,
Egor

### DIFF
--- a/DeviceAdapters/PrecisExcite/PrecisExcite.cpp
+++ b/DeviceAdapters/PrecisExcite/PrecisExcite.cpp
@@ -470,7 +470,7 @@ void Controller::Illuminate()
    {
       if (triggerMode_ == OFF) {
          msg << "SQZ" << carriage_return;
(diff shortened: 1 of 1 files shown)


marktsuchida:
This is now in 1.4 SVN (r16231-r16233) and will shortly be in mm2.
Apologies for the long delay.

PR #468, opened 2016-10-20
Mm2 : added an option of FrameCombiner to combin Z dimension images

BioinfoTongLI:


hadim:
I have tested the PR and it works well.

The only issue we have is when doing an acquisition without Z slices and asking the plugin to combine Z slices. We don't know how to catch error... Anyway this situation should not happen very often.

hadim:
@marktsuchida could you have a look at this PR when you'll have sometime please ?

hadim:
@nicost any reasons for closing this ?

nicost:
I could swear that I merged it!  Must have pressed the wrong button!

hadim:
Ahah. Thanks @nicost

hadim:
@nicost please look at https://github.com/micro-manager/micro-manager/pull/486

nicost:
It looks like the modified file FrameCombinerConfigurator.form was omitted from this PR.  Can you please open a new PR for that?  Or did you manually change the code?  If so, not a good idea without rewriting the UI completely (which would be a good thing to do, we have been moving everything from the Netbeans designer to MigLayout.  In any case, the dialog currently looks quite bad on Windows, and most of the text is quite incomprehensible.  Would be great if that all could be fixed.  Also, please stick to coding style (https://micro-manager.org/wiki/Micro-Manager_Coding_Style_and_Conventions).  I'll shortly submit a PR to clean some things up.

hadim:
Sorry for the mess. Netbeans wasn't properly configured and I didn't notice it.

As for the UI, the original plugin never had a `.form` file. I have always used the Netbeans designer and it never have been an issue.

Also the configuration window looks ok to me on Windows (tested some weeks ago). What did you notice ?

hadim:
Maybe some changes added by @BioinfoTongLI modified the interface. @BioinfoTongLI could you have a look at this please ?

nicost:
The .form file is in the repository:  https://github.com/micro-manager/micro-manager/tree/mm2/plugins/FrameCombiner/src/main/java/org/micromanager/plugins/framecombiner.  The window looks horrid, lots of blank space on top, the "z", "Time", dropdown is way too wide, some text is incomprehensible (what does " < total frame/slice if using MDA" mean?  Without the correct .form file I have no way to correct this.

hadim:
Ok sorry for the noise, in fact there is a `.form` file and it has always be here. I just never noticed it since Netbeans hides it... :-(

I am pretty sure @BioinfoTongLI didn't use it to modify the configuration dialog. I am doing a PR.

hadim:
Please see https://github.com/micro-manager/micro-manager/pull/490

BioinfoTongLI:
@hadim sorry, i noticed the .form from the beginning. But i didn't understand that it is related to the interface, that's why didn't asked you why it is there. I will be more caucious next time.

PR #484, opened 2016-11-12
added Tenengrad measure to available list of focus scoring algorithms

jondaniels:


jondaniels:
added new focus scoring algorithm "Tenengrad" inspired by recent paper claiming after much comparison/testing this is best-performing non-spectral algorithm

nicost:
This change is now in 1.4 (r16219), and there is a pull request for the same change in 2.0.  I'll close this PR now.  Thanks Jon!

PR #486, opened 2016-12-04
Revert java 1.8 only syntax to be compatible with -source 1.6 in FrameCombiner

hadim:
`FrameCombiner` plugin used java 1.8 only syntax and I think (need to be confirmed) that MM is build with java 1.6. I guess that could cause issues...

This PR revert the syntax to a one compatible with java 1.6.


nicost:
Yes, that certainly causes issues.  MM still uses 1.6 and will not compile with 1.8.  Guess I should have checked that, but I don't really have the bandwidth to do so.

PR #489, opened 2016-12-04
[WIP] PositionSplitter plugin

hadim:
This PR is only for discussion. Do not merge.

### DIFF
--- a/mmstudio/src/main/java/org/micromanager/internal/script/ScriptPanel.java
+++ b/mmstudio/src/main/java/org/micromanager/internal/script/ScriptPanel.java
@@ -112,12 +112,12 @@ public final class ScriptPanel extends MMFrame implements MouseListener, ScriptC
    private StyleContext sc_;
    private Interpreter beanshellREPLint_;
    private JConsole cons_;
-   
+
    public static final FileType BSH_FILE
            = new FileType("BSH_FILE","Beanshell files",
                     System.getProperty("user.home") + "/MyScript.bsh",
                     true, "bsh");
-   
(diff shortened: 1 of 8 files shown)


hadim:
I am not sure this is the correct way to achieve what I want.

hadim:
The thing is I am creating multiple `Datastore` in the `ProcessorPlugin` and I am not sure it is the correct way to do it.

@nicost  : Do you think this kind of feature could fit in a `ProcessorPlugin` ? Or does it need a deeper integration in the core of MMStudio ?

I think this feature even would help some MM users such as @julou for example (see this for some background http://micro-manager.3463995.n2.nabble.com/On-stack-datasets-with-separate-files-per-position-td7586816.html).

hadim:
Thank you for taking time to comment about that.

So my feeling was correct : `ProcessorPlugin` is not the right way to achieve this.

Here is my idea : 

First I find "Save XY positions in separate Image Stack Files" confusing because even if XY positions are saved in a different files, they are all linked together into the same OME data model (that is why bioformat or IJ2 can't open them individually).

What I want is a *true* "Save XY positions in separate Image Stack Files" option that save a different and independent OME data model per XY positions.

What I propose is a new `Storage` class that will hold a different `StorageMultipageTiff` object per XY positions. I am not very familiar with the internal MM api so I don't know if it's something feasible.

I'll try to come up with a more concret proposal in a PR.

hadim:
Let's continue the discussion there https://github.com/micro-manager/micro-manager/pull/494

PR #490, opened 2016-12-04
Fix FrameCombiner UI

hadim:


hadim:
@nicost I think  it's good to be merged.

PR #494, opened 2016-12-10
New save mode "Image Stack File per XY Positions"

hadim:
See https://github.com/micro-manager/micro-manager/pull/489 for the motivation.

This PR is a **WORK IN PROGRESS. DO NOT MERGE.**

### DIFF
--- a/mmstudio/src/main/java/org/micromanager/acquisition/internal/MMAcquisition.java
+++ b/mmstudio/src/main/java/org/micromanager/acquisition/internal/MMAcquisition.java
@@ -49,12 +49,14 @@ import org.micromanager.data.NewImageEvent;
 import org.micromanager.data.Pipeline;
 import org.micromanager.data.PipelineErrorException;
 import org.micromanager.data.Storage;
+import org.micromanager.data.SummaryMetadata;
 import org.micromanager.data.internal.CommentsHelper;
 import org.micromanager.data.internal.DefaultDatastore;
 import org.micromanager.data.internal.DefaultSummaryMetadata;
 import org.micromanager.data.internal.StorageRAM;
 import org.micromanager.data.internal.StorageSinglePlaneTiffSeries;
 import org.micromanager.data.internal.multipagetiff.StorageMultipageTiff;
+import org.micromanager.data.internal.multipagetiff.XYSplitStorageMultipageTiff;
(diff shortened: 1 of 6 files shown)


hadim:
Code is not working for now because I need access to metadata during the `XYSplitStorageMultipageTiff` construction and `store_.getSummaryMetadata()` return a `null`.

When someone has time to have a look and tell me if I am on the right track, that would be nice.

nicost:
I don't think that you want to be doing all this work in the constructor.  That approach mandates that the DataStore knows in advance how many positions it is going to contain (which is a very limiting requirement and certainly currently not true).  Instead, you could use the putImage function to detect the XY position (or index) of the given image and create a new DataStore if needed.  I am a little bit suspicious of the semi-circular approach (a DataStore that uses a Storage, which in turn creates more DataStores which all have their own Storage), but it could work.

hadim:
I have made some progress trying to move all the logic into `putImage()` but the task is hard for someone who does not have a good understanding of the API.

I won't probably be able to work on it in a near future.

hadim:
In the meantime I have created a script that do the job **after** the acquisition (script process the current active window).

For now the generated image stacks can't be open in BF/SCIFIO (only in MM). I guess something is wrong with the way OME XML is written. I'll try to fix that later.

The script : https://github.com/hadim/mm_scripts/blob/master/scripts/Separate_XY_Positions.bsh

(MM2 API for scripting is really awesome !)

hadim:
The new version of https://github.com/hadim/mm_scripts/blob/master/scripts/Separate_XY_Positions.bsh fix the reading issue with BF/SCIFIO.

It needs to be tested on real data now :-)

marktsuchida:
I realize I'm joining the discussion late, but one idea would be to create a custom `Datastore` (let's call it `IndependentPositionDatastore`). You would implement this façade class by internally creating a new `DefaultDataStore` (obtained from `createMultipageTIFFDatastore()`) for each stage position.
This has the advantage of not having to modify the internals of the file writing code (which already could use some cleanup).

hadim:
That is what I am doing with XYSplitStorageMultipageTiff but I probably won't have time in a near future to put more effort on it since the beanshell script seems to worl well for now.

nicost:
OK to close?

hadim:
Yes

Issue #481, opened 2016-11-04
Create grid dialog box should remember overlap setting


kthorn:
It would be nice if the create grid dialog box remembered the overlap settings between openings.

PR #501, opened 2016-12-15
TileCreator: remembers overlap setting.

nicost:
Also handles application shutdown appropriately.  Addresses issue #481.

### DIFF
--- a/mmstudio/src/main/java/org/micromanager/internal/positionlist/TileCreatorDlg.java
+++ b/mmstudio/src/main/java/org/micromanager/internal/positionlist/TileCreatorDlg.java
@@ -38,13 +38,15 @@ import org.micromanager.MultiStagePosition;
 import org.micromanager.StagePosition;
 import org.micromanager.Studio;
 import org.micromanager.events.PixelSizeChangedEvent;
+import org.micromanager.events.ShutdownCommencingEvent;
 import org.micromanager.internal.utils.MMDialog;
 import org.micromanager.internal.utils.NumberUtils;
 import org.micromanager.internal.utils.ReportingUtils;
 
 public final class TileCreatorDlg extends MMDialog {
    private static final long serialVersionUID = 1L;
-   private CMMCore core_;
+   private final CMMCore core_;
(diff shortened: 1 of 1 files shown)


mmstudio/src/main/java/org/micromanager/internal/positionlist/TileCreatorDlg.java
```diff
@@ -364,6 +376,13 @@ public void actionPerformed(ActionEvent arg0) {
       studio.events().registerForEvents(this);
    }
 
+   @Subscribe
+   public void stuttingDown(ShutdownCommencingEvent se) {
```
marktsuchida, line 380: Is this needed, given that the profile is updated every time the values is edited?
Just checking, since if this is needed it might indicate an issue with the profile system.

nicost: Sorry, I do not understand the question.  Does this refer to the "registerForEvent", or to the "stuttingDown' (sic!) function?  The latter is needed to close the window on shutdown. Maybe you refer to something else?

marktsuchida: Sorry, it was off by one line (first time I tried GitHub's code commenting feature).
I was referring to the setting of `OVERLAP_PREF` in the profile at shutdown, right below.
(Though I'm also not sure what the `activateAxisTable()` accomplishes here.)

nicost: Ah!  The reason for the first line is that the actionEvent is not always fired when changing the value of the field (which is also why this field is read and the value stored when pressing the OK or Cancel buttons).  The call to activateAxisTable is superfluous and may even be dangerous since it is possible that positionLIstDlg_ is already disposed off.  I'll open another PR..

marktsuchida: Understood, thanks!

PR #511, opened 2017-01-09
add "Serializable" interface to PositionList

maierl:
Add `Serializable` marker interface to `PositionList` to allow transfer via Java RMI.
My local workaround extends the PositionList with a serializable subclass and works fine, so I don't expect any problems.

Keeping the warning about missing `serialVersionUID` deliberately, as the maintainers of micro-manager probably have a better strategy for this.

### DIFF
--- a/mmstudio/src/org/micromanager/api/PositionList.java
+++ b/mmstudio/src/org/micromanager/api/PositionList.java
@@ -45,7 +45,7 @@ import org.micromanager.utils.MMSerializationException;
  * Navigation list of positions for the XYStage.
  * Used for multi site acquisition support.
  */
-public class PositionList {
+public class PositionList implements Serializable {
    private ArrayList<MultiStagePosition> positions_;
    private final static String ID = "Micro-Manager XY-position list";
    private final static String ID_KEY = "ID";


marktsuchida:
Just to check, this is intended for Micro-Manager 1.4, correct?

I'm okay with adding this to 1.4 since there will never be any changes to `PositionList` there, but would be hesitant to do so in mm2, where we expect there to eventually be a better-designed data structure to replace `PositionList` (backward compatibility methods will be provided at such time). (I'm also not yet sure if it will be a good idea to add to the maintenance burden by declaring API classes `Serializable`.)

maierl:
Hi Mark, yes it is intended for Micro-Manager 1.4. At least here, the single interface line was enough for a short-lived serialization (obviously, the JSON representation makes more sense for archival). 

Does MM2 have any plan for IPC? For me, the client exists already and was written in Java, so Java RMI seemed the best choice.

If you keep the JSON export, maybe this could be used for serialization and keeping the maintenance low?

marktsuchida:
Committed from SVN.

I can see the utility of being able to use RMI. Using JSON serialization is probably a good idea in 2.0.

PR #524, opened 2017-02-04
Updated Localization Microscopy plugin

nicost:
Now uses MigLayout, many parts have been refactored.  No longer attempts to function with Micro-Manager present, makes full use of the 2.0 api.  Sorry for several commits that look to be duplicated, probably caused by old attempts to rebase.


marktsuchida:
@nicost The duplicate commit doesn't seem to actually be a duplicate (the two commits with the same message have different diffs), but it looks like two commits by Henry got mixed in here.
Would you mind amending the pull request?
1. Run
`git rebase -i 5e3ba95ef`
(that SHA1 being the commit where G3 branched off mm2) and remove the two lines (lines 4 and 5):
```
pick 362b4e440 Fixed bug where grid and surface tabs weren't enabled
pick 6b1acb3cd fixed default color bug
```
2. Do a forced push (`git push -f`) to your G3 branch on GitHub (this will update the PR)

(If you're worried about the pair of commits that have a duplicate commit message, you could also "reword" them during the `git rebase -i`.)

(I would be happy to take care of this, but if I make the change I'd have to either create a separate pull request or push manually; plus, it's considered bad practice to rewrite other people's history. Nice explanation I found: http://blog.adamspiers.org/2015/03/24/why-and-how-to-correctly-amend-github-pull-requests/)

nicost:
Done.  Thanks for the clear instructions!

marktsuchida:
It turns out that this doesn't compile, so reverted for now. Perhaps something got lost in a rebase?

Running `git rebase -i --exec 'make -C plugins' 5e3ba95ef` should stop at the first error, where you can fix the build and then `git commit --amend` and `git rebase --continue`.

nicost:
What was the error for you?  Did you update the 3rdpartypubli to use the new TSFProto.jar?  If not, you will get a single missing dependency in LoadAndSave.java.  Otherwise, I am not sure what I can do since it builds fine here.  B.t.w., TSFProto-SVN.jar is now hosted on Github (https://github.com/nicost/TSFProto/tree/mmGaussian), and the version in Micro-Manager is the one build from the mmGaussian branch (it contains extensions specific to Micro-Manager).  Now thinking about that, I probably should change the name of that jar to reflect this.

nicost:
I can reproduce the build problem when building from the command line.  It turns out that "ant fetchdeps.xml" indeed does not pick up the new version of TSFProto.jar, and therefore the dependency on a class provided in the new version of TSFProto.jar is not fulfilled.  It would be best to use https://github.com/nicost/TSFProto/blob/mmGaussian/java/target/TSFProto-mmg-1.0.0.jar (which should be identical to the TSFProto.jar that now is in 3rdpartypublic).   I am happy to make changes to the buildscripts, but would most likely break stuff, so it would be much better if you could do so.  Since TSFProto is now mavenized, I am also looking into making it available as a maven repository.

nicost:
Just realized that I probably should re-organize TSFProto a bit. I'll let you know when I am done.

marktsuchida:
Thanks, @nicost - I should have noticed that. The issue is that the JAR gets cached by Ivy (which we use in lieu of Maven to resolve dependencies) and, due to the way I configured Ivy for `3rdpartypublic`, it has no way of knowing that the file had changed, since it doesn't include a version number.

There are quick-and-dirty ways of fixing this for now, but I'll give it some more thought while you are working on TSFProto and will make sure to apply some fix when you are ready.

nicost:
Simple (temporary) solution: I added the source for TSFProto and TSFProto-mmg to the plugin code  itself, so that no external dependencies are needed.  I also started the process of adding both projects to Maven central.  When they are there, we can do it the right way, but for now it should be fine to merge in the G3 branch and delete the dependency on TSFProto.jar (if there is a stale version, that is no problem either, since I did change the path to the source to become maven compliant).

PR #526, opened 2017-02-04
Travis Support

hadim:
Here is a script for Travis. The build fails for now because of a weird svn bug I have.

Look at the Travis log here : https://travis-ci.org/hadim/micro-manager

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,61 @@
+language: java
+dist: trusty
+
+global:
+  - CASHER_TIME_OUT=500
+
+jdk:
+  - oraclejdk8
+
+addons:
+  apt:
+    packages:
+    - subversion
+    - build-essential
+    - autoconf
+    - automake
+    - libtool
+    - pkg-config
+    - libboost-all-dev
+    - zlib1g-dev
+      swig
+    - ant
+    - python-dev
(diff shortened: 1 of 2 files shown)


marktsuchida:
Thanks. Unfortunately, doing a full checkout of `3rdpartypublic` is probably never going to work (definitely not with git-svn, but it can be tricky even with the SVN client). Usually the fetching needs to be restarted a few times.

I've had to resort to various workarounds when setting up our automated builds. For this and other reasons I'm working on getting rid of the requirement for `3rdpartypublic` (replacing it with a combination of automated downloads and automated builds) and I'm pretty close now, so this problem should be solved soon.

hadim:
Ok good.

csachs:
Most of 3rdpartypublic does not seem to be needed for Linux builds, anyways.
Here's my take at Travis support, which downloads two jars from 3rdpartypublic web-svn:
https://github.com/csachs/micro-manager/blob/d0f3b699735bcc5de663f2bcc627cf618b68785c/.travis.yml
Which builds it:
https://travis-ci.org/csachs/micro-manager/builds/200435372
Code to push it somewhere for testing is missing, and I did not test if the resulting binaries work …

PR #527, opened 2017-02-08
Localization Microscopy plugin updated for MM 2.0

nicost:
It seems that the previous merge that was rolled back causes merge problems now.  I don't know why only the last 7 commits are shown.  Not sure how to fix that other than forcing the changes in this branch back onto mm2. I'll leave it up to you @marktsuchida


marktsuchida:
@nicost No problem. Would it be correct if I just resurrect the previous PR and then applied these 7 commits?

nicost:
@marktsuchida I think that should work.  The G3 branch contains all the changes to the plugin, and I have not re-based or done anything else to it other than adding a few more commits.  In fact, I do not really understand where these problems come from, probably because of my like in understanding what really happened when the PR was applied and reverted.

marktsuchida:
Ok, I'll go ahead and manually merge your G3 branch.

If you're curious, what happened is that when I first merged `G3` into `mm2`, a merge commit was created. That merge commit has two (symmetrical) parents: the previous commit on `mm2` and the then-last commit on `G3`. There is no way to truly revert a merge, so my "reversion" of the merge was just an ordinary commit on `mm2` that consisted of the diffs needed to get back to the original state of `mm2`.

Therefore, even after the revert, git "knows" that all the `G3` commits (except for the 7 newer ones) have been merged into `mm2` or, in other words, the current head of `mm2` has those commits as ancestors. This is why only the 7 newer commits are picked up as being in need of a merge.

The list of conflicts is then due to the conflict between the (single) reversion commit on `mm2` (between the older `G3` commits and the current head) and the lack of such a commit on `G3` (between the older `G3` commits and the 7 newer ones).

PR #528, opened 2017-02-10
Fixes for some typos in MMCore (mainly) comments

csachs:
During developing a device adapter, I've looked at the MMCore/ source for reference.
As my IDE highlights typos, I stumbled upon quite some of them.
In this PR I tried to fix all automatically detectable typos in these files' comments (first commit) - which should not alter behavior in any way, and few typos in error/log messages (second commit).
There's even a typo in the API `Camera::PrepareSequenceAcqusition` (acquisition being properly spelled in three other functions next to it ;)) … which should probably be addressed when incompatible API changes have to be introduced anyways.

### DIFF
--- a/MMCore/CircularBuffer.cpp
+++ b/MMCore/CircularBuffer.cpp
@@ -5,8 +5,8 @@
 //-----------------------------------------------------------------------------
 // DESCRIPTION:   Generic implementation of the circular buffer. The buffer
 //                allows only one thread to enter at a time by using a mutex lock.
-//                This makes the buffer succeptible to race conditions if the
-//                calling threads are mutally dependent.
+//                This makes the buffer susceptible to race conditions if the
+//                calling threads are mutually dependent.
 //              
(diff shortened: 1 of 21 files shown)


csachs:
Another commit, forgot to look into MMDevice. 

Another API typo which should be scheduled to be changed at some point: 
`g_Keyword_Meatdata_Exposure` in `MMDevice/MMDeviceConstants.h`. It's apparently not used anywhere in MM (it just repeats itself through SWIG generated code), still _if_ third-party code would rely on this spelling, it would be a breaking change…

marktsuchida:
Yep, I'm aware of the `Acqusition`. But fixing it would break source compatibility with existing device adapters, so it will have to stay. Same with the `Meatdata` (I hadn't noticed this one), although for this one we could allow both spellings.

Thanks for the corrections! I will apply these.

marktsuchida:
This has become SVN r16305-16307 and will later be merged into mm2.

Issue #516, opened 2017-01-26
Bad preprocessor directive in IIDC device adapters causing mm2 build to fail in Linux


kmdouglass:
# Problem description

Building mm2 for Linux x64 will outright fail due to a supposedly bad preprocessor directive in the file **MMIIDCCamera.cpp**. The offending line is number 245:

```c++
#   elif (defined(__clang__) && __has_builtin(__builtin_bswap16))
```

The relevant output from make is

```
Making all in IIDC                                                                      [6/1937]
make[2]: Entering directory '/home/douglass/src/micro-manager/micro-manager/DeviceAdapters/IIDC'
/bin/bash ../libtool  --tag=CXX   --mode=compile g++ -DPACKAGE_NAME=\"Micro-Manager\" -DPACKAGE_
TARNAME=\"micro-manager\" -DPACKAGE_VERSION=\"1.4\" -DPACKAGE_STRING=\"Micro-Manager\ 1.4\" -DPA
CKAGE_BUGREPORT=\"info@micro-manager.org\" -DPACKAGE_URL=\"\" -DPACKAGE=\"micro-manager\" -DVERS
ION=\"1.4\" -DSTDC_HEADERS=1 -DHAVE_SYS_TYPES_H=1 -DHAVE_SYS_STAT_H=1 -DHAVE_STDLIB_H=1 -DHAVE_S
TRING_H=1 -DHAVE_MEMORY_H=1 -DHAVE_STRINGS_H=1 -DHAVE_INTTYPES_H=1 -DHAVE_STDINT_H=1 -DHAVE_UNIS
TD_H=1 -DHAVE_DLFCN_H=1 -DLT_OBJDIR=\".libs/\" -DHAVE_BOOST=/\*\*/ -DHAVE_BOOST_THREAD=/\*\*/ -D
HAVE_BOOST_ASIO=/\*\*/ -DHAVE_BOOST_SYSTEM=/\*\*/ -DHAVE_BOOST_CHRONO=/\*\*/ -DHAVE_BOOST_DATE_T
IME=/\*\*/ -DHAVE__BOOL=1 -DHAVE_STDBOOL_H=1 -DSTDC_HEADERS=1 -DHAVE_MEMSET=1 -I.  -pthread -I/u
sr/include -DBOOST_THREAD_VERSION=2   -I/home/douglass/src/micro-manager/micro-manager/DeviceAda
[...]

PR #529, opened 2017-02-10
Added macro for (older) gcc compatibility

csachs:
See #516 .


marktsuchida:
Committed to SVN as r16308.
Appreciate the clean fix.

PR #531, opened 2017-02-13
OBIS/Coherent Scientific Remote support added

Raimooo:
Added new device adaptor to support the OBIS/Coherent Scientific Remote
for up to 6 lasers
(https://www.coherent.com/lasers/laser/cw-solid-state-lasers/obis-scientific-remote).
The adaptor is based on the already existing adator "CoherentOBIS" which
supports the OBIS/Coherent Single Laser Remote only. It would be great to have this device adaptor included in MM!

### DIFF
--- a/DeviceAdapters/CoherentScientificRemote/CoherentScientificRemote.cpp
+++ b/DeviceAdapters/CoherentScientificRemote/CoherentScientificRemote.cpp
@@ -1,0 +1,788 @@
+///////////////////////////////////////////////////////////////////////////////
+// FILE:          CoherentScientificRemote.cpp
+// PROJECT:       Micro-Manager
+// SUBSYSTEM:     DeviceAdapters
+//-----------------------------------------------------------------------------
+// DESCRIPTION:   Coherent Scientific Remote controller adapter for up to 6 Coherent Obis lasers
+//    
+// COPYRIGHT:     35037 Marburg, Germany
+//                Max Planck Institute for Terrestrial Microbiology, 2017 
+//
+// AUTHOR:        Raimo Hartmann
+//                Adapted from CoherentScientificRemote driver written by Forrest Collman
+//
+// LICENSE:       This file is distributed under the BSD license.
+//                License text is included with the source distribution.
+//
+//                This file is distributed in the hope that it will be useful,
+//                but WITHOUT ANY WARRANTY; without even the implied warranty
+//                of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+//
+//                IN NO EVENT SHALL THE COPYRIGHT OWNER OR
+//                CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+//                INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES.
+
+#ifdef WIN32
+   #include <windows.h>
+   #define snprintf _snprintf 
+#endif
+
+#include "CoherentScientificRemote.h"
+
+#include "../../MMDevice/MMDevice.h"
+#include "../../MMDevice/ModuleInterface.h"
+#include "../../MMDevice/DeviceUtils.h"
+
+#include <algorithm>
+#include <math.h>
+#include <sstream>
+#include <string>
+#include <cstring>
+
+
+
+// Controller
+const char* g_ControllerName = "Coherent-Scientific Remote";
+const char* g_Keyword_PowerSetpointPercent = "PowerSetpoint (%)";
+const char* g_Keyword_PowerReadback = "PowerReadback (mW)";
+const char* g_Keyword_TriggerNum = "Shutter Laser";
+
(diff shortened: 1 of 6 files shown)


marktsuchida:
@Raimooo Thank you for the PR! I'd like to include this in some form for sure. May I ask your rational for creating a separate device adapter rather than extending or replacing the existing `CoherentOBIS`? Should the existing adapter be considered obsolete once we have yours?

(If there is a way to allow existing .cfg files to work with the new adapter we might be able to avoid some duplication of code/logic.)

Raimooo:
Dear marktsuchida,
Thank you for considering this! The existing device adapter supports only the single laser remote fom OBIS/Coherent: https://www.coherent.com/lasers/laser/cw-solid-state-lasers/obis-single-laser-remote
In the new adapter the MM laser properties are dynamically created depending on what kind of lasers you attach to the remote. The old device adapter was static.
To use my adapter with the single remote actually does not require a lot of modifications but unfortunately I have none of theses devices for testing at the moment. The only difference while communicating with the single remote is that you don't need to address the laser (because only one can be connected, also in pricipal the carriage return character is different, but somehow in MM this does not matter). By the identification string of the device one could determine which kind of device is connected and then adapt the command syntax.
Best, Raimooo

nicost:
Your adapter was added to the Micro-Manager 1.4 branch, and should find its way into the 2.0 branch soon.  Can you add a page with device specific information on https://micro-manager.org/wiki/Device%20Support ?  Without it, people will never be aware of its existence (or know how to use it).

Thanks for your contribution, and sorry for the long delay in adding your code to the distribution!

raimoh:
No problem, thanks a lot! In the meanwhile we slightly adapted the code that the device adapter will also recognize the Coherent Single Laser Remote. The exisiting adapter "CoherentObis" can now be considered obsolete. Can you accept my user request on the micromanager site that I can add a description page?

nicost:
Done.  You can either email me the new code or open another pull request.

danielbholland:
@raimoh - This sounds very useful, thanks! Can you tell me if it will work with the OBIS LX/LS Laser Box for up to 5 lasers? 
(https://www.coherent.com/lasers/laser/obis-accessories/obis-laser-box) 
Thank again.

raimoh:
@danielbholland: We developed it for the  OBIS LX/LS Scientific Remote and also tested it with the OBIS LX/LS Single Laser Remote (when communicating with them both differ only in the carriage return character) . I cannot guarantee that it will work with the OBIS LX/LS Laser Box but I am pretty sure. The device adapter queries the attached lasers by itself - so in principal it does not matter whether you connect one, two or six. If you can send me the documentation for it I can probably tell you.

raimoh:
@ nicost: Thanks a lot, I opened another pull request.

nicost:
@raimoh: I could not find the pull request.  Did you indeed open one?

danielbholland:
@raimoh Thank you for your help! The Coherent website does not seem to have documentation about the serial communication for the 5-line laser box; I have requested it from Coherent and will post as soon as I receive it. In the meantime, in the chance that you are familiar with LabVIEW, I have attached to this message the LabVIEW 2012 examples Coherent had on their website for the laser box. Thanks again!
[OBIS-LabView-Examples-LV2012.zip](https://github.com/micro-manager/micro-manager/files/1405293/OBIS-LabView-Examples-LV2012.zip)

raimoh:
@nicost: No, I indeed did not, by mistake I send the request to my forked repository. Please have a look here: https://github.com/micro-manager/micro-manager/pull/585

@danielbholland: You are welcome. Unfortunately I don't have a running LabView installation to check. I just had a closer look at that device: It looks like that you can only turn the lasers on and off as the lasers only connect by a 12V power cable to the remote. So there is probably not way to query the lasers for wavelength, temperature and so on. I fear that the device adapter will return errors in that case and won't work.

danielbholland:
@raimoh I asked Coherent for documentation for the serial comm with the 5-line laser box and single laser remote, and asked if there was something like a 'programmers's manual'...They sent me the OBIS LX/LS "Operator's Manual" (attached, ~23 MB), though you probably already have this. It specifically notes (doc p. 12-2, PDF p. 222) that the OBIS 5-line laser box connects to the single SDR-type connector for power, I/O signals and communication. And in appendix C, on 'host interface', page C-7 (pdf page 283), they discuss device selection syntax, but no distinction is made for laser box vs. single laser vs. scientific remote. However, from discussion with other colleagues, I should note that the 5-line box will show up only as 1 COM port to device manager, individual lasers aren't visible as separate COM port numbers. Only in the Coherent software do you see lasers listed as being connected to COM X:Y, where the ":Y" is an additional identification number. Does any of this info help you guess about compatibility? Thanks again! 
[Coherent_OBIS_LX_LS_1184163_RevAG.pdf](https://github.com/micro-manager/micro-manager/files/1408060/Coherent_OBIS_LX_LS_1184163_RevAG.pdf)

PR #538, opened 2017-03-08
Modbus DeviceAdapter using libmodbus

csachs:
A little thing I've been working on. A DeviceAdapter (BSD-licensed) allowing [Modbus](https://en.wikipedia.org/wiki/Modbus)-connected devices to be controlled, via [libmodbus](http://libmodbus.org/) (LGPL licensed).

We're using this to control a valve system controlling a microfluidic setup (valve hardware setup as designed by Dr. Christopher Probst (@chrisp87), Thanks!).

Modbus allows for data I/O via two data types: binary "coils", and 16-bit wide "registers" holding arbitrary data.
As for our setup only "coils" are currently necessary, only access to these has been implemented and tested, however registers can easily be added, and I might work on this in the future…
Nonetheless, it works quite well for our purposes and could serve as a general starting point for a modbus device adapter in MM, allowing access to a plethora of modbus-controlled periphery.

For usage, on adding the adapter, the user sets the connection by a connection URI, and defines modbus devices by a configuration string (e.g. `name=valves type=coils write_address=0 read_address=512 count=8`), and after the device adapter has initialized, has the properties they named at their disposal for IO with the modbus device.

For building it under *NIX, I've added the (hopefully) appropriate lines to the autotools files to auto detect a distribution installed libmodbus; for Windows, unfortunately no official binary builds exist. It is however straightforward to `git clone` it into the `lib` subfolder, build it there according to its documentation, and then build the VS project. (If you want to add it to MM, you could probably build libmodbus once, and add the binary to 3rdpartypublic, and change the links accordingly).

If you want to include this adapter in MM, I can write documentation for it on the Wiki.

What do you think?
Best regards,
Christian

### DIFF
--- a/DeviceAdapters/Makefile.am
+++ b/DeviceAdapters/Makefile.am
@@ -29,6 +29,9 @@ endif
 if BUILD_OPENCV
    OPENCVGRABBER = OpenCVgrabber
 endif
+if BUILD_MODBUS
+   MODBUS = Modbus
+endif
 if BUILD_PVCAM
    PVCAM = PVCAM PrincetonInstruments
 endif
@@ -68,6 +71,7 @@ SUBDIRS = \
 	$(IIDC) \
 	$(ITC) \
 	$(NIDAQ) \
+	$(MODBUS) \
 	$(OPENCVGRABBER) \
 	$(PVCAM) \
 	$(QCAM) \
--- a/DeviceAdapters/Modbus/Makefile.am
+++ b/DeviceAdapters/Modbus/Makefile.am
@@ -1,0 +1,8 @@
+AM_CXXFLAGS = $(MMDEVAPI_CXXFLAGS) $(MODBUS_CPPFLAGS)
+deviceadapter_LTLIBRARIES = libmmgr_dal_Modbus.la
+libmmgr_dal_Modbus_la_SOURCES = ModbusModule.cpp ModbusModule.h ../../MMDevice/MMDevice.h
+libmmgr_dal_Modbus_la_LDFLAGS = $(MMDEVAPI_LDFLAGS) $(MODBUS_LIBS)
+libmmgr_dal_Modbus_la_LIBADD = $(MMDEVAPI_LIBADD)
+
+EXTRA_DIST = Modbus.vcproj
+
--- a/DeviceAdapters/Modbus/Modbus.vcxproj
+++ b/DeviceAdapters/Modbus/Modbus.vcxproj
@@ -1,0 +1,180 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Project DefaultTargets="Build" ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <ItemGroup Label="ProjectConfigurations">
+    <ProjectConfiguration Include="Debug|Win32">
+      <Configuration>Debug</Configuration>
+      <Platform>Win32</Platform>
+    </ProjectConfiguration>
+    <ProjectConfiguration Include="Debug|x64">
+      <Configuration>Debug</Configuration>
+      <Platform>x64</Platform>
+    </ProjectConfiguration>
+    <ProjectConfiguration Include="Release|Win32">
+      <Configuration>Release</Configuration>
+      <Platform>Win32</Platform>
+    </ProjectConfiguration>
+    <ProjectConfiguration Include="Release|x64">
+      <Configuration>Release</Configuration>
+      <Platform>x64</Platform>
+    </ProjectConfiguration>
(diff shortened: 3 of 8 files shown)


csachs:
It's has been some months since opening the PR, and I still wonder whether this could be generally eligible to be added to MM, or not. I'd be happy to hear your feedback.
Best regards, Christian

nicost:
Hi Christian,  Sorry for the lack of response.  I tried to build libmodbus on Windows, but failed with: Microsoft.Build.Exceptions.BuildAbortedException: Build was canceled. Failed to successfully launch or connect to a child MSBuild.exe process. Verify that the MSBuild.exe "C:\Windows\Microsoft.NET\Framework\v4.0.30319\MSBuild.exe" launches successfully, and that it is loading the same microsoft.build.dll that the launching process loaded. If the location seems incorrect, try specifying the correct location in the BuildParameters object, or with the MSBUILD_EXE_PATH environment variable.

Do you have binaries for 32 and 64 bit?  That would greatly reduce the amount of effort needed to add this to the build.

csachs:
Hi, great to hear back!
Building stuff for Windows is a pain ;) I've developed the PR with a libmodbus version I twiddled together, but that's kinda not an option in the long term. I've as such investigated creating automated libmodbus-builds via the AppVeyor CI service, and apparently managed to get binaries.
I've opened PR https://github.com/stephane/libmodbus/pull/390 at the libmodbus project; meanwhile you can try the binaries at my fork:
(ZIP Archives containing the `.dll`, `.lib`, `modbus.h` etc.)
https://ci.appveyor.com/project/csachs/libmodbus/build/job/1774gfinhq88vlv5/artifacts (32 bit)
https://ci.appveyor.com/project/csachs/libmodbus/build/job/jnhrgumj1gojrnxs/artifacts (64 bit)

As I'm currently home I have so far not tested them, I'll try to do this (atleast some in silico tests) when I get to the institute tomorrow.
Best regards, Christian

nicost:
Thanks for those.  I made a little bit of progress.  I had to delete the stdint.h from the directories you send me (otherwise, VS2010 would not find the uint8_t and uint16_t types).  Now, compilation ends because "modbus-tcp.h" and "modbus-rtus.h" are missing.  Are those available somewhere?

csachs:
Hi, sorry! I've skimmed thru `modbus.h` to see if any other files are included, but didn't spot the `-rtu`/`-tcp`... yes they are just part of libmodbus; I've changed the AppVeyor build definition and the new ZIPs include them now:

https://ci.appveyor.com/project/csachs/libmodbus/build/job/wmknulndheoo8vw6/artifacts (32 bit)
https://ci.appveyor.com/project/csachs/libmodbus/build/job/0mkpu4gd8edb2iwv/artifacts (64 bit)

Didn't find time to test the build with the `.dll`/`.lib`s. And yes, the `stdint.h` situation is expected, older VS versions didn't contain it, new versions bring their own.

nicost:
Great!  I added your adapter code to both the Windows and Unix build systems, so it should appear in the next nightly build.  However, the Mac build machine will need libmodbus to be installed (simple since homebrew provides it), and the dlls will need to be added to the distribution on both Mac and Windows.  Mark will need to do that, since I completely forgot how to.  

Can you add a page to the Micro-manager website (https://micro-manager.org/wiki/Device%20Support) describing what this is and how to use it?  See any of the other device description pages as an example.  

Thanks for your contribution, and sorry for the delay in adding this to the repository and build!

csachs:
Thanks a lot! I'll try to test the Windows version next week in our lab. I'll write some documentation and add it to the wiki.

csachs:
Just looked at the recent nightly builds (20170709, 1.4 and 2.0, 32 and 64 bit), and it does not seem to be included? (By the way, was it just added to 1.4? I'd greatly appreciate 2.0 as well, as I'm currently using 2.0 for developments…)

nicost:
Indeed, it is not there.  I do not know why not.  Pinging @marktsuchida...

Deviceadapters first have to go into 1.4.  They are then automatically propagated into 2.0.

csachs:
Any update why it is not in the build?

nicost:
I was wondering about that as well.  Just noted some missing tags in the solution file and pushed thos to the subversion repository.  Let's look what happens in tonight's build....

nicost:
After adding the missing tags, the build includes modbus.  However, I forgot to actually commit the needed libraries to 3rdpartypublic.  Those are now added, so hopefully tonight's build will finally include the modbus adaptor!

csachs:
Short remark; I've seen the device adapter is now present in the build, but I'm currently very busy (paper/thesis). I'll try to test it next week on our hardware, and write some documentation in the wiki.

PR #572, opened 2017-06-21
Add DoubleVector to enable loadXYStageSequence()

daferna:
loadXYStageSequence() requires a vector<double>

lines 203-209 needed to be moved down because SWIG %apply works indiscriminately and will match types inside the source code for std::vector causing failure to compile. It may be better in the future to use %apply and then %clear, or use a more specific name & type matching, but this works for now.

This bug was discovered by @fcollman and myself at the Allen Institute for Brain Science.

### DIFF
--- a/MMCorePy_wrap/MMCorePy.i
+++ b/MMCorePy_wrap/MMCorePy.i
@@ -55,16 +55,6 @@ import_array();
 #include "string.h"
 %}
 
-
-// output arguments
-%apply double &OUTPUT { double &x };
-%apply double &OUTPUT { double &y };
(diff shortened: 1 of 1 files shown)


nicost:
I just manually added your changes to the 1.4 branch (from where it should propagate into 2.0).  Can you check tonight's (or later) build and check that everything works as expected?  Thanks!

nicost:
Assuming that all works as expected...

fcollman:
Looks like this change never made it into the 2.0 branch.

nicost:
Manually pushed it in.

PR #576, opened 2017-07-08
Updated PointGrey device adapter with additional trigger functionality

mjc449:
-Added camera-specific detection of trigger properties including:
		-Available modes
		-Available sources
		-External trigger presence (through source detection)
		-Polarity presence
	-Added MM properties for each of the above as well as user-defined
		external grab timeout property
	-Added trigger on/off property for switching between internal
		and other sources
	-Fixed bug causing MM crash when loading device that was left in
		external trigger mode
		-Cameras supporting external triggers are now loaded in
			SW trigger mode
	-Fixed bug causing sequence acquisitions to reset to internal trigger

### DIFF
--- a/DeviceAdapters/PointGrey/PointGrey.cpp
+++ b/DeviceAdapters/PointGrey/PointGrey.cpp
@@ -55,9 +55,23 @@ const char* g_AdvancedMode             = "Use Advanced Mode?";
 const char* g_VideoModeAndFrameRate    = "Video Mode and Frame Rate";
 const char* g_NotSet                   = "Not set";
 const char* g_Format7Mode              = "Format-7 Mode";
-const char* g_InternalTrigger          = "Internal";
-const char* g_ExternalTrigger          = "External";
-const char* g_SoftwareTrigger          = "Software";
+const char* g_TriggerGPIO0             = "GPIO 0";
+const char* g_TriggerGPIO1             = "GPIO 1";
+const char* g_TriggerGPIO2             = "GPIO 2";
+const char* g_TriggerGPIO3             = "GPIO 3";
+const char* g_TriggerSoftware          = "Software";
+const char* g_TModeStandard            = "Standard";
+const char* g_TModeExposure            = "Exposure";
+const char* g_TModeSkipN               = "Skip N Frames";
+const char* g_TModeMultipleExposure    = "Multiple Exposure";
+const char* g_TModeMultiplePulseWidth  = "Multiple Exposure Pulse Width";
+const char* g_TModeLowSmear            = "Low Smear";
+const char* g_TModeOverlappedReadout   = "Overlapped Readout";
+const char* g_TModeMultipleShot        = "Multiple Shot";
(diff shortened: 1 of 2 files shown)


nicost:
The diffs show mainly changes in white space and indentation, making it very difficulty to figure out what the actual changes/additions are.  Can you please revert the white space/indentation to the original, and recommit (so that github will show only the relevant changes)?

Also, I really do want to keep the nomenclature "Software, Internal, External" for trigger modes (additional ones are fine, but these three should be in there).

mjc449:
No problem, I'll clean up the formatting.  I think that the problem is VS "normalized line endings" to Unix \n, which I have as the default, so I'll have to figure out how to undo that if that's the cause.  I also setup tabs as 3 spaces, per the style guide and tabulated both .h and .cpp files to 3 spaces, which is probably also causing a lot of the diffs.  I may just have to go through and cut and paste the changes into a fresh clone.

As far as the trigger mode labels, I changed it to source to be inline with the FlyCapture2 SDK.  They designate sources as what you had labeled modes, and modes are a different parameter that controls the operation of the trigger, and are independent of source (except the internal trigger).  I can rename source to mode, and call the SDK modes triggerType if you want.  I think this would be more obvious to the user, but it'll make the source more confusing to read.  As far as the external triggers go, I'll rename them 1-4  to correspond with the GPIO pin numbers-1 that they refer to.  I'll also add an "Internal" trigger back in that'll just turn off the triggers.

nicost:
Thanks, and sorry to ask you to do the work!  I do prefer that the user facing nomenclature stays the same, otherwise people's configuration files will break, which is annoying and unnecessary.  I think that the cleanest way to deal with it is to follow the Point Grey nomenclature in the code (i.e. have OnTriggerSource, and OnTriggerMode propertyaction functions that do what PointGrey wants them to do), but to label them for the use as "TriggerMode" (for OnTriggerSource), and "TriggerType (for "OnTriggerMode").  A comment to explain will be useful!

mjc449:
I don't mind, the reformatting isn't that big of a deal.  I probably won't have time to get to it for a day or two.  As far as nomenclature I think it'll be easiest to keep your labels, then add another property to select which GPIO pin is the trigger input.  That's a pretty simple solution in terms of code and should prevent breaking configurations.  I am a bit ignorant about config files, are they just saving the text strings in regard to the property name and values?  Also, I'm assuming that adding properties to the adapter won't break configs, is this the case?  I've never re-used a config on different builds.

nicost:
Yes, config files are saving key-value pairs as text strings (open one up, it should be quickly obvious how they work).  Adding properties indeed does not break existing configurations.  Thanks!

nicost:
It would still be nice to pull your changes.  Did you have a chance to clean up the formatting?

mjc449:
I'm sorry about the delay, I got started on it but had some other stuff take priority.  I'll sit down with it this weekend and get it cleaned up.

nicost:
Since so much time has passed, closing for now.  Feel free to re-open when you think it is useful.

PR #578, opened 2017-07-14
ThorlabsAPTStage: Fixed MM2.0 specific crash 

seamusholden:
Fixed MM2.0 specific crash by requiring APT re-initialization in device Initialize() method.
APT is initialized in InitializeModuleData() but the device initialization does not persist in MM2.0 for some reason.
Manually reset the "aptInitilized" flag back to false at end of InitializeModuleData(), then everything works fine.


nicost:
Thanks for fixing this Seamus!  Is there any downside to putting this in 1.4 as well?  Asking since we are keeping the source trees for both for the device adapters identical.

seamusholden:
No problem!  I tested it on 64-bit 2.0 and a recent 1.4 build so go ahead
and integrate into both. Not tested on 32-bit but I would be really shocked
if it causes a bug there very minor change really.

nicost:
Fix committed to mm1.4 in svn commit 16432 (https://valelab4.ucsf.edu/trac/micromanager/changeset/16432).  The change should propagate soon to 2.0.

PR #579, opened 2017-08-09
Fix 'ordered comparison between pointer and zero'

JiangXL:
When I try to compile Micro-manager, I trouble into `  ordered comparison between pointer and zero ('usb_dev_handle *' and 'int')`. Then I find same error in [here](https://github.com/ZoneMinder/ZoneMinder/issues/1760). It work. 

Error: 

      Making all in USBManager
      make[2]: Entering directory '/home/igem/Downloads/mm/micro-manager/DeviceAdapters/USBManager'
      /bin/sh ../libtool  --tag=CXX   --mode=compile clang++ -DPACKAGE_NAME=\"Micro-
       Manager\" -DPACKAGE_TARNAME=\"micro-manager\" -DPACKAGE_VERSION=\"1.4\" -
       DPACKAGE_STRING=\"Micro-Manager\ 1.4\" -DPACKAGE_BUGREPORT=\"info@micro-
       manager.org\" -DPACKAGE_URL=\"\" -DPACKAGE=\"micro-manager\" -
        DVERSION=\"1.4\" -DSTDC_HEADERS=1 -DHAVE_SYS_TYPES_H=1 -
        DHAVE_SYS_STAT_H=1 -DHAVE_STDLIB_H=1 -DHAVE_STRING_H=1 -
        DHAVE_MEMORY_H=1 -DHAVE_STRINGS_H=1 -DHAVE_INTTYPES_H=1 -
        DHAVE_STDINT_H=1 -DHAVE_UNISTD_H=1 -DHAVE_DLFCN_H=1 -DLT_OBJDIR=\".libs/\" 
        -DHAVE_BOOST=/\*\*/ -DHAVE_BOOST_THREAD=/\*\*/ -DHAVE_BOOST_ASIO=/\*\*/ -
        DHAVE_BOOST_SYSTEM=/\*\*/ -DHAVE_BOOST_CHRONO=/\*\*/ -
        DHAVE_BOOST_DATE_TIME=/\*\*/ -DHAVE__BOOL=1 -DHAVE_STDBOOL_H=1 -
        DSTDC_HEADERS=1 -DHAVE_MEMSET=1 -I.  -I/usr/include/libusb-1.0  -
         I/home/igem/Downloads/mm/micro-manager/DeviceAdapters/../MMDevice -pthread -
         I/usr/include  -g -O2 -MT USBManager.lo -MD -MP -MF .deps/USBManager.Tpo -c -o 
         USBManager.lo USBManager.cpp
[...]

PR #580, opened 2017-08-09
Fixed 'ordered comparison between pointer and zero'

JiangXL:
Same error with Issue  #579


nicost:
Committed to 1.4 in svn commit 16451.  Will propagate to 2.0.

PR #581, opened 2017-08-11
fix focus scoring to not modify images in-place

jondaniels:
Fix bug in 2 focus scoring algorithms that I added.  Images were modified instead of copied before modifying.  Not noticed before now b/c usually images are discarded after scoring.

I'm not set up to compile this code but the changes are straightforward.


nicost:
Applied to the subversion repository of 1.4.

PR #585, opened 2017-10-23
Coherent Scientific Remote adapter update

raimoh:
- added support for the Single Laser Remote
(https://www.coherent.com/lasers/laser/optoskand-beam-delivery-components/obis-single-laser-remote)
- existing device adapter "CoherentOBIS" may now be considered to be
obsolete

### DIFF
--- a/DeviceAdapters/CoherentScientificRemote/CoherentScientificRemote.cpp
+++ b/DeviceAdapters/CoherentScientificRemote/CoherentScientificRemote.cpp
@@ -3,7 +3,8 @@
 // PROJECT:       Micro-Manager
 // SUBSYSTEM:     DeviceAdapters
 //-----------------------------------------------------------------------------
-// DESCRIPTION:   Coherent Scientific Remote controller adapter for up to 6 Coherent Obis lasers
+// DESCRIPTION:   Coherent Scientific Remote laser controller adapter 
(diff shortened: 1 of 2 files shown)


nicost:
Committed to the subversion repository in commit #16599.  Please note that I reverted line 357.  The variables in that function are not used.  If you leave them un-commented, the compiler will rightfully issue warnings.  Too many warnings lead to the important warnings becoming invisible.   

Can you please confer with Forrest about retirement of the other Coherent adapter?  At the very least removing that adapter from the build will force users to redo their configuration files, which is kind of annoying for them.  I forgot if there is a way to no longer list the old device adapter in the Hardware Configuration Wizard, as that would be the preferable solution.

nicost:
Just helped someone at UCSF who was using this adpater and had problem re-opening configuration files that had channel definitions with properties coming from this adapter.  It turned out that the laser name contained and eol character (either \r or \n, not sure), that was faithfully reproduced in the config file and broke our code reading these back in.  This eol character is coming from the hardware itself (the lasername ands with it).  I added code to check for this and strip it when found.  I think that this will not break anything for people who do not have this problem (it could well depend on the firmware version used).  I also realize that MMCore should probably protect everyone from running into this problem, but for now I think this solution is fine.

PR #588, opened 2017-10-29
Video4Linux: also use colors and output RGBA32

dedeibel:
Hi,

I was trying to get my cheap usb microscope running on linux and was wondering why only grayscale display was working.
Digging into the code it seemed like the correct input format was used but only the luma values were copied into the output image. [line 354](https://github.com/micro-manager/micro-manager/compare/mm2...dedeibel:feature/video4linux2_with_color?expand=1#diff-1c7389b8b9b38ea61f702c3342af5b44L354)

I did not reduce any compatibility with this since the input parameters are unchanged. So I consider the risk pretty low.

Maybe you want to integrate this, seems like the v4l2 Adapter is not the most popular anyway. If not maybe this serves as a solution for others to find my branch.

_(Looking into other drivers it seems like mm does handle RGBA32 output the best – had figure this out in a trial an error way and convert the input. Is this documented somewhere? I could not find it here [wiki/Building_Micro-Manager_Device_Adapters](https://micro-manager.org/wiki/Building_Micro-Manager_Device_Adapters), might save people writing adapters some time.)_

Thank you, Ben

### DIFF
--- a/DeviceAdapters/Video4Linux/video4linux2.cpp
+++ b/DeviceAdapters/Video4Linux/video4linux2.cpp
@@ -300,7 +300,7 @@ VideoInit(State*state)
     // Pixel type
     pAct = new CPropertyAction (this, &V4L2::OnPixelType);
     nRet = CreateProperty(MM::g_Keyword_PixelType,
-			  "8bit", MM::String,true, pAct);
+			  "YUYV", MM::String,true, pAct);
     if (nRet != DEVICE_OK)
       return nRet;
     // Gain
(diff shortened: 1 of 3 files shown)


nicost:
Hi Ben, thanks for the contribution.  Even though I do not currently use linux myself, this should be useful to several people.  Would it be possible to modify the code such that the old behavior is still possible (if one uses one of these cameras for fluorescence microscopy, it is desirable to get monochromatic output).  For instance, you could leave the existing 8-bit mode in place, add YUYV, and return buffer and buffer size according to that setting.  That would add to the existing behavior, and not take something away that someone may like.  Thanks!

dedeibel:
Hi, should be possible but may take a while.

dedeibel:
Strangely I am now having problems with NullPointerExceptions in the Java Parts when I provide the 4 bytes per pixel image. The code worked fine based on the older revision. I need to investigate more, closing the PR for now.

PR #589, opened 2017-10-30
Added FakeCamera device adapter that is backed by the filesystem

lukas-lang:
Since there seem to be a few people in need of something like this, I thought I'd share my take on a camera that can provide arbitrary images.

### What it does
This is an implementation of a camera device adapter with the following features:
* Supports most common file formats (basically anything supported by OpenCV's [`cv::imread`](https://docs.opencv.org/2.4/modules/highgui/doc/reading_and_writing_images_and_video.html?highlight=imread#imread))
* Supports 8 & 16bit grayscale and 32 & 64bit RGB (although 64bit RGB seems to be broken on micro-managers side)
* The specified path mask can contain placeholders that are replaced by the position of the specified stages. A quick overview of the format:
  * `??` is the position of the focus stage, rounded down to micrometres
  * `?{<prec>}?` is the position of the focus stage, rounded to the specified precision
  * `?[<stage-name>]` is the position of the specified stage, rounded down to micrometres.
  * `?{<prec>}[<stage-name>]` is the positon of the specified stage, rounded to the specified precision

So, for example: `C:\imgs\img_?{1}?_?[x-Axis].png` would be interpreted as `C:\imgs\img_2.1_4.png` assuming stage positions of e.g. `2.1um` for the focusing stage and e.g. `4.3um` for the stage named `x-Axis`.

### Remarks
I didn't know what is needed in terms of build scripts and binaries, so I only included the source files and the VS project files - I hope that's ok


nicost:
Thanks!  This idea is great (we talked about this internally for years!).  I added the code to 1.4 repository, and included it in the Mac/Linux builds (had to make a few changes to the code to avoid compiler errors and warnings, please check if it still works as expected).  Will check the build on a Windows machine and add the build files there soon.  
My Micro-Manager version on my Mac does not find the OpenCV libraries.  I completely forgot how all that works, if you happen to have any pointers, that would be great, otherwise, it will be most important this actually works on Windows.

nicost:
Now added this adapter to the Windows build, but I don't know how to get it to work (i.e. what OpenCV stuff is needed, configuration file to make, etc..).  Do you mind writing a short how to on the Micro-Manager website at https://micro-manager.org/wiki/Device%20Support (you will probably need to request an account first).

lukas-lang:
Thanks for adding it!

What kind of problems do you encounter on Windows? I downloaded the latest nightly build on a non-dev machine and it worked without any issues. (If you're building it with VS, you need to use the Release configuration or else you need to copy the OpenCV debug dlls from `3rdpartypublic\OpenCV2.4.8\build\x64\vc10\bin` to the MM installation folder)
Regarding Linux:
I managed to build it (but not to run it - apparently, the OpenCV libraries are not added to the base folder of MM as they are on Windows) by following the guide [here](https://micro-manager.org/wiki/How_To_Build_Micro_Manager_From_Source_On_Ubuntu_13.10), where I added the `--with-opencv` option to the `./configure` call. (you need to install OpenCV manually beforehand, e.g. following [this guide](https://docs.opencv.org/2.4/doc/tutorials/introduction/linux_install/linux_install.html)). I added the required `build.xml` and `Makefile.am`, see the latest commit (I also added your fixes in there)

nicost:
I actually did not test on Windows.  Will try tonight's build (tonight;)

The code builds fine on the Mac, but the adapter does not appear in the Hardware configuration Wizard (presumably due to missing libraries, but I have no clue which ones to add...).

Also, if you plan on making more changes, it will be easier if I send you (off-github) an svn account.  Otherwise, since the svn repository is the "source" for the device adapters, I will need to manually transfer you changes.

lukas-lang:
Regarding the svn account: I'd be glad to commit it there directly if that's easier for you.
Regarding the libraries: It seems like OpenCVGrabber (the other device adapter that uses OpenCV) simply contains the OpenCV dlls (`opencv_core248.dll` and `opencv_highgui248.dll`) inside the `lib` folder and copies them to the install directory upon build (at least for the Windows installer).

I don't know what the "proper" way to handle something like this would be on other platforms - do you think it's possible to do the same for macOS as is done for Windows? I don't know how portable the macOS equivalent to dll files is, so this might not work.

By the way, I managed to get it to work on Linux (although the RGB mode doesn't work for me) - it could just be that I needed to restart after installing OpenCV (if my account gets approved, I can try to write a setup guide for the wiki so someone else can test it on a clean system)

nicost:
Trying to remember how these things work on the Mac;)  Tried copying the opencv dylibs to the Micro-Manager directory, but that did not help.  I guess that the opencv libs should be on the PATH.  In the past, I used static linking to make it easy to have the binaries work, but that is a bit of an ugly fix.  Maybe OK if it only works on Windows.

nicost:
Added to 1.4.  Should work its way into 2.0 eventually.

csachs:
Oh, this adapter appeared quickly and is already in … but just a short question:

We had this discussion on the mailing list, and for certain use cases (e.g. providing highly dynamic images, such as crops of large slide scans matching the current stage position), my approach pulling images from a separate HTTP server has certain upsides.

So what do you think:

- If it's OK to have another highly specific testing camera adapter, I'd just open another PR ( https://github.com/csachs/micro-manager/tree/SimulatingCamera/ ).
- If camera-feeding mechanisms should be bundled, would it be OK if I'd ported over my parts to this adapter?

Btw, is there any planned process towards having a github-based workflow? Moving everything back and forth between git and svn seems to be a lot of hassle. (I know the non-public device adapters need to stay out of public git, nonetheless the rest could probably profit from github)

Best regards,
Christian

Mathe172:
I thought about this a bit as well, since it would certainly be nice to have a more dynamic option available. Do you think it would be enough to simply allow URLs to be specified as path mask? This could then use the same placeholder specifications, but instead of loading the images from disk, they would be pulled from a server.
The only question that remains in that case is how image size should be handled. Should it be added to the header of the image that is provided by the server?

lukas-lang:
I thought about this a bit as well, since it would certainly be nice to have a more dynamic option available. Do you think it would be enough to simply allow URLs to be specified as path mask? This could then use the same placeholder specifications, but instead of loading the images from disk, they would be pulled from a server.
The only question that remains in that case is how image size should be handled. Should it be added to the header of the image that is provided by the server?

(sorry for posting twice, but the last one was sent from the wrong account)

nicost:
As for workflow:  There are no plans to move 1.4 from svn, so as long as 1.4 is maintained, and the device-adapter code remains the same between the two versions, we will keep the subversion repository as the primary and propagate changes to github.  That also is less work for me;)

csachs:
@lukas-lang 
It's currently implemented differently in my adapter: The image size is specified akin to the DemoCamera adapter, and the URL then contains it as a query parameter. It's assumed that the server responds with exactly the right binary data.

I thought as well that the easiest way would be to use the Path Mask (possibly renamed) as the entry value for that, and just divert logic whether it starts with "http://" or not.

Concerning the server side, as aforementioned, currently image dimensions are sent along the other parameters. That could be reversed. I actually devised a specification to pass around raw image data with dimension/datatype information in the content-type header some time ago ( https://github.com/csachs/rioh ) which I could implement in C++.
It'd certainly a possibly cleaner way to have the server send a defined image size (the server choses), and the DeviceAdapter just consume that. The upside of a header-based approach would be, that standard HTTP servers could be used to serve standard image files, and they could be decoded in memory using `cv2::imdecode` in the adapter.

One thing I noticed in your adapter is that apparently it currently only supports focus stages.
To allow for versatile microscopy simulation, support for x/y stages is necessary, as well as for state devices (virtual filter wheels => fluorescence channels, possibly different objectives => magnification sizes).
If a way to escape `?` is added, it could be used to build URLs (query parameters) easily, tho.

mutterer:
Dear all,
for what it's worth, I just shared a slightly similar project that
addresses @csachs suggestions to some extent.
It's not a fake camera, but an on-the-fly image processor that was written
to work with the Demo configuration of MM and replaces images from the Demo
Camera with portions of a built-in multichannel image. I use it for
demonstrating MM features in trainings in the absence of a microscope. It
works with demo channels, objectives and stages.
It's definitely not as advanced as the fake camera device adapter, but I
wanted to advertise it as microscopy simulation was mentioned in this
thread.

You can get it with instructions at :
https://github.com/mutterer/MM-Demo-Sample
I'm happy for any feedback, suggestions or code contribution.

Best.

Jerome.


2017-11-09 10:27 GMT+01:00 Christian Sachs <notifications@github.com>:
>
> @lukas-lang
> It's currently implemented differently in my adapter: The image size is
specified akin to the DemoCamera adapter, and the URL then contains it as a
query parameter. It's assumed that the server responds with exactly the
right binary data.
>
> I thought as well that the easiest way would be to use the Path Mask
(possibly renamed) as the entry value for that, and just divert logic
whether it starts with "http://" or not.
>
> Concerning the server side, as aforementioned, currently image dimensions
are sent along the other parameters. That could be reversed. I actually
devised a specification to pass around raw image data with
dimension/datatype information in the content-type header some time ago (
https://github.com/csachs/rioh ) which I could implement in C++.
> It'd certainly a possibly cleaner way to have the server send a defined
image size (the server choses), and the DeviceAdapter just consume that.
The upside of a header-based approach would be, that standard HTTP servers
could be used to serve standard image files, and they could be decoded in
memory using cv2::imdecode in the adapter.
>
> One thing I noticed in your adapter is that apparently it currently only
supports focus stages.
> To allow for versatile microscopy simulation, support for x/y stages is
necessary, as well as for state devices (virtual filter wheels =>
fluorescence channels, possibly different objectives => magnification
sizes).
> If a way to escape ? is added, it could be used to build URLs (query
parameters) easily, tho.
>
> —
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly, view it on GitHub, or mute the thread.




--
Jerome Mutterer
CNRS - Institut de biologie moléculaire des plantes
12, rue du Général Zimmer
67084 Strasbourg Cedex
www.ibmp.cnrs.fr

lukas-lang:
@csachs: Putting the size and bit-depth information in the header seems like a reasonable idea. 

Regarding the path mask format: Building query parameters shouldn't be a problem, we just need to teach the parser to ignore single `?` - there should never be any ambiguity this way (Since neither `??`,`?{` nor `?[` are valid parts of an URL).  And adding support for more device types should also be more or less straightforward. The only issue is that we need a way to specify the axis of a xy-stage - do you think something like `?{<prec>}(<x|y>)[<xy-stage-name>]` would work?

PR #590, opened 2017-11-01
Leica driver fixes

fcollman:
This makes a number of changes to the Leica driver that fixes issue with the AFC, and the fast filter wheel which we are using at the Allen Institute with our DMI-8 systems.  This will make the fast filter wheel work if it in the only one installed on the system.  Command 22 the the fast filter wheel can't properly set the busy flag of the device to not busy because the response does not indicate which filter wheel was told to change position.  This is a flaw in the Leica API, I'm complaining to Leica, but this at least works for 1 filter wheel systems, where the old driver wouldn't work at all.


nicost:
Applied to 1.4 repository.  Should be propagating to 2.0 (needs manual merge by Mark).  Thanks!

PR #591, opened 2017-11-01
Add TCPIPPort device adapter

lukas-lang:
### What is this?
This device adapter can be used as serial port and forwards all communication to a server over TCP/IP. It can for example be used with a program like [this one](https://github.com/lukas-lang/COMoverTCP) to connect to a COM port on another computer.

### Multiple TCPIPPorts
You can add multiple ports to a single hardware configuration. To do so, you need to use the first one for a device, then save the configuration and edit it again, This way, you get an additional port to choose from. This process would be made easier by something like #587.

I hope there are no major issues, I didn't have a lot of devices to test with.

Hoping someone finds this useful :)

P.S.: One question: Which files do l need to modify in order for the device adapter to be built together with the rest?


nicost:
To get a device adapter to build on Windows, I think you already know how to do that.  For Unix, see my FakeCamera related commits here: https://valelab4.ucsf.edu/trac/micromanager/timeline

nicost:
Thanks!  I added your code to the 1.4 repository, and this should be in tomorrow's build on Windows.  Will look tomorrow at the Unix build.  I also gave you write access to the subversion directory DeviceAdapters/TCPIPPort.  Let me know if there are any problems with the code in the distribution.

PR #596, opened 2017-11-27
ASIStage: Allow clearing of buffer values

berl:
Remove the check on runningFastSequence_ in ZStage::ClearStageSequence().  loadStageSequence would append new sequence positions instead of replacing them.

### DIFF
--- a/DeviceAdapters/ASIStage/ASIStage.cpp
+++ b/DeviceAdapters/ASIStage/ASIStage.cpp
@@ -2664,10 +2664,6 @@ int ZStage::SendStageSequence()
 
 int ZStage::ClearStageSequence()
 {
-   if (runningFastSequence_)
-   {
-      return DEVICE_OK;
-   }
 
    sequence_.clear();
 


berl:
@jondaniels do you/ASI have feedback on this?  Thanks!

jondaniels:
@berl I believe you have identified a logical bug, but a bug that is impossible to exercise.  I suspect you mis-understand the use of the runningFastSequence_ flag.

There is a special mode set using property "Use Fast Sequence" which corresponds to the boolean flag runningFastSequence_.  This mode exists to allow for sequential TTL-controlled stacks without any serial communication and is described on the [ASIStage device adapter wiki page](https://micro-manager.org/wiki/ASIStage) in the paragraph beginning with "There is a further feature". 

Briefly, when the "Use Fast Sequence" property is set to "Yes" then all the sequence-related methods in the device adapter are short-circuited (without any communication to the controller) including the method you suggest changing.  In other words, when property "Use Fast Sequence" is set to "Yes" then loadStageSequence() **should** have no effect whatsoever... I'll come back to this **should**.  When set to "No", i.e. when useFastSequence_ is false, then a call to loadStageSequence() will clear the controller's ring buffer and repopulate it with new sequence positions via the device adapter's ClearStageSequence(), repeated calls to AddToStageSequence(), and finally a call to SendStageSequence() ([see core's code for loadStageSequence()]( https://github.com/micro-manager/micro-manager/blob/df7e3750918c36659edded0527e00409e4f75129/MMCore/MMCore.cpp/#L2137)).

The following is true when "Use Fast Sequence" property is set to "Yes": AddToStageSequence() is **not** short-circuited.  Thus calls to loadStageSequence() will repeatedly append the sequence positions to the internal array held in the class variable sequence_.  This is a bug.

However, the bug has no detrimental side effects, because the same class variable sequence_ is cleared in the first call to loadStageSequence after "Use Fast Sequence" is set to "No" in the ClearStageSequence() method.

Does this make sense?

berl:
@jondaniels I'm using the python API and so may access to the driver is a bit different.  I've confirmed the behavior you describe- if I toggle the UseFastSequence before and after loading the positions there's no problem.   I can also confirm that it's easy for the bug to show up in real life through the python API, all you have to do is not toggle UseFastSequence and the loadStageSequence method just appends locations.   Perhaps a solution would be to raise an error if the user tries to load positions when UseFastSequence is true.

jondaniels:
@berl Two things:

1. I just checked in a bugfix that ensures that the list of sequence points can't be added to when UseFastSecquence is enabled.  [SVN r16549](https://valelab4.ucsf.edu/trac/micromanager/changeset/16549).

2. The core's loadStageSequence() code should take care of clearing the sequence points, but if you are calling any of the device's methods yourself (e.g. SendStageSequence(), StartStageSequence, etc.) then you need to be careful to call ClearStageSequence() before calls to AddToStageSequence().

Hopefully between those two the issue will be resolved.

nicost:
Is it OK to close this PR?

berl:
yes thanks!

PR #605, opened 2018-01-20
Opencvgrabber device enumeration camera choice by name

bfritscher:
Although there seemed to be a camera number property after reviewing the code it did nothing as CV_CAP_ANY was hardcoded, and the camera number property ignored.
The following changes enumerate the detected cameras by name and initialise the correct camera.
Tested with China USB Camera x500 recognized as Lenovo EasyCamera on Windows 10.

PR #606, opened 2018-01-24
Opencvgrabber device enumeration master

bfritscher:
Although there seemed to be a camera number property after reviewing the code it did nothing as CV_CAP_ANY was hardcoded, and the camera number property ignored.
The following changes enumerate the detected cameras by name and initialise the correct camera.
Tested with China USB Camera x500 recognized as Lenovo EasyCamera on Windows 10.

This version compiles against master with Windows SDK 7.1. 

PS: Could not test with downloaded version 1.4 of Micro-Manager because I get "incompatible device interface version required(required = 65; found = 68)"

Old version for 2.0beta was here https://github.com/micro-manager/micro-manager/pull/605

### DIFF
--- a/DeviceAdapters/OpenCVgrabber/OpenCVgrabber.cpp
+++ b/DeviceAdapters/OpenCVgrabber/OpenCVgrabber.cpp
@@ -168,10 +168,11 @@ COpenCVgrabber::COpenCVgrabber() :
    DeviceEnumerator de;
    // Video Devices
    map<int, OpenCVDevice> devices = de.getVideoDevicesMap();
+   map<int, OpenCVDevice>::iterator it;
 
-   for (int i = 0; i++; devices.size())
+   for ( it = devices.begin(); it != devices.end(); it++)
    {
-       AddAllowedValue(cIDName, devices.at(i).deviceName.c_str(), long(i));
+       AddAllowedValue(cIDName, it->second.deviceName.c_str(), long(it->first));
    }
 #else
    String cIDNameReally = "Camera Number";


nicost:
Figured out that I need to download the github diff page directly.  Copying and pasting into vim apparently results in errors.  That may have been the problem all along.  Sorry!
As for 1.4, always test against the latest nightly build rather than the release build.  The latest nightly will have the same device interface version as the current source code.  In any case, your patch is committed to the 1.4 branch, I'll check the build tonight or tomorrow, fix if needed, and the patch will make its way into 2.0 in a little bit (next time I'll merge the 1.4 C++ changes into 2.0).

nicost:
Turns out that this breaks the build on Mac (and linux).  Quite understandable given the inclusion of "Windows.h".  That one will be easy to ifdef out, but there will be likely consequences elsewhere.  Is there any way you can make the device enumeration cross-platform?  Otherwise, can you make sure that everything on Mac/Linux will still works as it used to (i.e. make sure that the code path without the device enumeration still works)?  Since I applied the pr already, please rebase against the master branch for any new code.

nicost:
OK, I just committed code to the svn repository that hopefully restores the UNix build to the situation before your patch.  The Windows code should be unaffected.  A bit ugly, but hopefully the easiest fix.  When you have a chance, can you test the nightly 1.4 build to make sure everything is correct?

bfritscher:
Thanks for your support, I managed to test with the nighltly build x64 from the 23th. It works now with the provided fix in this updated pull request. Sorry I am learning C++ as I am going.

Your fix seems ok. Just that it restores the camera number property which does nothing, since cvCaptureFromCAM(CV_CAP_ANY) take the first available camera.

nicost:
Got it.  Would it be safe to remove the ifdef at line 237 in OpenCVGrabber.cpp and let Unix also use the cameraID (which can be set numerically through the property)?

bfritscher:
As I understand it and have tested it seems so, as long as the default value is 0, which is the case.

enum | {   CV_CAP_ANY =0,
https://docs.opencv.org/trunk/dd/d01/group__videoio__c.html

nicost:
Perfect!  Committed that change.  Hope we are all good to go!

PR #609, opened 2018-01-31
Support Zeiss Definite Focus 2  stabilizing position

tinekelenstra:
This commit changes the ZeissCAN29 device adapter by adding a property "DefiniteFocusVersion" (either 1 or 2) and using its value while fosussing (using the StabilizeThisPosition command) to determine the amount of data to read from the response.


tinekelenstra:
This was developed on top of commit 61ebc9d998 as we've been using 1.4.22 of micro-manager. The ZeissCAN29 device code has only changed trivially since (2e75ff1 only comments out unused parameters in another method) so we hope this easily merges and works on the current development head of micro-manager.

nicost:
Thanks Tineke!  Applied in svn commit 16613.  The diff applied cleanly against head.  It should be in tonight's nightly build and find its way into the various 2.0 versions soon.

PR #610, opened 2018-02-07
Merge C++ changes from 1.4

marktsuchida:


marktsuchida:
It looks like some inconsistencies were introduced between master (mm1.4) and mm2 during the past few merges of the C++ layer. This merge should resolve those, bringing mm2 C++ layer to an exact match to 1.4.

nicost:
I do not really understand what happened with this PR.  The PR lists 216 commits, many of which had already been applied to mm2 , but now appear again with a different hash (for instance, the "Core: Added storage for affine transforms".  Also, there are many commits that deal with the 1.4 Java code included in this PR.  That should lead to chaos, but it doesn't.  Can you explain what happened here and how it was done?

marktsuchida:
This is a case where GitHub's PR display makes it look a little weird. Viewing in `gitk` should make it clearer.

In terms of pure Git, it's just a merge from `master` into `mm2`, nothing more. When resolving the merge, I set all the C++ parts to match `master`, and kept everything else from `mm2`.

This is done by running `git merge --no-commit -s ours master` on `mm2`, followed by replacing the C++ files with those of `master` (taking care not to miss files deleted in `master`), then committing.

GitHub doesn't know the intent of the merge, only that `master` was merged into `mm2`, so it lists all the commits that happened on `master` since the last merge.

nicost:
How do you "replace the C++ files with those of master"?

marktsuchida:
By doing `git checkout master -- MMDevice` etc.
This is the part that cannot be automated because there can be subtle issues with buildscripts. Also, files deleted in 1.4 need to be removed.

PR #618, opened 2018-03-16
Open shutter automatically when starting live

henrypinkard:


nicost:
Hi Henry!  Why is this needed?  The core already opens the shutter when autoshutter is checked.  Having this functionality on multiple places is scary, and bringing it up from the core to the UI seems a step backwards.  I guess that you put this in since the Core/Device are not doing their job?  Does this happen to be with a Photometrics camera?  There appears to be a bug in their device adapter that causes the shutter to misbehave.

On the other hand, we seem to make it difficult for device adapter writers to write the code correctly, so it is certainly worth re-considering how this is handled (but since there is code in many places related to this issue, it may be a lot of work to do this correctly).

henrypinkard:
Hi Nico, yes, I was having this problem with a Photometrics camera, as well as a new device adapter I'm working on. Makes sense that this shouldn't be in the UI, though it's unclear to me what the correct behavior is supposed to be. 

When snapping an image the core handles shutter opening if autoshutter is checked, but I didn't see any similar behavior core.startConstinuousSequenceAcquisition. Should this function be handling autoshutter behavior? Or is it supposed to be somewhere else--in the camera device adapters themselves?

nicost:
Good questions!  My memory is hazy.  However, the deviceadapter is supposed to make the callback:
 int ret = GetCoreCallback()->PrepareForAcq(this);
when starting the acquisition.  That callback (code defined in CoreCallback.app) will open the shutter when autoshutter is det.
The device adapter is also supposed to call:
 GetCoreCallback()->AcqFinished(this,0);
when stopping the sequence acquisition.  That code (also defined in CoreCallback.cpp) will close the shutter.
I think that the reason for this implementation is to make sure that the shutter is open only for the smallest amount of time.  It is possible that the camera needs to prepare some things before starting a sequence, and it would be sad if the shutter were already open.  The downside is that it relies on correct implementations by the device adapter, but luckily, it becomes quite clear when there are problems.  

For instance, looking at the PVCAM adapter code, it seems they folded the PrepareForAcq callback in a block that will not be executed everytime sequence acquisition starts. I just committed a change that puts in the right place (or so I think).  Can you check the next nightly build (of 1.4)?  I am not sure of the placement of AcqFinished() in the PVCAM adapter.  Check your own adapter for those callbacks..

henrypinkard:
Got it. With that added my device adapter is behaving as expected. PVCAM appears to be fixed as well in the latest nightly. 

I'm still a little confused that the autoshutter behavior is implemented in core.SnapImage. Seems like based on the logic for sequence acquisitions that maybe the same thing might be desirable for snapping. Unless in practice no cameras are doing extra preparation for snapping a single image. Though it also might break the shutter behavior for existing adapters if it were to be removed...

nicost:
You are abslutely right!  In fact, there are cameras (forgot which one, Andor or PVCAM), where changing the exposure time leads to the shutter being open ~500ms longer than the exposure time.  I think this is mainly a historic decision.  We started with SnapImage, and implemented it as simple as possible.  When implementing sequence acquisitions, we ran into the problem that starting a sequence took a long time, hence this approach.  It may have been better to also have the camera decide when to open/close the shutter.  If/when a redesign of the Core takes place, these aspects definitely should be reconsidered.  Doing it now, would uproot quite a bit of code..

PR #624, opened 2018-04-16
implement SetRelativePositionUm() for MultiStage

jondaniels:
I was recently wanting to move 2 linear stages together with a scale factor and I discovered that there already is a MultiStage device for doing this.  Great!

However, this MultiStage device is missing the ability to do "native" relative moves which is important to me.  The default implementation utilizes absolute moves and only knows the current position of the 1st stage and assumes that the other stages haven't been moved otherwise (which for me isn't a valid assumption).  The fix is very straightforward, invoking relative moves on all the physical stages.  I wouldn't expect any trouble unless somebody is depending on the odd behavior of the existing adapter.  If that's the case then we could add a property to determine which behavior to use.

### DIFF
--- a/DeviceAdapters/Utilities/Utilities.cpp
+++ b/DeviceAdapters/Utilities/Utilities.cpp
@@ -1121,6 +1121,22 @@ int MultiStage::SetPositionUm(double pos)
 }
 
 
+int MultiStage::SetRelativePositionUm(double d)
+{
+   for (unsigned i = 0; i < nrPhysicalStages_; ++i)
+   {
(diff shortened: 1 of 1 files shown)


nicost:
Kurt Thorn wrote this for his home-build light sheet.  I doubt that anyone else depends on this behavior, so it would be easiest to push this change in and I can then check with Delaine (who currently runs the imaging center)  if things are still working correctly.

nicost:
No, I just committed it to the subversion repository. I did have to add the method signature to the header file as well.  Let me know if it works as expected.

Issue #625, opened 2018-04-27
device detection (port scanning) broken with TCP/IP ports


jondaniels:
With TCPIPPort device adapter loaded the previously-working port scanning fails.  Problem goes away if TCPIPPort DLL is deleted.

I think I have figured out the cause.  With the TCPIPPort device adapter installed there is a TCP/IP port that now appears including in the list of available ports during the HCW.  When that port is evaluated in the following loop an error is thrown because the TCP/IP port doesn't have a Baud rate property (among others) and the MMCore code assumes all ports will have such a property.

https://github.com/micro-manager/micro-manager/blob/c8fc57369edd795a4c500d9ed3642fa74d34b9a8/MMCore/MMCore.cpp#L7375

(then property value is fetched in L7386)

The fix would be to either not scan TCP/IP ports or somehow make the MMCore code robust against this sort of error.  Note that the error happens and autodetect fails even if the device has already successfully been found on another port.

I'm going to try to propose a fix in a PR but I don't have a build system set up to test MMCore changes.

marktsuchida:
PR #626 from @jondaniels fixes this.

PR #626, opened 2018-04-27
attempt to fix device detection (port scanning)

jondaniels:
See https://github.com/micro-manager/micro-manager/issues/625.  This is attempt to gracefully exit when port properties can't be read.  Currently failure happens when the port is a TCP/IP port because this port doesn't have such properties and the whole detection process is aborted (even if the device has been found).

I am not set up to test this code until it's committed and nightly build happens.


marktsuchida:
Manually merged to SVN (r16706). Thanks, @jondaniels, and @nicost for testing.

PR #627, opened 2018-05-04
Updated data type for trigger counts in Arduino AOTF firmware.

aaristov:
Changing int to long solves the problem of triggering stopping after 32768 frames. Tested in my lab.


nicost:
Incorporated in Micro-Manager 1.4 subversion repository in commit 16701.  Should make its way to the 2.0 branch soon.

PR #629, opened 2018-05-08
TUCam adapter added new function

ZRen86:
1.Added the trigger function
2.Added the DPC function
3.Closed the fan gear function
4.Changed the display property name


nicost:
Added to 1.4 branch in subversion commit r16702

PR #630, opened 2018-05-18
SmarAct HCU-3D

jdeschamps:
Hi,

I greatly improved the SmarAct HCU-3D device adapter by adding error checking and fixing a bug that would prevent people from using the grid generation in the stage position list.

Best,
Joran


nicost:
Applied in subversion commit r16804.  I am unable to close this pull request (but it is no longer needed).

PR #639, opened 2018-07-10
Fix the SnapImage bug

ZRen86:
Fix the SnapImage bug and open the software trigger options


nicost:
Incoporated in https://github.com/micro-manager/micro-manager/commit/064ba06eac27bd15afa744f2edb0c959ec1ef193

PR #642, opened 2018-07-13
LaserQuantum lasers

jdeschamps:
I wrote a simple device adapter to control the gem/ventus/opus/axiom laser series from LaserQuantum. It discriminates between the laser modified to allow current control and those designed only for power control. It then only display the corresponding properties.


nicost:
About to add your code to the 1.4 repository (sorry for the delay!), but get what looks like an important compiler warning:

LaserQuantumLaser.cpp:441:35: warning: variable 's' is uninitialized when used
      within its own initialization [-Wuninitialized]
                std::string s = answer.substr(0,s.length()-1);
                            ~                   ^
1 warning generated.


Did you mean to use the length of "answer" there (i.e. are you trying to strip the last character from the response from the laser)?  This code looks extremely crash prone...

jdeschamps:
Yes, I will check next week!

jdeschamps:
I tried back then, and tried again now, but I can't find the device adapter in the most recent nightly build. Any issue there?

nicost:
Looks like it was never added to the Windows solution file.  I'll try to do so later today.  Please do get back with these things at your earliest convenience, because nothing will happen without you reporting;(

jdeschamps:
Worked for the 1.4.23NB and the 2.0gammaNB!

PR #648, opened 2018-07-27
Make Serial port communication faster for USB based serial ports

lukas-lang:
This PR changes the timing settings for serial ports under Windows. Essentially, this speeds up USB based serial ports significantly - I have observed an improvement of up to **10x** (from ~70ms to ~7ms per command) compared to the previous implementation. This makes navigating using serial port based stages significantly smoother.

The new timing settings effectively tell Windows to complete the read calls to the serial port immediately instead of waiting for a while. For a more detailed explanation see [this post on StackOverflow](https://stackoverflow.com/a/48977270/3032376).

Since I do not have access to a computer with a true serial port, I could not quantify the impact on this type of port. However, my understanding is that it should not negatively affect performance.

### DIFF
--- a/DeviceAdapters/SerialManager/SerialManager.cpp
+++ b/DeviceAdapters/SerialManager/SerialManager.cpp
@@ -641,7 +641,10 @@ int SerialPort::OpenWin32SerialPort(const std::string& portName,
 
    COMMTIMEOUTS timeouts;
    memset(&timeouts, 0, sizeof(timeouts));
-   timeouts.ReadIntervalTimeout = 1;
+   timeouts.ReadIntervalTimeout = MAXDWORD;
+   timeouts.ReadTotalTimeoutMultiplier = MAXDWORD;
+   timeouts.ReadTotalTimeoutConstant = 1;
+
    if (!SetCommTimeouts(portHandle, &timeouts))
    {
       DWORD err = GetLastError();


nicost:
This sounds amazing!  Thanks so much for looking into this!  Since serial communication is critical, I'll test this our on one or two of our own rigs before committing, but I am very much looking foreword to this patch doing its magic!

nicost:
I tested your patch on one of our systems and it is amazing how much faster the Device/Property Browser becomes!  Thanks so much for figuring this out, what an unbelievable pleasant patch!  I committed it to svn (r16757).  If it breaks anything, we'll hopefully hear about it soon.  Hopefully it will make its way soon to the various 2.0 branches.

lukas-lang:
That's great to hear!

Regarding stability: I have had this code running on our setup for over 6 months now and we didn't notice any issues resulting from it, so I'm hopeful ;). (But then again, there's always someone with an unexpected configuration that behaves very strangely)

nicost:
In subversion r16762 I added a property to use the timeout settings you propose or the previous one.  I did so, since we are currently experiencing an instability (including blue screens) with one of the Arduinos on one of our microscopes.  Impossible to reproduce, hence hard to pinpoint to this change, but it seems to have coincided, and older micro-manager versions seemed more stable, hence I liked having the ability to change this.  
I also tested the effect on speed, and found no difference for the Arduino (Uno).  In both cases, a serial command was send in 4 ms (on average).

nicost:
Received some more reports about blue screens that seem to correlate with this change (they are rare and very difficult to troubleshoot, but clearly correlate with this change).  Made the default to not use this setting (but keep the pre-initialization property to enable it).

nicost:
I can not close this PR.  Lukas, can you?

lukas-lang:
Should be closed now.

Regarding the bluescreens: That's unfortunate to hear - But having the possibility to opt-in also works nicely I think (sorry for not reacting sooner, I was occupied with different things and completely forgot about this)

PR #654, opened 2018-09-07
Reliability fixes for pixel calibrator.

nanthony21:
After struggling to get the pixel calibrator to work for me I decided to take a look at the code and made some changes that I think will make it a bit more reliable and also make the code easier to read.

1: Plugin will now always recenter at the end of a failed calibration

2: Safe travel radius is now treated as a radius from the start point of the acquisition rather than a maximum relative movement distance.

3: A useful cross correlation plot is displayed.

4: Added lots of comments to explain code.

5: Simplified code by removing unnecessary class variables and using Point2D objects instead of pairs of variables.


nanthony21:
Thanks Nico,

I'll give Gamma a try and see if any of my PRs here can be applied to your branch.

Would it be possible to get your gamma branch included in this repository or perhaps to add some information to the readme.md? Source code, and information about MicroManager is so fragmented across many locations at this point that it is very difficult for a newcomer to figure out which version to use.

--Nick

nicost:
Very much agree.  Things are a bit chaotic right now.  Hopefully everything will be cleaned up in the near future, and at that point in time we can add decent descriptions.  Till then, best to communicate through the mailing list, where I will be happily giving my understanding of the situation.

PR #655, opened 2018-09-11
Fix variLC crashes

nanthony21:
This pull request improves the stability and versatility of the variLC.

Previously the adapter would cause a sudden crash of MM if a wavelength outside of its tunable range was specified. Now it will log an error and prevent the crash from occuring.

Rather than assuming that the tunable range of the device is 400-800 nm the adapter will now query this information from the device during initialization. Different models have different tunable ranges.

### DIFF
--- a/DeviceAdapters/VariLC/VariLC.cpp
+++ b/DeviceAdapters/VariLC/VariLC.cpp
@@ -283,11 +283,16 @@ int VariLC::Initialize()
 		return ret;
 
 	// Wavelength
+	std:string ans;
+	ret = sendCmd("V?", ans);	//The serial number response also contains the tuning range of the device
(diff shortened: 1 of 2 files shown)


nicost:
Applied to the sub-version repository.  At the moment, these changes are no longer back-propagated to github.  That is another problem that needs to be fixed.

nanthony21:
Thanks. I'll close it. What needs to be done to fix svn/github syncing?

nicost:
I accepted a PR to the master branch, which I should not have done.  Either that commit should be nuked (risking breaking git repositories for people building on top of it), another solution for syncing should be found, or the subversion repository should be retired completely.  Regretfully, there is not much I can do.

PR #656, opened 2018-09-12
Fix display on acquire multiple regions

nanthony21:
The status display on this plugin is outside of the window and cannot be seen. This moved it back into the viewable portion of the window.


nicost:
I guess this is no longer relevant.  OK to close?

nanthony21:
I beleive so.

PR #658, opened 2018-09-17
Fix corrupt metadata acquire multiple

nanthony21:
This pull request fixes an issue where datastores created by the acquire multiple regions plugin were not frozen prior to being closed. This would leave them in a state where their metadata was not complete.

I have also changed the plugin so that it will name it's positions in the position list with a grid location.

This opens up two possibilities for using ImageJ to stitch together the images (by metadata and by filename) that were not available before.


nicost:
I guess this is no longer relevant.  OK to close?

nanthony21:
This was long ago that I don't remember but I think this is fixed in the ViewerCV branch now.

PR #661, opened 2018-10-24
Toptica iBeamSmart laser 

jdeschamps:
This device adapter allows controlling the iBeamSmart laser series from Toptica, which are popular lasers. After discussion with the company, the decision was made to use a hidden control mode (CW) of the laser. Using this mode simplifies the interface with the laser (a single channel instead of two, no need to switch between the channels when using FINE and external trigger). The device adapter includes two classes:

* iBeamSmartCW:
Class used to control the laser, upon instantiation it switches the laser to CW mode. It is compatible with the pulse option from Toptica. It also includes the possibility to change the FINE settings (typical of iBeamSmart lasers). Upon quitting MM, the laser is switched back to normal control mode.
* iBeamSmartNormal:
In the event of a MM crash, the laser might remain in CW mode. If this is the case, Topas (the Toptica software) cannot be used anymore to control the laser. The laser can be reversed to normal mode by loading the iBeamSmartCW again and quitting MM properly. Another way is to load the laser as an iBeamSmartNormal. This class sole purpose is to quickly make sure that the laser is in normal mode and cannot be used to control the laser. 

This device adapter is already used routinely by our group and other collaborators. The need for this adapter has also been discussed on the mailing list (http://micro-manager.3463995.n2.nabble.com/Toptica-iBeam-smart-in-MM-td7589416.html).

PS: I am not sure the iBeamSmartNormal class is entirely really necessary. If for some reason there are issues with the laser that makes iBeamSmartCW not properly load, iBeamSmartNormal contains the minimum code to switch it back to normal mode (which is a bit involved still).


nicost:
This code was added to the subversion repository.  The PR can be closed (by anyone allowed to do so, which is not me;)

PR #665, opened 2018-11-23
Fix step size issue in RAMPS stage controller

nvladimus:
I found an issue in [RAMPS](https://github.com/micro-manager/micro-manager/tree/master/DeviceAdapters/CNCMicroscope/RAMPSStage) (Arduino) stage controller device adapter which didn't allow to move Z-stage by increments smaller than 5 micron. The issue is fixed.  The new code was tested on Thorlabs DRV001 stage and reliably moves Z in 0.625-um steps.
List of changes:
* new min step in all axes is 0.025 um. This value is below the capability of typical stepper - driven stage, but it's useful for converting microns to microsteps for motors with 200 steps/revolution and 500-um pitch lead screws.
* max speeds of all stages XYZ were set to 240 mm/min.
* XYZ min travel limits were set to -20 mm to allow moves in both directions from arbitrary starting point (zero).


nvladimus:
Sorry, included too many files in the pull request. Still learning this pull thing.

PR #667, opened 2019-02-04
Feature/video4linux2 improvements

dedeibel:
Hello Nico,

improving on PR https://github.com/micro-manager/micro-manager/pull/588 I had some time for more changes on the device. It improves the usability of the device greatly.

There are lot of changed lines in the diff. Here is a list of the performed actions:

  - Adding support for RGB image capturing
  - Allow device path to be configured using a property. Reinit the device if it changes.
  - Allow the resolution to be configured using properties.
  - Used tryioctl function to honor busy state of devices and retry the action.
  - Call v4l2 capability negotiation api to get more detailed error messages in case of
    incompatible devices.
  - Unify logging and leverage errno value of ioctl calls.
  - Move some functions into the device class as they are actually accessing it's state
  - Rearranged functions to follow the call flow better (init, prepare, read, close, ...)

Thank you for considering merging the code, please let me know if you see some problems.

Ben.

### DIFF
--- a/DeviceAdapters/Video4Linux/video4linux2.cpp
+++ b/DeviceAdapters/Video4Linux/video4linux2.cpp
@@ -21,6 +21,24 @@ sudo cp libmmgr_dal_video4linux2.so /usr/lib/micro-manager/libmmgr_dal_video4lin
 export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/usr/lib/micro-manager
 cd /usr/share/imagej
 strace -f java -Djava.library.path=/usr/lib/micro-manager -classpath /usr/share/imagej/plugins/bsh-2.0b4.jar:/usr/share/imagej/ij.jar:/usr/share/imagej/plugins/Micro-Manager/MMJ_.jar org/micromanager/MMStudioMainFrame
+*
+* Modified:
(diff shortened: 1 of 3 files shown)


nicost:
Hi Ben,

I saw nothing wrong and committed the code (to the subversion repository).  I can not test easily.  Can you have a look?

Also the include for arpa/inet.h in the WieneckeSinske adapter is already there, so you may be working against an older branch.  For Device Adapters, the authorative code is in the subversion repository (https://valelab4.ucsf.edu/svn/micromanager2/trunk), which is mirrored in the svn-mirror here on github.

dedeibel:
Thanks for the information about the svn branches.

I tested the svn checkout and found a small problem regarding the colored mode. It seems like the depth value is now defined as per component. I created a follow up PR as this one is already merged.

https://github.com/micro-manager/micro-manager/pull/668

I also added another small improvement regarding handling invalid device property values. The return value was not propagated completely by my property handler.

It seems I started off with the wrong branch, since "mm2" is configured as base branch of this repo. Maybe this should be changed as it is confusing to newcomers.

dedeibel:
Was commited to r16909

PR #668, opened 2019-02-17
Video4Linux: Define bit depth per component (required because of base…

dedeibel:
Follow up to https://github.com/micro-manager/micro-manager/pull/667

> I tested the svn checkout and found a small problem regarding the colored mode. It seems like the depth value is now defined as per component. I created a follow up PR as this one is already merged.
> 
> I also added another small improvement regarding handling invalid device property values. The return value was not propagated completely by my property handler.


nicost:
It would make things much easier if you would have write access to the subversion repository and push your changes there directly.  Can you send me an email?

dedeibel:
Sure, thanks, on the way.

dedeibel:
Commited  with r16924

PR #673, opened 2019-05-13
Add New Project YodnE600 to Solution

maobaoqi:
YodnE600 is a light contoller.As device adapter, this poject has been test successfully and the module has been used in product.

### DIFF
--- a/DeviceAdapters/YodnE600/E600.cpp
+++ b/DeviceAdapters/YodnE600/E600.cpp
@@ -3,7 +3,7 @@
 // PROJECT:       Micro-Manager
 // SUBSYSTEM:     DeviceAdapters
 //-----------------------------------------------------------------------------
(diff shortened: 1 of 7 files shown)


nicost:
Added to the build in subversion r117030 (in the 1.4 source tree).  Please do test tonight's build to make sure the code still works for you.  Also, please do request an account on https://micro-manager.org and send me an email so that I can approve your request, then create an information page about your device, so that others will be able to use it as well.
Also, this branch here does not get updates, and I do not have permission to do anything, so please do close this PR.

nicost:
Also, I encountered a couple of errors due to the use of C++11 functions (yes, MM uses an old compiler).  Please do get the current code from the svn repository in case you want to make changes later (I do not want to make the same changes again;)

maobaoqi:
@nicost thank you for reply!:) I have request an account on website just now.

PR #692, opened 2020-02-07
Allow specification of default User Profile to skip splashscreen

nanthony21:
This PR extends the changes in: https://github.com/nicost/micro-manager/pull/48 so that they can be used when MMStudio is launched from ImageJ rather than only from Matlab.

This adds the ability for a user to pass arguments to micromanager at startup just like you would with any CLI program. 

In the case of launching MMStudio from ImageJ this would be done in an imageJ macro as `run("Micro-Manager Studio", "-profile MyProfileName");` A user could potentially modify the following line  to make this happen as soon as ImageJ is launched: https://github.com/micro-manager/micro-manager/blob/7a37a23f9134a16807a1f6662322d6301588bfd4/bindist/any-platform/macros/StartupMacros.txt#L45

Currently the only option supported is the `-profile {UserProfileName}` option which causes MicroManager to skip the splash screen and load directly into a user profile. Other options could easily be added to the code in this pull request.

Assuming that this feature is accepted I see two questions that should be resolved before merging:

  - The code to parse the arguments is currently implemented in three places in order to support multiple methods of running Micro-Manager (MMStudioPlugin, MMStudio::main, and the StartMMStudio.m matlab script). The convergence point of these three methods is the constructor of MMStudio. Should the string parsing be moved into the constructor of MMStudio so there is only one implementation?

  - MMStudio::main claims to be used for running MMStudio as "stand-alone". Is this still possible, I couldn't get it to work? If not, then maybe this would be a good time to remove the function altogether.

### DIFF
--- a/mmstudio/src/main/java/MMStudioPlugin.java
+++ b/mmstudio/src/main/java/MMStudioPlugin.java
@@ -25,6 +25,7 @@
 import ij.CommandListener;
 import ij.Executer;
 import ij.IJ;
+import ij.Macro;
 import ij.plugin.Duplicator;
 import ij.plugin.PlugIn;
 import java.lang.reflect.InvocationTargetException;
@@ -49,7 +50,7 @@ public class MMStudioPlugin implements PlugIn, CommandListener {
     * @param arg the plugin argument (not used)
     */
    @Override
-   public void run(final String arg) {
+   public void run(final String arg) {      
       SwingUtilities.invokeLater(new Runnable() {
          @Override
          public void run() {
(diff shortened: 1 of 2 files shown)


marktsuchida:
@nanthony21 I think this is ok to merge unless you were planning on further changes.

I actually think the string parsing should remain separate for each language, in case different (language-specific) syntax is required. Also because some of the `StartMMStudio.m` arguments are consumed by that script and not MMStudio. For the IJ1 macro language, the IJ1 plugin object `MMStudioPlugin` seems like the right place.

Although adding constructor arguments to `MMStudio` is not scalable (if/when we need more), that can be addressed at a later time, since `MMStudio`, including its constructor, is an internal class. (The API should probably have a public method to construct a `Studio` other than through `MMStudioPlugin`, and whatever we come up for that in the future can accept a builder or "startup parameters" object of some sort, if warranted.)

Regarding `MMStudio::main()`, it certainly worked last time I checked (it's been a while). The autotools build (`./configure && make`) will build an installation that uses this by default, if `--enable-imagej-plugin=IJPATH` is not used. In theory I guess this should also support a user profile parameter, but I don't think that many people run Micro-Manager this way, so it doesn't seem like a high priority.

nanthony21:
@marktsuchida Great! In that case I think this is ready to merge.

PR #696, opened 2020-02-10
Enable by-passing the splash-screen again

nicost:
Options were still present but not functional.  This PR makes them functional again.  However, I can not makes these options work for the non-default-users.  In addition, it makes sense that only the default user can set these settings, so that is the approach I took.

In general, it would be nice to have a clearly documented mechanism for loading properties that need to be in effect before the profile is loaded/selected.

### DIFF
--- a/mmstudio/src/main/java/org/micromanager/internal/dialogs/IntroDlg.java
+++ b/mmstudio/src/main/java/org/micromanager/internal/dialogs/IntroDlg.java
@@ -70,6 +70,8 @@ public final class IntroDlg extends JDialog {
 
    private ProfileSelectionUIController profileController_;
    private ConfigSelectionUIController configController_;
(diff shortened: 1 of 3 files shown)


marktsuchida:
I think this looks good. Only being able to set this from the Default profile makes sense, because such a setting implies that only the Default profile will ever be used.

I suspect we need to actually implement a coherent mechanism before being able to document it. I definitely only got 1/3 of the way when I implemented the `UserProfileAdmin` class. One problem is that we do not have a dedicated GUI for "admin settings" or "user profile management", to handle the few meta-settings that logically do not belong inside individual profiles. (Such an interface could perhaps also provide a GUI way for an admin user to create the global profile.)

nicost:
Agreed.  In practice it needs a bit of discipline to separate the two activities, but I'll try more.

Issue #704, opened 2020-02-14
Scripts in script panel are not remembered for each profile


nicost:
Scripts listed in the Script panel are the same for all profiles.  It seems that only the scripts used by the default user are stored, and used by all other profiles.  They should be stored in each individual profile.

nicost:
Fixed in PR #706

PR #706, opened 2020-02-14
Studio-ScriptPanel: Correctly stores scripts in profile.

nicost:
Previously, because the profile was used before the UserProfile was activated,
and after it was shutdown, settings were read and written to the default
profile most (but not all) of the time. This commit mainly changes the
order during startup, and ensures that settings will not be saved
twice during shutdown (which apparently often runs twice).


nicost:
Addresses issue #704

PR #718, opened 2020-03-08
Merges Henry's ZMQ branch into master

nicost:
Adds ZMQ-based RPC bridge that is used by Pygellan.  Java classes are described in JSON, and only primitives are transferred through the port, ShadowClasses are created for other objects.  The bridge can be activated under Tools > Options.  @henrypinkard  and @marktsuchida: OK to merge?


henrypinkard:
Looks good to me!

marktsuchida:
Looks good. Just a couple of things:
- Is there a write-up of the wire (JSON) protocol somewhere? If not it would be good to have (after merge is fine).
- In my quick look it looks like use of json.org is purely internal and it won't get in the way of upcoming changes (we'll rename our ancient copy of `org.json` to something internal, so that we don't clash with current versions `org.json` Jars). Am I right?

Otherwise, ok to merge!

henrypinkard:
Yes, can write that up. This should still be considered alpha stage functionality and subject to change, but hopefully can finalize it in the next couple months.

Yes, I think that's right. JSON is just used to serialize things into byte arrays and is not externally exposed at the moment.

I think it might ultimately make sense to split this out into a separate library that the studio instantiates, so that, for example, one could run this in a headless mode without having to launch ImageJ and the GUI. What do you think?

marktsuchida:
Alpha stage is fine, you can just say so in the document.

I totally agree that a separate library makes sense! Would that be usable even independent of MM and IJ (as a generic Python-to-Java access)?

henrypinkard:
I think it should be, though there are still some details to iron out

nicost:
Will merge this in now, to make it easier to go forward.  Transitioning to a separate library should be easy as there is only one point of contact with the current code.

PR #721, opened 2020-03-11
fix compilation issues for Ubuntu 18.04

triccyx:
I am current working on Ubuntu 18.04 gcc 7.5.0.
Some minor compilation issues and a java fetch issue

### DIFF
--- a/DeviceAdapters/FakeCamera/FakeCamera.h
+++ b/DeviceAdapters/FakeCamera/FakeCamera.h
@@ -27,7 +27,9 @@
 
 #include "DeviceBase.h"
 
-#include "opencv/highgui.h"
+
+#include <opencv/cv.hpp>
+
 
 #define ERR_INVALID_DEVICE_NAME 10000
 #define OUT_OF_RANGE 10001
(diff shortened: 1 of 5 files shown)


nicost:
Accepting this PR would break the build on all other platforms.  
Some things that can help mitigate that problem:
- Use a platform-specific ifdef for te change in FakeCamera.h
- Spell out s-squared in CAN29, rather than replacing one unicode character with another (which also no longer renders correctly in my browser)
- MMClearVolume build.xml: I can not oversee the consequences of setting plugin.classpath, but setting it a fixed path only present on Ubuntu, can not be good for others. Likewise, simply deleting the includes for Windows (where there is a clear intent to make this Windows-specific) will make this not work on Windows.

Let me know if you can fix these issues, or at least split out in issues that should be no problem on  other platforms, and things that will and that you do not have a solution for that works for others.

triccyx:
Regarding MMClearVolume build.xml agree, I can remove from the commit.
Regarding FakeCamera.h I agree with platform-specific ifdef.
But regarding others fixes,  I think they are due to the fact that new compilers are more sensitive than the older and the corrections are ok.
What do you think?

nicost:
Yes, agreed. Happy to test if you can make a branch that I can test without doing these other things (just a volunteer here;)

dschneiderch:
also running into lots of issues on ubuntu 18.04. see #722 
will try this tomorrow. I'd help generalize this for both W and L but i've never used cpp before

triccyx:
> also running into lots of issues on ubuntu 18.04. see #722 . Thanks for tackling!
> will try this tomorrow. I'd help generalize this for both W and L but i've never used cpp before

This error will be fixed by one of my commits.

PR #725, opened 2020-03-12
Ubuntu18.04 fix only cpp

triccyx:
Only c++ files. Java is out of this fix.

Refer also to https://github.com/micro-manager/micro-manager/pull/721 aborted

### DIFF
--- a/DeviceAdapters/FakeCamera/FakeCamera.h
+++ b/DeviceAdapters/FakeCamera/FakeCamera.h
@@ -28,6 +28,8 @@
 #include "DeviceBase.h"
 
 #ifdef __linux__
+// This include has been added for compile on Ubuntu 18.04 and newer.
+// The include refers to libopencv version 3.2
 #include <opencv/cv.hpp>
 #else
 #include "opencv/highgui.h"
--- a/DeviceAdapters/FakeCamera/FakeCamera.h
+++ b/DeviceAdapters/FakeCamera/FakeCamera.h
@@ -27,7 +27,9 @@
 
 #include "DeviceBase.h"
 
-#include "opencv/highgui.h"
+
+#include <opencv/cv.hpp>
+
 
 #define ERR_INVALID_DEVICE_NAME 10000
 #define OUT_OF_RANGE 10001
--- a/DeviceAdapters/FakeCamera/FakeCamera.h
+++ b/DeviceAdapters/FakeCamera/FakeCamera.h
@@ -27,9 +27,11 @@
 
 #include "DeviceBase.h"
 
-
+#ifdef __linux__
 #include <opencv/cv.hpp>
-
+#else
+#include "opencv/highgui.h"
+#endif
 
 #define ERR_INVALID_DEVICE_NAME 10000
 #define OUT_OF_RANGE 10001
--- a/DeviceAdapters/MicroPoint/MicroPoint.cpp
+++ b/DeviceAdapters/MicroPoint/MicroPoint.cpp
@@ -185,7 +185,7 @@ double MicroPoint::AttenuatorTransmissionFromIndex(long n)
 
 int MicroPoint::StepAttenuatorPosition(bool positive)
 {
-   unsigned char buf[] = {'C', (positive ? 0xc0 : 0x80), 'C', 0x00};
+   unsigned char buf[] = {'C', (unsigned char)(positive ? 0xc0 : 0x80), 'C', 0x00};
    return WriteBytes(buf, 4);
 }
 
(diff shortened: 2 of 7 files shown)


triccyx:
This should be better for Windows and Mac compatibility.

nicost:
Just tested on my Mac, and builds correctly there.  Will test on my Windows machine (hopefully by tomorrow) and then merge if it builds correctly. Thanks for the great work!

DeviceAdapters/FakeCamera/FakeCamera.h
```diff
@@ -27,7 +27,11 @@
 
 #include "DeviceBase.h"
 
+#ifdef __linux__
+#include <opencv/cv.hpp>
+#else
 #include "opencv/highgui.h"
+#endif
```
marktsuchida: My guess is that you are using a recent version of OpenCV rather than the ancient version we're currently using (i.e. nothing to do with Linux). This change may therefore break Linux build environments that currently work. More importantly we should probably upgrade OpenCV on all platforms, but that is a harder task. So could you add a comment saying that the `#ifdef __linux__` was added to enable building on Ubuntu 18?

DeviceAdapters/MicroPoint/MicroPoint.cpp
```diff
@@ -185,7 +185,7 @@ double MicroPoint::AttenuatorTransmissionFromIndex(long n)
 
 int MicroPoint::StepAttenuatorPosition(bool positive)
 {
-   unsigned char buf[] = {'C', (positive ? 0xc0 : 0x80), 'C', 0x00};
+   unsigned char buf[] = {'C', (unsigned char)(positive ? 0xc0 : 0x80), 'C', 0x00};
```
marktsuchida, line 188: Just to note for the record, these casts to `unsigned char` are probably needed to compile in C++11 (or later) mode -- they should not be needed if `-std=c++03` is added to the compiler flags. But we'd like to move to modern C++ soon, so thanks!

DeviceAdapters/WieneckeSinske/WieneckeSinske.cpp
```diff
@@ -409,7 +409,7 @@ int CAN29Axis::SetTrajectoryVelocity(CAN29Long velocity)
 }
 
 /*
-* Sets acceleration for position moves in nm/s� 
+* Sets acceleration for position moves in nm/s�
```
marktsuchida: Still incorrectly encoded?

DeviceAdapters/WieneckeSinske/WieneckeSinske.cpp
```diff
@@ -437,7 +437,7 @@ int CAN29Axis::GetTrajectoryVelocity(CAN29Long& velocity)
 }
 
 /*
-* Gets acceleration for position moves in nm/s� 
+* Gets acceleration for position moves in nm/s�
```
marktsuchida: Ditto; would be good to just say `nm/s^2`.

plugins/ZProjector/src/main/java/org/micromanager/zprojector/CoordsComparator.java
```diff
@@ -14,12 +14,12 @@
...
-      cAxesNumeric.put(Coords.C, 1);
-      cAxesNumeric.put(Coords.CHANNEL, 1);
+      cAxesNumeric.put(Coords.T, 1);
+      cAxesNumeric.put(Coords.TIME_POINT, 1);
+      cAxesNumeric.put(Coords.C,3);
+      cAxesNumeric.put(Coords.CHANNEL, 3);
       cAxesNumeric.put(Coords.Z, 2);
       cAxesNumeric.put(Coords.Z_SLICE, 2);
-      cAxesNumeric.put(Coords.T, 3);
-      cAxesNumeric.put(Coords.TIME_POINT, 3);
```
marktsuchida, line 22: Perhaps this change was accidentally included?

DeviceAdapters/OxxiusCombiner/Oxxius_combiner.cpp
```diff
@@ -326,7 +326,7 @@ int OxxiusCombinerHub::DetectInstalledDevices()
 			com << "IP" << convertable[j];
 
 			std::ostringstream InfoMessage4;   /////test in hub
-			InfoMessage4 << "test" << com;
+			InfoMessage4 << "test" << com.str();
```
marktsuchida, line 329: Nice catch!

marktsuchida:
P.S. To update your pull request you can push additional commits to your branch (or force-push alternative ones), instead of creating a new PR. Much appreciated!

plugins/ZProjector/src/main/java/org/micromanager/zprojector/CoordsComparator.java
triccyx: Regards to this the code is aligned to the one in origin/master. I think it is safe to keep.

triccyx:
Done apart from:
plugins/ZProjector/src/main/java/org/micromanager/zprojector/CoordsComparator.java

triccyx:
@marktsuchida do you think we can proceed with this pull request despite
plugins/ZProjector/src/main/java/org/micromanager/zprojector/CoordsComparator.java
see above.

marktsuchida:
@triccyx Thanks for the fixes!

PR #728, opened 2020-03-23
No longer use two outdated copies of org.json

nicost:
The copy of org.json in MMCoreJ_wrap has been renamed to mmcorej.org.json and is now included in MMCoreJ_.jar. org. json was deleted from the mmstudio source directory and is no longer include in MMJ_.jar.  All references in our source code to org.json have been changed to mmcorej.org.json.  The project builds on Windows, and I expect to problems on Mac/Unix (but currently can not test).

These changes make it possible to include a maven version of org.json to the project and depend on that when so desired.


marktsuchida:
Looks great! Thanks for taking care of this.

henrypinkard:
@nicost It appears that there are indeed problems with mac. On the MMCoreJ.jar that comes with the Mac nightly builds this change hasn't taken effect. There is still `org.json` but no `mmcorej.json`, so it doesn't work with latest source code. Thoughts?

nicost:
@henrypinkard @marktsuchida That does not make sense at all!  Problem is that I do not have a mac at home, and can not access the Mac build machine, so this is a bit hard to trouble shoot for me, but if the Mac is building from the latest code, there is no way org.json is still there, since it is no longer in the source code....  Thinking about bringing that Mac build machine home, as I also want the XBoxController that is sitting on my desk;)

henrypinkard:
I'm happy to test if that would be helpful. I've never built the core though so you might have to walk me through how to do it

nicost:
Build instructions for Mac are here: https://micro-manager.org/wiki/Build_on_MacOS_X

Bit outdated, but most probably still usable.  Nevertheless, I would be highly surprised if you can reproduce the problem in a new build.  I more suspect there to be an issue specif to the build machine.

marktsuchida:
In my local mac build I do not see this issue.

henrypinkard:
I never managed to debug the build process on my mac, so I just copied MMCoreJ from a windows build instead as a workaround. Sounds like it's specific to the build machine then...

nicost:
@marktsuchida could you test the dmg from the download page as well?  Just to be sure nothing funky is going on.  I could imagine that the Jenkins job on the build machine is not doing a needed clean, but before feeling compelled to go into the lab to trouble shoot, I'd like to be sure;)

marktsuchida:
20200416 nightly looks fine:
```
bsh % org.json.JSONObject;
// Error: EvalError: Class or variable not found: org.json.JSONObject : at Line: 2 : in file: <unknown file> : org .json .JSONObject 

bsh % mmcorej.org.json.JSONObject;
<Class Identifier: mmcorej.org.json.JSONObject>
```

I also checked that `MMCoreJ.i` contains the `mmcorej.org.json` classes and that `MMJ_.jar` does not contain any `org.json`.

nicost:
Thanks @marktsuchida ! Guess that Henry had mixed versions of binaries lying around.

Issue #724, opened 2020-03-12
Display can lag far behind a fast acquisition


nicost:
Test case:
DemoCamera, 256 x 256 2 bytes per image, 1 ms exposure, FastImage property set to 1. 

Run MDA with 2000 images.  Even though the images are acquired in about 2 seconds, the display continues to update for about 10 seconds.  This is unnecessary, and a bit confusing.  Not very high priority, but a bit of an issue for high speed cameras.

nicost:
The Faster saving PR #735 changes the behavior.  No images are displayed until all images are saved.  This happens much faster than 10s, so probably more desirable than the previous behavior, although it would be nice to update the display a few times during the acquisition.

nicost:
Re-open if it is still annoying.

Issue #723, opened 2020-03-12
mm.data().convertTaggedImage() is slow for large images


nicost:
While investigating bottlenecks in the Micro-Manager data pipeline, I discovered that the function mm.data().convertTaggedImage() is rate limiting for large (2k x 2k x 2 byte) images.  The upside is that metadata apparently are not an issue, the downside that this most likely is caused by the conversion of a pixel array into a DirectByteBuffer.  The following script illustrates the issue.  This problem limits the speed with which data can be stored in RAM, hence is a serious limitation in performance.

```
/**
 * Testing which part is slow in handling images in Micro-Manager
 * Use the Democamera, 2048x2048, 16 bit, 5 msec exposure
 * Snap one image, then set the "FastImage" property to 1
 * 
 * 2000 images, so ideally this will take 10000 ms.
 * 
 * when not calling mmc.data().convertTaggedImage() we get:
 * mean: 10021 +- 7 (n=5)
 * 
 * Stage 2 of testing: converting the Tagged Image to a MM Image 
 * This takes considerable time:
 * 17688 (and lost images total 1767)
 * 17999 (2000) 17954 (2000) 17821 (2000) 17616 (2000) 17785 (2000)
 * mean: 17835 +- 151 (n=5)
 * 
 * With 512 x 512:
 * 10021 (2000) 10004  10023 10016 10020
 * mean: 10017 +- 8 (n=5)
 * 
 * 512 x 512 1 ms exposure
 * 2131 2303 2477 2366 2456
 * mean 2347 +- 139
 */

nrFrames = 2000;
exposureMs = mmc.getExposure();

builder = mm.data().getCoordsBuilder().z(0).channel(0).stagePosition(0);
coord = builder.build();

mmc.startSequenceAcquisition(nrFrames, 0, true);

int curFrame = 0;
start = System.currentTimeMillis();
while (mmc.getRemainingImageCount() > 0 || 
			mmc.isSequenceRunning(mmc.getCameraDevice())) {
   if (mmc.getRemainingImageCount() > 0) {
      tagged = mmc.popNextTaggedImage();
      // Convert to an Image at the desired timepoint.
      image = mm.data().convertTaggedImage(tagged,
         coord, null);
      curFrame++;
   }
   else {
      mmc.sleep(Math.min(.5 * exposureMs, 20));
   }
}

end = System.currentTimeMillis();

mmc.stopSequenceAcquisition();

duration = end - start;
[...]

PR #735, opened 2020-03-29
Faster saving

nicost:
This PR addresses slowness in saving data in MM 2.0.  It aims to make it possible again to save images of 2k x 2k x 2 bytes at 100fps (possible a few years ago).  Two bottlenecks were found:
1. Pixel data coming from the core were copied into a Direct Byte Buffer.  This is a slow operation.  All consumers of the pixel data copied them back into a Java array, which kind of made the initial copy useless. This PR keeps the original Java array, and hands that data to "consumers", This obviates two copies over the JVM border. The DefaultImage class was refactored to no longer use Direct Buffers anywhere (as there is no advantage to use them in the current design).  As a side effect, users who have learned to lower the JVM memory requirements will need to be informed this is no longer the best strategy, and that most memory use happens inside the JVM (as in the disrtant past).
2. The Datastore provides an event indicating that a new image arrived.  This event ran synchronously, and took 10-20 ms to execute (by the display animationcontroller and the histogram stats thingie), seriously degrading performance of the saving code.  This PR both makes the eventbus sending the event asynchronous and adds code to the consumers of the event to look at the size of the circular buffers and to return immediately if it is too full.  

These two changes drastically improve saving speed, and enable saving of the 2k x 2k 2 x 2 bytes at 100fps on my laptop with NVME hard drive (which writes data at ~2.4GB/s).  This addresses issue #723 and possibly issue #724.


mmstudio/src/main/java/org/micromanager/data/internal/BufferTools.java
```diff
@@ -1,100 +1,117 @@
...
+      Buffer buffer;
+      switch (bytesPerPixel) {
+         case 1:
+            buffer = ByteBuffer.wrap((byte[]) pixels);
+            break;
+         case 2:
+            buffer = ShortBuffer.wrap((short[]) pixels);
+            break;
+         default:
+            throw new AssertionError("Unimplemented pixel component size");
```
marktsuchida: Seeing how this method is being called, `UnsupportedOperationException` or `IllegalArgumentException` is more appropriate. Or else all callers need to carefully check `pixels` before calling this.

marktsuchida: Oh, I see it was carried over from existing code. ~~It should still be updated, but can be out of scope for this PR.~~

marktsuchida: Ok, now I see the problem: `AssertionError` made sense in the old code, because we had a guarantee that `TaggedImage` only ever contained byte or short arrays, therefore anything else indicates a violation of programmer's assumption. Now that this code is being used for conversions other than `TaggedImage` as well, it definitely needs to change.

mmstudio/src/main/java/org/micromanager/data/internal/DefaultImage.java
```diff
@@ -144,81 +147,91 @@ public DefaultImage(TaggedImage tagged, Coords coords, Metadata metadata)
...
+    * @param pixels Image pixels.  Should be a Java array of bytes or shorts (not null).
+    * @param format PropertyMap specifying image width, height, and pixelType (not null).
+    * @param coords Coords to be used for this new image (can be null).
+    * @param metadata Metadata to be used this new image (can be null).
+    * @throws IllegalArgumentException
+    */
    public DefaultImage(Object pixels, PropertyMap format, Coords coords,
          Metadata metadata) throws IllegalArgumentException {
-      Preconditions.checkNotNull(pixels);
-      Preconditions.checkNotNull(format);
```
marktsuchida, line 166: Why remove these checks?

mmstudio/src/main/java/org/micromanager/data/internal/BufferTools.java
```diff
@@ -1,100 +1,117 @@
...
+import java.nio.ByteBuffer;
+import java.nio.ByteOrder;
+import java.nio.IntBuffer;
+import java.nio.ShortBuffer;
+
+/**
+ *
+ * @author Arthur
+ */
+public final class BufferTools {
```
marktsuchida, line 17: I like this name, too.

mmstudio/src/main/java/org/micromanager/data/internal/DefaultImage.java
```diff
@@ -144,81 +147,91 @@ public DefaultImage(TaggedImage tagged, Coords coords, Metadata metadata)
...
          int numComponents, Coords coords, Metadata metadata) 
          throws IllegalArgumentException {
-      metadata_ = (DefaultMetadata) metadata;
-      if (metadata_ == null) {
-         // Don't allow images with null metadata.
-         metadata_ = new DefaultMetadata.Builder().build();
-      }
-      coords_ = coords;
-
-      rawPixels_ = DirectBuffers.bufferFromArray(pixels);
```
marktsuchida, line 194: Is this constructor also critical for MDA->saving performance?

Unlike the case of coming from `TaggedImage` (where user code is not expected to manipulate the `pix` of this deprecated class -- although perhaps we should document this requirement), it *does* make sense to make a copy of a user-provided pixel array, where we have no idea what changes may be made afterwards (it might, for example, belong to an ImageJ image that can be edited in the GUI).

Of course one could argue that user code also needs high performance, but in that case we really need a separate API that comes with stark warnings against passing in an array that may be modified by other code.

marktsuchida: Let's talk more if we're going to add such an API - we can split `DefaultImage` into `PerformanceImage` and `SafeImage`, for example, and instantiate the appropriate one in the API methods.

mmstudio/src/main/java/org/micromanager/data/internal/multipagetiff/MultipageTiffWriter.java
```diff
@@ -236,11 +257,11 @@ private static ByteBuffer allocateByteBuffer(int capacity) {
    }
 
    private static final int BUFFER_POOL_SIZE =
-         System.getProperty("sun.arch.data.model").equals("32") ? 0 : 3;
+         System.getProperty("sun.arch.data.model").equals("32") ? 0 : 10;
```
marktsuchida: Please look at the commit that introduced this limit (f34c2b9a9e6b3d7adde2cedc7576fd09d32bba3b). This change will cause memory issues with multiposition acquisition.

mmstudio/src/main/java/org/micromanager/display/inspector/internal/panels/intensity/IntensityInspectorPanelController.java
```diff
@@ -587,18 +587,22 @@ public void onEvent(DisplaySettingsChangedEvent e) {
...
    @Subscribe
    public void onEvent(DataProviderHasNewImageEvent event) {
-      final int channel = event.getImage().getCoords().getChannel();
-      SwingUtilities.invokeLater(() -> {
-         try {
-            if (channel >= channelControllers_.size()) {
-               setUpChannelHistogramsPanel(channel + 1);
+      // (NS - 2020-03-27)
+      // Hack: handle only if the circular buffer is not too full.  How full is highly arbitrary!
+      if ( !studio_.acquisitions().isAcquisitionRunning() || studio_.core().getRemainingImageCount() < 6 ) {
```
marktsuchida: As we discussed on the phone I'm very much for using the sequence buffer fill level to regulate display rate. When things are working smoothly the buffer stays at 0-2 images most of the time, so 6 seems reasonable until proven otherwise.

I think there is one set of edge cases where this will cause problems: MDAs can also be run with snaps only (making `isAcquisitionRunning` return `true` even if no sequence acquisition is running). Yet a previous sequence acquisition could have left more than 5 images in the sequence buffer (especially if the previous sequence acquisition was Live).

mmstudio/src/main/java/org/micromanager/display/internal/displaywindow/DisplayController.java
```diff
@@ -817,20 +817,23 @@ public void setAxisAnimationLock(String axis, String state) {
...
    @Subscribe
    public void onNewImage(final DataProviderHasNewImageEvent event) {
-      perfMon_.sampleTimeInterval("NewImageEvent");
-
-      synchronized (this) {
-         if (closeCompleted_) {
-            return;
+      // (NS - 2020-03-27)
+      // Hack: handle only if the circular buffer is not too full.  How full is highly arbitrary.
+      if ( !studio_.acquisitions().isAcquisitionRunning() || studio_.core().getRemainingImageCount() < 6 ) {
```
marktsuchida: (Please see my comment at IntensityInspectorPanelController.)

mmstudio/src/main/java/org/micromanager/internal/utils/PrioritizedEventBus.java
```diff
@@ -29,7 +45,11 @@ public void register(Object o) {
    public void register(Object o, Integer priority) {
       EventBus subBus;
       if (!prioritizedBuses_.containsKey(priority)) {
-         subBus = new EventBus(EventBusExceptionLogger.getInstance());
+         if (async_) {
+            subBus = new AsyncEventBus(executorService_, EventBusExceptionLogger.getInstance());
```
marktsuchida, line 49: (Mostly from curiosity), did it turn out that async event bus actually improves performance?

mmstudio/src/main/java/org/micromanager/data/internal/BufferTools.java
nicost: Changed to UnsupportedOperationException in 2ee6b78c8

mmstudio/src/main/java/org/micromanager/data/internal/DefaultImage.java
nicost: Decided to call another constructor from this one, so nothing can go before.  Not null check for pixels happens in that other constructor, information from format is checked for sanity as well.  If format is null, the caller will get a NullPointer Exception either way.

mmstudio/src/main/java/org/micromanager/data/internal/BufferTools.java
nicost: Also note this class was moved from utils.

mmstudio/src/main/java/org/micromanager/data/internal/DefaultImage.java
nicost: No, I was updating all other constructors in DefaultImage to stop using DirectBuffers.  This specific constructor is used by `DataManager.createImage(Object pixels, int width, int height, int bytesPerPixel, int numComponents, Coords coords, Metadata metadata);`. where the documentation says: "Generate a new Image with the provided pixel data, rules for interpreting that pixel data, coordinates, and metadata.", suggesting that pixel data are used without copying.  It is also used to insert blank images by displaywindow.imagej. 

I suggest deciding what the behavior of data().creatImage should be.  There definitely should be an internal method that re-uses pixels, and I would argue there also should an external one that does not copy.

mmstudio/src/main/java/org/micromanager/data/internal/multipagetiff/MultipageTiffWriter.java
nicost: Ah!  That explanation helps a lot!  I updated the comments to reflect this.  

Initially, I thought that this number made a difference (i.e., I got circular buffer overflow), but increasing the circular buffer size by about 10 images solved that issue (I must have been working very close to the limit).  Reverted to 3.

mmstudio/src/main/java/org/micromanager/display/inspector/internal/panels/intensity/IntensityInspectorPanelController.java
nicost: Great catch!  I added a check to isSequenceRunning, and at least the edge case you describe is now working.  I hope that there are not more....

mmstudio/src/main/java/org/micromanager/display/internal/displaywindow/DisplayController.java
nicost: Same fix.

mmstudio/src/main/java/org/micromanager/internal/utils/PrioritizedEventBus.java
nicost: My testing was not intensive, but it did not seem to make much of a difference.  Decided to do it asynchronous in any case, as this event is public and plugin code consuming the event may want to take its time.

nicost:
Let me know if you are OK with the code in DefaultImage or that you would like changes now.  

At first I was copying pixel data in at least some of these constructors, but the available documentation suggested this is not what is expected.  We should be very explicit in our api whether or not pixels and metadata are copied.

mmstudio/src/main/java/org/micromanager/data/internal/multipagetiff/MultipageTiffWriter.java
nicost: Just re-read the commit message, and realized that there is actually only one such buffer, so I doubt there was an issue with multiposition acquisition to begin with (as there was only one buffer shared between all file writers, this only increased the size of the memory used by 7 images worth of RAM).  Nevertheless, 3 seems to do the job, so let's keep it there.

mmstudio/src/main/java/org/micromanager/data/internal/DefaultImage.java
marktsuchida: Of course.... Sorry I missed that.

mmstudio/src/main/java/org/micromanager/data/internal/DefaultImage.java
marktsuchida: I'm not sure why "Generate a new Image" suggests no copying. The behavior of `createImage` has always been to copy, and changing this is not safe.

Anyway, how about `createImage` always making a copy and adding a `wrapImmutablePixels` that doesn't copy?

mmstudio/src/main/java/org/micromanager/data/internal/multipagetiff/MultipageTiffWriter.java
marktsuchida: You're right -- I guess the commit in question also switched to a shared pool of buffers. I seem to be having trouble paying attention today.

mmstudio/src/main/java/org/micromanager/data/internal/DefaultImage.java
nicost: "Generate a new Image with the provided pixel data" very much suggests that no copying takes place.  In general, the documentation has to be explicit about the desired behavior (again, I first had those functions copying pixels, but decided to change back to match the documentation).  Documentation to the convertTaggedImage functions in DataManager should also make clear whether or not pixel data are copied. 
I don't know what a `wrapImmutablePixels ` function should look like or what it should do. Why not add a flag to `createImage`, or if you do not like that make a `createImageCopy` that does the same as `createImage` but copies pixels?

mmstudio/src/main/java/org/micromanager/display/inspector/internal/panels/intensity/IntensityInspectorPanelController.java
marktsuchida: Well, I think that covers most cases. Because of the time delay between acquisition and this code, an MDA that mixes snap and sequence acquisition might have strange display behavior, but that's a harder one to fix. 

The other issue that I would worry about is that the last image(s) of an acquisition may never reach the display -- but I think you would have seen that in your testing, so I'm assuming some other mechanism is covering that case?

mmstudio/src/main/java/org/micromanager/internal/utils/PrioritizedEventBus.java
marktsuchida: Sounds good.

marktsuchida:
@nicost The documentation was never very clear, but I think the principle of least surprises should apply. (And in this context, not corrupting data is much more important than performance.)

I don't see how we can prevent nasty bugs without making 100% sure that buffers that may later be modified are never used to back an `Image`, and that any array backing an `Image` is never modified. And by "nasty bugs", I'm talking about e.g. noticing after many weeks that the files that should have contained raw data instead contain some processed images. The fact that this may not appear to happen in a few (or even a few dozen) tests guarantees nothing. A system that might silently overwrite images due to unfortunate and hard-to-predict interactions between two unrelated pieces of code cannot possibly be better than a system that doesn't offer maximum performance 100% of the time.

(In fact, it might be ideal to even keep the copying behavior of conversion from `TaggedImage`, instead adding a new method that doesn't copy but has an explicit name, such as `convertImmutableTaggedImage`. It will be safe to use the latter for MDA, of course.)

mmstudio/src/main/java/org/micromanager/data/internal/DefaultImage.java
marktsuchida: What I was suggesting was that `wrapImmutablePixels` should do what your current `createImage` implementation is doing, i.e., wrap the array without making a copy.

I'm proposing that it should be safe for users to use `Image` without necessarily understanding that it needs to be immutable (that would be a lot to expect from casual, first-time script writers). All of the methods were explicitly designed to protect such users (and other people that use code written by such users), which is why a copy was being made upon construction.

If the natural-sounding `createImage` has scary data integrity issues, I wouldn't trust any data handled by that system. I think most of the world would agree that scientific data integrity is important, and I'm sure you do, too, so I must be failing to convey how scary the possibilities are.

A flag has two downsides:
- Looking at the code, `createImage(foo, false)` conveys nothing until you look at the documentation, and I can't remember which of `true` or `false` skips the copy.
- Again, anything named `createImage` should not do dangerous things.

I'm proposing `wrapImmutablePixels` specifically because it sounds scary, which it what it is.

nicost:
@marktsuchida I very much disagree.  We have been suffering from drastic lack of performance, in large part because of multiple extra copy steps.  We really should have a pipeline that does not do superfluous copies of massive amounts of data (with data rates of 1-2GB/s this can become serious).  Since our own code is using the API, the API needs to provide methods that reuse pixels rather than copy them.  I am fine with having two versions of each function (or using a flag in one function), and the expected behavior with respect to copying pixels should clearly documented, but if we only take in pixels by copying them, we sign our own death warrant.

nicost:
@marktsuchida Added the wrapImage function to the DataManager, changed the behavior of createImage as we discussed.

As you brought up, manipulations of displayed image in ImageJ are destructive for data stored in RAM (but not on disk).  A brute force method would be to change the flag on:
`org.micromanager.display.internal.displaywindow.imagej.MMVirtualStack` line 131 from false to true.  However, it would be nicer to only copy pixels residing in a RAM store, but I can not easily find that information in the relevant place.  Or possibly images in a RAM store should always hand out pixel data copies.  Not sure what the best approach is here.

marktsuchida:
@nicost Thanks!

I think one extra copy for display can be tolerated for now. In fact, if editing in ImageJ rewrites RAM data, it probably also has the potential to rewrite any in-flight images from a disk datastore (likely only observed under special circumstances), so I don't think there is any way around making a copy. (Any modifications won't get written back to disk, but may appear in any other process that happens to fetch the same images around the same time.)

PR #742, opened 2020-04-03
Studio: duplicates pixels before handing them to ImageJ.

nicost:
As discussed in previous PR.  We can change this again in the future, but for now this will protect our images from accidental changes in ImageJ.


marktsuchida:
Thanks!

Issue #736, opened 2020-03-31
Live Camera fps values are only valid until the circular buffer is reset


nicost:

nicost:
Fixed in PR #743.  However, estimating camera fps is a bit of a black art given the tools available though the core api.  I wonder if the core could do a better job generating the fps number.

PR #743, opened 2020-04-03
Fps fix

nicost:


nicost:
Fixes issue #736

nicost:
Fixes issue #736

PR #756, opened 2020-04-06
Libraries + pycro-manager

henrypinkard:
This needs to be cleaned up before merging, but thought I'd get it started anyway. Contains

- N-D multires viewer library
- N-D multires storage library
- New acquisition engine library
- Many changes to magellan now that these libraries are factored out
- changes to ZMQ server
- addition of new package `org.micromanager.remote`, which makes new package `pycromanager` (successor to `pygellan`) possible. The python side of this is already on PyPi and documentation is [live](https://pycro-manager.readthedocs.io/en/latest/index.html)

What needs to happen before merge:
- [x] fix conflicts
- [ ] split out Viewer library into its own repo and add to build
- [ ] split out Storage library into its own repo and add to build
- [ ] split out acquisition library into its own repo and add to build


henrypinkard:
I've now split out viewer, storage, and acq eng into their own repos:

https://github.com/henrypinkard/AcqEngJ
https://github.com/henrypinkard/NDViewer
https://github.com/henrypinkard/NDTiffStorage

The only thing remaining before this can be merged is that the jars for these libraries must be added to the build. @nicost @marktsuchida do you think that these should be compiled or added to build automatically? 

Related, would be good to add these repos to micro-manager organization

nicost:
Hope that @marktsuchida can chime in, but it makes most sense to me to move those 3 repos to micro-manager, and then add them to the Jenkins builds.  The difficult part may be the order of building things, since I assume that the 3 repos all depend on the micro-manager repo, but the latter contains the Magellan plugin code, which depends on these 3 repositories. This is compounded by only me having access to the Windows build machine, and nobody to the Mac build machine;).  Waiting for direction from @marktsuchida...

henrypinkard:
the 3 repos only depend on mmcorej, in case that helps...

henrypinkard:
got `make fetchdeps` to run successfully, but then `make` failed on compiling mmcorej with a bunch of errors like this:


```
MMCoreJ.compile:
 [mm-javac] Compiling 29 source files to /Users/henrypinkard/Google Drive/Code/GitRepos/micro-manager/MMCoreJ_wrap/build
 [mm-javac] /Users/henrypinkard/Google Drive/Code/GitRepos/micro-manager/MMCoreJ_wrap/gensrc/mmcorej/CMMCore.java:231: error: incompatible types: Long cannot be converted to int
 [mm-javac]          Rectangle r = new Rectangle((int) xs.get(i), (int) ys.get(i),
 [mm-javac]                                                  ^
 [mm-javac] /Users/henrypinkard/Google Drive/Code/GitRepos/micro-manager/MMCoreJ_wrap/gensrc/mmcorej/CMMCore.java:231: error: incompatible types: Long cannot be converted to int
 [mm-javac]          Rectangle r = new Rectangle((int) xs.get(i), (int) ys.get(i),
 [mm-javac]
```

Any ideas?

nicost:
Maybe good to move this into a separate issue.
Is it possible that you are using another swig version? The difference lies in the UnsignedVector type, which is a data structure generated by swig.  You can find the generated sources in your build (in my case in build/intermediates/Swig.mmcorej.  My version return a long for the "get(int)" function, apparently, yours returns a long, which can not be case to int. Yikes. 

Hacks are possible to work around this, but those will depend on swig version, so it may be best to use the same swig version (use "swig -version" mine is 3.0.12)

kdmarchuk:
Hey All, this looks awesome! 

@marktsuchida Any idea when this might be available in nightly builds?

Cheers!

henrypinkard:
@nicost @marktsuchida Anything helpful I can do to get this into the build?

nicost:
I now have branches (all called mvnBuild) for AcqEngJ, NDViewer, NDTiffStorage, and pycro-manager that build the (Java part of) code, and deploy it to bintray.  I also have a branch (mavenCore) in micro-manager that fetched all 4 projects as dependencies and places them in dependencies/artifacts/compile (which will be copied to plugins/Micro-Manager in the binary distributions).  The main tricky thing is that Micro-Manager itself should not import MMCoreJ-10.1.0.0 that these 4 projects depend on (and I figured out how to do that).  Still somehwhat ugly that there are two ways to generate MMCoreJ.jar (one through maven, which these 4 projects depend on, the other through the normal build process: ant on Windows, make on Unix).  Those jars should be equivalent, but need manual intervention to keep them synchronous.  The best thing is if the next iteration of the Micro-Manager build process generates the mavenized MMCoreJ and distributes it whenever the version changes.

I looked a bit at this specific PR, but it contains a lot of thing, some of which definitely should not be merged.  I'll first open PRs for the 4 projects (and one for micro-manager).  Let's merge those first, and then have @henrypinkard generate a new PR based on that code.

henrypinkard:
Awesome! Closing this PR and will remake a new one

PR #772, opened 2020-04-12
MMCoreJ throws specific exceptions instead of `java.lang.Exception`

### LINKED ISSUE #766: MMCoreJ should throw a specific exception type instead of `Exception`
The majority of the methods in MMCoreJ are defined with `throws Exception` when really there are probably only a narrow set of exception types that each method can throw. It is generally agreed that it is best to keep try-catch statements as narrowly defined as possible to avoid inadvertently catching unchecked exceptions that you would really like to allow to propagate. An example of this are `InterruptedException`s which are used to cancel an operation running a separate thread. For example image you want to set a property and log an error if it occurs. The obvious thing to write is:

    try {
         doStuff();
         core.setProperty(devName, propName, propVal);
    } catch (Exception e) {
        ReportingUtils.logError(e);
    }

but if this code is run on a separate thread and the thread is interrupted you will find that this broad `catch` statement will catch the `InterruptedException` and log it rather than allow it to actually cancel the thread.

This can potentially be solved by specifically catching any unchecked exceptions that you expect might crop up and then rethrowing them..

    try {
         doStuff();
         core.setProperty(devName, propName, propVal);
    } catch (InterruptedException) {
       Thread.currentThread().interrupt();
       throw new InterruptedException();
    } catch (Exception e) {
        ReportingUtils.logError(e);
    }

However it would be much better if exceptions from MMCoreJ could be narrowly caught allowing unchecked exceptions to automatically propagate:

    try {
         doStuff();
         core.setProperty(devName, propName, propVal);
    } catch (CMMException e) {
        ReportingUtils.logError(e);
    }

nanthony21:
This PR resolves #766.

Three generic subclasses of `java.lang.Exception` have been created which mirror the C++ exceptions in MMCore. Methods in MMCoreJ will now throw these specific exceptions instead of throwing `java.lang.Exception`. This will give code that calls methods from MMCoreJ more flexibility in exception handling and will discourage the bad practice of broad catch statements:

    try {
        someCode()
    } catch (Exception e) {
        logError()
    }

which can result in improper handling of other unexpected errors such as `InterruptedException`.

### DIFF
--- a/MMCoreJ_wrap/CMMError.java
+++ b/MMCoreJ_wrap/CMMError.java
@@ -1,0 +1,8 @@
+package mmcorej;
+
+public class CMMError extends java.lang.Exception {
+	public CMMError() { super(); }
+	public CMMError(String message) { super(message); }
+	public CMMError(String message, Throwable cause) { super(message, cause); }
+	public CMMError(Throwable cause) { super(cause); }
+}
--- a/MMCoreJ_wrap/MMCoreJ.i
+++ b/MMCoreJ_wrap/MMCoreJ.i
@@ -599,24 +599,24 @@
 %rename(eql) operator=;
 
 // CMMError used by MMCore
-%typemap(throws, throws="java.lang.Exception") CMMError {
-   jclass excep = jenv->FindClass("java/lang/Exception");
+%typemap(throws, throws="mmcorej.CMMError") CMMError {
+   jclass excep = jenv->FindClass("mmcorej/CMMError");
    if (excep)
      jenv->ThrowNew(excep, $1.getFullMsg().c_str());
    return $null;
 }
 
 // MetadataKeyError used by Metadata class
-%typemap(throws, throws="java.lang.Exception") MetadataKeyError {
-   jclass excep = jenv->FindClass("java/lang/Exception");
+%typemap(throws, throws="mmcorej.MetadataKeyError") MetadataKeyError {
+   jclass excep = jenv->FindClass("mmcorej.MetadataKeyError");
    if (excep)
      jenv->ThrowNew(excep, $1.getMsg().c_str());
    return $null;
 }
 
 // MetadataIndexError used by Metadata class
-%typemap(throws, throws="java.lang.Exception") MetadataIndexError {
-   jclass excep = jenv->FindClass("java/lang/Exception");
+%typemap(throws, throws="mmcorej.MetadataIndexError") MetadataIndexError {
+   jclass excep = jenv->FindClass("mmcorej/MetadataIndexError");
    if (excep)
      jenv->ThrowNew(excep, $1.getMsg().c_str());
    return $null;
--- a/MMCoreJ_wrap/MMCoreJ.i
+++ b/MMCoreJ_wrap/MMCoreJ.i
@@ -622,7 +622,7 @@
    return $null;
 }
 
-// We've translated exceptions to java.lang.Exception, so don't wrap the unused
+// We've translated exceptions to java versions that extend java.lang.Exception, so don't automatically wrap the unused
 // C++ exception classes.
 %ignore CMMError;
 %ignore MetadataKeyError;
--- a/MMCoreJ_wrap/MMCoreJ.i
+++ b/MMCoreJ_wrap/MMCoreJ.i
@@ -608,7 +608,7 @@
 
 // MetadataKeyError used by Metadata class
 %typemap(throws, throws="mmcorej.MetadataKeyError") MetadataKeyError {
-   jclass excep = jenv->FindClass("mmcorej.MetadataKeyError");
+   jclass excep = jenv->FindClass("mmcorej/MetadataKeyError");
    if (excep)
      jenv->ThrowNew(excep, $1.getMsg().c_str());
    return $null;
--- a/MMCoreJ_wrap/MMCoreJ.i
+++ b/MMCoreJ_wrap/MMCoreJ.i
@@ -850,7 +850,7 @@
    }
 
    /**
-    * Convenience function.  Retuns affine transform as a String
+    * Convenience function.  Returns affine transform as a String
     * Used in this class and by the acquisition engine 
     * (rather than duplicating this code there
     */
--- a/MMCoreJ_wrap/MMCoreJ.i
+++ b/MMCoreJ_wrap/MMCoreJ.i
@@ -698,7 +698,7 @@
 
    }
 
-   private TaggedImage createTaggedImage(Object pixels, Metadata md, int cameraChannelIndex) throws java.lang.Exception {
+   private TaggedImage createTaggedImage(Object pixels, Metadata md, int cameraChannelIndex) throws mmcorej.CMMError, mmcorej.org.json.JSONException {
       TaggedImage image = createTaggedImage(pixels, md);
       JSONObject tags = image.tags;
       
@@ -716,7 +716,7 @@
       return image;
    }
 
-   private TaggedImage createTaggedImage(Object pixels, Metadata md) throws java.lang.Exception {
+   private TaggedImage createTaggedImage(Object pixels, Metadata md) throws mmcorej.CMMError, mmcorej.org.json.JSONException  {
       JSONObject tags = metadataToMap(md);
       PropertySetting setting;
       Configuration config = getSystemStateCache();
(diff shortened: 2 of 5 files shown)


nicost:
Looks great to me.  Thanks!
@marktsuchida  OK to merge?

marktsuchida:
Sorry for the late comment; wish I could have replied before you went to the trouble of creating this PR. I agree that throwing plain `Exception` is really bad and it has been bothering me, too, but I also don't think associating exception types to the source of error, as opposed to the meaning of the error, is a good idea.

A key limitation of Java's design is that you cannot broaden the set of checked exceptions thrown by an API method without breaking all code that calls the method in question.

Ideally, the exceptions would inherit from the appropriate type within the Java exception hierarchy. So for example, calling `mmc.setExposure(-10)` should throw `IllegalArgumentException` (or a subclass thereof), while `mmc.setProperty("MyStage", "Velocity", 100)` should probably throw `IOException` (or a subclass) if the device failed to set the property.

(`JSONException` is not an exemplary design in this regard.)

The problem with introducing a blanket `CMMError` type now (even though in the short term it is a net improvement) is that the above, fine-grained, idiomatic interface will become impossible to implement later on. (It also should not be named "Error", which, in the Java world, is reserved for severe system-level errors.)

Note especially that, in the above example, `IllegalArgumentException` is an unchecked exception that usually indicates a programming error (for developers to diagnose bugs; not usually requiring explicit handling) whereas `IOException` is something that may well happen in front of a user, and which calling code should catch and report (too often our current code logs and ignores such errors). Being able to distinguish between such exceptions is quite valuable.

If we keep the current plain `Exception`, at least we leave open the possibility of getting to a proper interface in the future. Unfortunately it's not a trivial amount of work to add appropriate exception types to all the methods in the Core (this is pure technical debt), although it should be possible in principle. I haven't thought through all the potential challenges.

There might be some happy middle ground that I haven't thought of. Perhaps something like having all methods throw `IOException` for now -- this might be acceptable if we can be reasonably sure that none of the Core methods throw any exceptions that ought to be represented by a checked exception other than `IOException`. I'll have to think about this a little more.

I also haven't yet fully reviewed the situation with the other exceptions (`MetadataKeyError` et al.) -- we need to check if/when they are actually thrown all the way to Java and make sure whatever we do with them is compatible with future improvements. IIRC, metadata arrays were never implemented (and probably are so broken that shouldn't be implemented) so `MetadataIndexError` may never be used(?). `JSONException` might be best to translate to the appropriate, probably unchecked, Java exception class.

nanthony21:
I can see what you mean about how accepting this minor improvement now could make moving to a more complete solution more difficult in the future. This would depend on what that hypothetical future solution looked like. If the specific error types were all subclasses of CMMError then I don't think it would be a problem.

As far as naming goes, I had considered naming the Java exception `CMMException` in order to stick with Java convention but ended up deciding to just keep the name the same as it is in C++. This is of course a trivial thing to fix.

The JSONExceptions that some of the methods throw are the result of using `JSONObject.put()`. Since I think we can be highly confident that that exception won't actually be thrown in this case I think we could catch it internally and remove it from the MMCoreJ API.

As far as I can tell you are correct that the MetaData Errors are never actually used anywhere in MMCoreJ

marktsuchida:
The problem is that any good Java library (or one that aspires to be one) needs to throw exceptions in, or derived from, the standard JDK exception hierarchy (such as `java.io.IOException`, `java.lang.IllegalArgumentException` -- I didn't mean these to be hypothetical). So it is not even theoretically possible for them to also inherit from an arbitrary other type (`CMMException`).

(See _Effective Java_, 3rd ed., Item 72, for background.)

I think we can change _most_ of the `throws Exception` to `throws IOException`. For the Core methods that do not do non-microscope I/O (such as reading/writing files), we can even say `throws DeviceIOException`, where `mmcorej.DeviceIOException` is our own subclass of `IOException`. For those methods, errors that are not device-related are mostly programming errors such as invalid arguments, and those can be represented by unchecked exceptions, meaning that we don't need to design them into our `throws` clauses.

There are some methods in the Core that need to throw other checked exceptions (stuff like allocating the sequence buffer or loading device adapter libraries). One way to move forward might be to leave those for later (i.e. keep `throws Exception` for the first iteration), instead focusing on the majority of methods that are purely device-control, making them throw `DeviceIOException`.

So for example `setProperty` would say `throws DeviceIOException`. Then, in an ideal world, `setProperty` would throw a `DeviceIOException` when the device adapter returns an error and `IllegalArgumentException` when MMCore detects an out-of-range value. But that distinction cannot be made in the MMCoreJ layer and would only be possible by carefully revising every method in `CMMCore` (and worse, some of the code might be in `MMDevice` from which only error codes can be returned). So a pragmatic near-term solution is to just throw `DeviceIOException` in all cases. (It's not ideal that the programming errors may later change to throw an unchecked exception, but that's not nearly as bad as the current situation of plain `Exception` for everything.)

I would need to look through every single method to propose more details about what might be done for the methods that have failure modes other than device adapter errors. But even if we leave those all throwing plain `Exception`, I think there will still be a significant net benefit to most code that uses the Core for device control. In particular, it addresses the problem you pointed out as long as the user code only calls pure device control methods.

The JSON errors (which I agree should not normally occur) can simply be wrapped in `RuntimeException` and rethrown without the need for a specific `throws` clause.

nanthony21:
That all makes sense. I've replaced the `JSONExceptions` with undeclared `RuntimeExceptions` and I've entirely replaced `CMMError` with `DeviceIOException`.

The issue of throwing exceptions other than `DeviceIOException` is something that I'm not quite sure about approaching.

Also I realized that while `MetadataIndexError` is entirely unused, `MetadataKeyError` is thrown by `Metadata.GetSingleTag` and `Metadata.GetArrayTag`. Should just be changed back to `Exception` for now?

marktsuchida:
@nanthony21 Apologies for the long blank. Below is what I meant by the need for checked exceptions other than `DeviceIOException`.

I went through every single method that MMCoreJ exposes (looking at a diff of the Swig-generated Java sources before and after this PR), and here are the ones that remain problematic:

These methods need to be marked `throw java.io.IOException` because they access disk files:
- `saveSystemState`
- `loadSystemState`
- `saveSystemConfiguration`
- `loadSystemConfiguration`
- `setPrimaryLogFile` (two overloads)
- `startSecondaryLogFile` (3 overloads)
- `stopSecondaryLogFile`
- `getDeviceAdapterNames`
- `getDeviceLibraries`

Methods that call `cbuf_->Initialize()` (directly or indirectly) need to throw another exception, perhaps a `mmcorej.BufferAllocationException` (a subclass of `Exception` but not of `IOException`):
- `setCircularBufferMemoryFootprint`
- `initializeCircularBuffer`
- `startSequenceAcquisition`
- `startContinuousSequenceAcquisition`
We want to avoid `OutOfMemoryError` because that would be treated as an unrecoverable error; often we can recover from somebody setting the sequence buffer huge. Also the error might be that the user-requested buffer size is smaller than the size of a single frame.

Sequence buffer underflow is also a non-device (non-IO) error, but one that occurs during normal operation so should be a checked exception. Perhaps introduce `mmcorej.BufferUnderflowException`.
- all `popNext*Image*` methods
- all `get*Last*Image*` methods
(Overflow is currently reported via `isBufferOverflowed` so does not appear here.)

I think this will be ready to merge if you can make the above changes.

There are quite a few other methods that do not need `throw DeviceIOException` (because they theoretically will never throw due to their specification, no matter how we may change the implementation in the future). Examples are `getDeviceName` and `hasProperty`, which can throw only an (unchecked) `NullPointerException` or `InvalidArgumentException`. It will be much more work to update all of these (they need to be properly documented in MMCore first), but they can be fixed later and in stages (because removing the `throw` will not break user code). I'm making a note to myself to create an issue for this once this PR is merged.

nanthony21:
Thanks @marktsuchida .

I'm a little unsure of how to go about individually declaring these methods to throw different exception types in SWIG.

Would it make sense to do this differentiation of exception types on the C++ level and then just map each C++ exception type to a Java exception in the same way that is already being done?

the `popNext*Image*` and `get*Last*Image*` methods are defined in the SWIG file so I could easily change those but if I understand correctly it would make the most sense to have the `mmcorej.BufferUnderflowException` thrown originally from `getLastImageMD`, `popNextImageMD`, etc.

--Nick

marktsuchida:
@nanthony21 To be honest, I don't know what the most expedient way would be. I might need to study the Swig manual.

I'm also not yet entirely sure about what the exceptions should be at the C++ level: they should certainly be more specific, but we don't want to model them too closely with Java exceptions, because we want it to work with Python, too. Some translation in the Swig layer is inevitable.

nanthony21:
I'm not familiar with the Python side of MMCore but here is a little table I threw together that we can work from and refine until it's finalized. A separate issue will be to determine what the inheritance of custom exceptions should be.

| C++ | Java | Python |
|-------|-------|--------|
| CMMError | mmcorej.DeviceIOException | mmcorepy.DeviceIOException |
| MetadataIndexError | mmcorej.MetadataIndexError | mmcorepy.MetadataIndexError  or `IndexError` |
| MetadataKeyError | mmcorej.MetadataKeyError | mmcorepy.MetadataKeyError or `KeyError` |
| std::underflow_error | mmcorej.BufferUnderflowException |  mmcorepy.BufferUnderflowException or `BufferError` |
| std::bad_alloc | mmcorej.BufferAllocationException | mmcorepy.BufferAllocationException |
| std::io_base::failure | java.io.IOException | `IOError` |

Links to standard error types:
https://docs.python.org/3.8/library/exceptions.html
https://en.cppreference.com/w/cpp/error/exception

marktsuchida:
@nanthony21 Thanks. I think we want to avoid the C++ `std::` exceptions for recoverable errors. For example, `bad_alloc` may happen (whether we like it or not) when some random memory allocation fails (resulting in a crash), but for the sequence buffer we want to throw an exception type of our own that is a subclass of `CMMError` (or whatever we rename our base exception class to).

Anyway, lots to think about. I think we might need to start by updating `CMMCore` to throw just slightly more specific exceptions (a few subclasses of `CMMError`). But we also need to think about the future of our C++ code because the C++ `throw` exception specifications are deprecated since C++11 and removed in C++17 (see https://en.cppreference.com/w/cpp/language/except_spec). One question is how recent versions of Swig recommend dealing with this (or do we need to manually specify for each method after all?).

PR #790, opened 2020-04-23
Add RTIntensities plugin

tronar:
Simple code to plot intensities of ROIs during acquisition

### DIFF
--- a/DeviceAdapters/ITC18/MMITC18.cpp
+++ b/DeviceAdapters/ITC18/MMITC18.cpp
@@ -131,15 +131,17 @@ initialized_ (false)
     SetErrorText(ERR_ITC18_NO_CHANNELS,"ITC18 needs at least one channel activated");
     SetErrorText(ERR_PROTOCOL_NOT_LOADED,"No protocol loaded to run");
 
-    #ifndef WIN32
+#ifdef __linux__
+
     hITC18Dll = dlopen("libitc18.0.1.so", RTLD_LAZY|RTLD_GLOBAL);
     if (!hITC18Dll)
     {	
         LogMessage("CITC18Hub(): Failed to find libitc18.0.1.so library",false);
         //fprintf(stderr,"CITC18Hub(): Failed to find itc18 library\n");
-        exit(ERR_ITC18_LIBRARY);
+        fprintf(stderr, "CITC18Hub(): %s\n", dlerror());
+	exit(ERR_ITC18_LIBRARY);
     }
-    #else
+#else
     hITC18Dll = ::GetModuleHandle("itcmm.dll"); 
     if (!hITC18Dll)
     {	
@@ -147,7 +149,7 @@ initialized_ (false)
         //fprintf(stderr,"CITC18Hub(): Failed to find itc18 library\n");
         exit(ERR_ITC18_LIBRARY);
     }
-    #endif
+#endif
     g_seqThread = new AcqSequenceThread(this);
 
 }
@@ -156,10 +158,10 @@ CITC18Hub::~CITC18Hub()
 {
     Shutdown();
 
-    #ifndef WIN32
+#ifdef __linux__
     if (hITC18Dll) dlclose(hITC18Dll);
     hITC18Dll = NULL;
-    #endif
+#endif
 
     g_seqThread->Stop();
     //g_seqThread->wait();
@@ -191,7 +193,8 @@ int CITC18Hub::Initialize()
     int busy;
     std::ostringstream eMsg;
     
-    // Name
+
+// Name
     int nRet = CreateProperty(MM::g_Keyword_Name, g_DeviceNameITC18Hub, MM::String, true);
     if (DEVICE_OK != nRet)
         return nRet;
@@ -312,6 +315,8 @@ CITC18Protocol::CITC18Protocol() :
                     frames_(1),
                     channels_(1),
                     slices_(1),
+                    zStep_um_(0),
+                    zStep_ms_(0),
                     minTime_(0),
                     maxTime_(0)
 {
@@ -381,7 +386,18 @@ int CITC18Protocol::Initialize()
 
     pAct = new CPropertyAction(this, &CITC18Protocol::OnSlices);
     CreateProperty("Slices","1", MM::Integer, true, pAct);
-    CreateProperty("Slices Tip1", "Slices currently not used", MM::String, true);
+
+    pAct = new CPropertyAction(this, &CITC18Protocol::OnStepSizeUm);
+    CreateProperty("StepSizeUm","0", MM::Float, true, pAct);
+    
+    pAct = new CPropertyAction(this, &CITC18Protocol::OnStepLowerLimitUm);
+    CreateProperty("StepLowerLimitUm","0", MM::Float, true, pAct);
+    
+    pAct = new CPropertyAction(this, &CITC18Protocol::OnStepUpperLimitUm);
+    CreateProperty("StepUpperLimitUm","0", MM::Float, true, pAct);
+
+    pAct = new CPropertyAction(this, &CITC18Protocol::OnStepTimeMs);
+    CreateProperty("StepTimeMs","0", MM::Integer, true, pAct);
 
     pAct = new CPropertyAction(this, &CITC18Protocol::OnChannels);
     CreateProperty("Channels","1", MM::Integer, true, pAct);
@@ -733,7 +749,13 @@ int CITC18Protocol::ParseHeader(char *line_)
      std::string line = line_;
      std::string line2;
      length = line.length();
+     slices_ = 1;
+     zStep_um_ = 0;
+     zStep_ll_um_ = 0;
+     zStep_ul_um_ = 0;
+     zStep_ms_ = 0;     
                
+     
           if (line[0] == '@')
           {    // these are some official xmgrace params that we will use for our own as well
                
@@ -761,10 +783,26 @@ int CITC18Protocol::ParseHeader(char *line_)
                } else
                if (line.find("# SLICES ") != std::string::npos )  {
                     if (length > 9) {
-                         line2 = line.substr(9,length - 9);
-                         slices_ = atoi(line2.c_str());
-                         if (slices_ < 1) slices_ = 1;
-                    }
+                        int time;
+			float ulimit = 0,llimit = 0;
+			std::string buf;
+			
+			line2 = line.substr(9,length - 9);
+			std::stringstream ss(line2);
+			std::vector<std::string> slices; 
+    			while (ss >> buf)
+        			slices.push_back(buf);			
+			
+			for (unsigned i = 0; (i + 4) <= slices.size(); ++i) {
+				zStep_ms_ = atoi(slices[i++].c_str());
+				zStep_um_ = atof(slices[i++].c_str());
+				zStep_ll_um_ = atof(slices[i++].c_str());
+				zStep_ul_um_ = atof(slices[i++].c_str());
+			} 		
+			if (zStep_ms_ < 0) zStep_ms_ = 0;
+			else slices_ = (zStep_ul_um_ - zStep_ll_um_) / zStep_um_;
+			if (slices_ <= 0) slices_ = 1;
+		    }		
                } else
                if (line.find("# CHANNELS ") != std::string::npos )  {
                     if (length > 11) {
@@ -950,7 +988,7 @@ int CITC18Protocol::OnTime (MM::PropertyBase* pProp, MM::ActionType pAct)
                     if (b_AD_[x] && pDataIn_ != NULL) 
                     {
                         g_currentAD[x] = pDataIn_[((g_currentTime - 1) * MaxChannels_) + FindInSequence(GetADid(x)) + PIPELINE];
-                        fprintf(stderr,"%d AD %d = %d %d\n",g_currentTime -1,x,g_currentAD[x],((g_currentTime - 1) * MaxChannels_) + FindInSequence(GetADid(x)) + PIPELINE);
+                        fprintf(stderr,"%ld AD %d = %d %ld\n",g_currentTime -1,x,g_currentAD[x],((g_currentTime - 1) * MaxChannels_) + FindInSequence(GetADid(x)) + PIPELINE);
                     }
                     else g_currentAD[x] = 0;
                 }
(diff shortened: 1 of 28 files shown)


nicost:
@tronar: This is a wonderful start, but before merging we should make a couple of improvements.  I pulled your code into nicost/micro-manager tronar branch and started on some of it:
- Plugin window should remember its position.  The latest preferred way is through internal/utils/WindowPositions.  I'll add this.
- It should not be needed to have a snap button.  Rather the user should be able to draw ROIs on the Snap/LIve window (or any other for that matter) and add them.  Maybe even nice to copy ROIs straight from the ImageJ ROI Manager?
- It should not be needed to have a "RunAcquisition" button.  Maybe there should be a "Attach", or "Start" button, and the plugin should look for any new display (or maybe even old ones?).  I can look at that as well.
- Indeed, there has to be a way to store and display the ROIs

tronar:
I do not know if there is a common etiquette for GIT collaboration, please do tell me if I do something I should not :)

Your #1 is fine with me, I prioritized getting it to working state, inline with CI and agility :)

#2 and #3 ... buttons do have a reason: I have come to the conclussion that users do not know how to use the tools, and they should not need to. Micro-Manager is a complex tool so making it obvious for the beginner (or casual) user on the steps to get it up and running was a design objective. Text on window guides you on steps to have it running. If there is a preferred way for the expert user, so be it. Also, starting the snap from the plugin gives the plugin control of what window is active. I found no easy way to determine that, given that each window is a separate world for ImageJ.

#4, is old news to me. I found no overlay controller that is compatible with the concept of ROI set. May be that's something that should be "imported" That's the line I was thinking for when I get some more time ... A ROI manager with a metadata format and corresponding overlay controller.

nicost:
@tronar: All is good (and I do not know much about etiquette either;). Only thing is not to take anything personally, we are both trying to make this a great tool!

Remember window position was easy.  Pushed into my branch @tronar.  I'll open a PR on yours, so that we can stay in sync.

Instead of having button to add and clear ROIs (and having to deal with ROI display), I would use the Image ROI Manager (ImageJ menu > Analyze > Tools > ROI Manager).  Add ROIs to the ROI Manager, and use the Show All button to show them all.  All you need to do in the plugin is possibly open the ROI Manager (I have code to open the ROIManager with the Show All button checked), and read the ROIs from the ROI Manager when the plugin is "activated".  The ROI Manager also "transfers" to new windows, so you have build-in visualization of ROIs, and the user can even save ROIs ( More > Save), although that is a bit cumbersome for the user.

As for 2 and 3, the UI could be simper.  All that should be needed is a checkbox or button, active/inactive, a help button linking to online documentation (https://micro-manager.org/wiki/Plugins/RTIntensities), and possibly a button to (re-)open the ROIManager.  You can get the new DataViewers using the Event display.internal.event.DataViewerDidBecomeActiveEvent that is send by the display manager (i.e. register with studio_.displays().registerForEvents(this).  You'll need to decide what to do when the user switches the active DataViewer (i.e., open a new graph, wipe the old one, associate graphs with displays, or something else).

If you'd like I can have a stab at it.

tronar:
Be my guest. I had seen the ROI manager but then, given the hard to get from Micro-Manager to ImageJ world, kind of lost "interest" in ImageJ based add ons. But if you see it (the path to have it working, that is) great!
I lack real user experience data, but I don't envision one person having multiple windows open while working on something. Why would you do that ? I don't see a gain in adding complexity to change scope of graph. May be attach to a display on open (some form of choosing one display in case there are many) and vanish if that display closes ? May be be able to open more than one...

tronar:
@nicost: Ok, I revisited the ROI manager and now (with current knowledge of the matter :) it seems like it's doable to leverage it. It kind of works even with Micro-Manager multi faceted view linkage (tm) and it has the plus of solving, at least functionaly, the show and save ROIset as you pointed out.

I'll see if I can "chase" it from the plugin context, that is. One way, not the pretiest but doable, would be to launch it from the plugin.  (i.e. replacing Clear and Set buttons with "Define ROI set")
which is what I guess you did hint. (I'm re reading your comments as I type this message, and discovering most of its meaning :)

So just "Define ROI set" and "Start", plus some simple usage. Sounds good.

Q: is it common to have more than 2 channels ?

tronar:
Never mind, bringing up the ROI Manager is/was quite easy. I'll have some new code soon, I hope.

nicost:
In a way, the plugin does not need any buttons, it can just grab the ROIs from the ROI manager.  Only tricky thing is how to communicate this to the user.

A. Yes, it is quite common to have 2 or more channels.  One possibility is to ask the Datastore how many channels it has, and then plot ROI intensity for all.  However, that may not be what the user wants.  I also think that quite a few people will be interested in ratio imaging, i.e. the ratio of the intensities under the ROI in two channels.  I would not go there now, but I can imagine some kind of UI that let's users define what output they would like.  Just brainstorming here.

tronar:
True, from a paper I browsed, defining backgroud points and compensating might also be a NTTD.
(Nice THing To Do :) But UI first. Let's see what can be done. May be one button (start) that actually brings the ROI Manager up if it has not been done already... but I don't really like mutating buttons that much. Given that the ROI Manager is going to be a fundamental part, having a dedicated button and a start one seems like an interesting balance.

tronar:
@nicost Well, it seems it works. Better than easier than expected. I'm lost in Github now, but if you are able to test the code... I'll appreciate it.
Basic stuff, multiple channels to come later.

tronar:
@nicost: I'm almost done with 2 channel support, but found a problem. I expected to receive one event per image and it seems I'm only getting one event per "round" so to say. Is this always so ?

nicost:
I have not looked at it in detail yet, but I suspect that the problem arises because the "throttling code" (the code limiting the analysis to 100Hz) will throw out half the images (i.e. the second channel will appear within 10ms and not be plotted).  The code should check which channel it is before throttling.  It may be easiest to not throttle at all for now, and re-introduce throttling if it is really needed.

tronar:
I see, yes, now that real time is in there, channels might be with a small time offset. I assumed timing was "global" channelwise.
(Throtle is needed for live for sure...)

tronar:
Hmm, on second thought, no, because I'm counting all events (imageReceived) and it counts one per time, not two.

tronar:
But nevertheless, assuming channel 0 precedes the rest, and they come in sequence, this could work.

nicost:
> (Throtle is needed for live for sure...)

Why?  It runs fine at 100Hz on my laptop (512x512x2bytes, 3 ROIs)?  If performance is an issue this could go into its own Runnable, and be made into a CoalescentRunnable (i.e., only the last task submitted after the previous finishes will actually be executed).  Premature optimization can be detrimental;)

> assuming channel 0 precedes the rest, and they come in sequence, this could work.

That assumption is invalid.  There is nothing in the api that guarantees the order, and - depending on the internals of the acquisition engine - will surely change. 

Shall I have a stab at making it work without assumptions about order?

tronar:
> Why? It runs fine at 100Hz on my laptop (512x512x2bytes, 3 ROIs)? If performance is an issue this could go into its own Runnable, and be made into a CoalescentRunnable (i.e., only the last task submitted after the previous finishes will actually be executed). Premature optimization can be detrimental;)

Just prejudice... it seems pointless to drive so many points realtime. If you are going at such a frequency, I guess you will overflow any resource  that is live and holding state.

> Shall I have a stab at making it work without assumptions about order?

I don't mind doing it order independent, but given the MDA does the controlling, would love to know how a unsorted acquisition could happen.
In any case, I'm not receiving the events in any order, just receiving one per time point.

nicost:
Ran the latest code and I am confused as a user;)

Ran a 2 channel MDA with 3ROIs.  A dialog pops up "Please select: 1, 1&2, 1/2".  I assume that it asks me about channels (but certainly much better to put the word "channels" in the dialog.

Selected "1&2".   The resulting graph has 3 lines, labeled: "1/1, 1/2, 2/1, 2/2, 3/1, 3/2"  I have no idea what these mean.  Does the first number refer to the ROI number, and the second one to the channel number?  Better to make that clear.

When I select "1", I get the same legend, and very strange lines that seem to have been gathered from both channels.  I guess that is a bug?

Selecting "1/2" again gives the same legend, but only 3 lines, and the values make sense with what I would expect for the intensity in the ROI in channel 1 divided by the intensity of the same ROI in channel 2.

tronar:
UI is in dev state :) Just trying to get it running.
The selection is channel 1 alone, channel 1 & 2, channel ratio. as I guess you guessed ?
The labels are ROI/channel if you selected 1&2. (Or else they are ROIs as they were slimmed by you :)
When I tried only channel 1, I got what I was hoping (and the legend was just the number of ROI)
What are you using as input ?
The title I was using as debug aid for now, cause my side, multichannel ain't work :(

nicost:
MDA 10 time points, (5 sec apart), 2 channels, ch1. 10 ms, ch. 2 200ms, pressed "1".

image

tronar:
Nice :)

tronar:
It seems my merging was not perfect ... that should resolve the naming and plots when graphing 1 channel only. Any was you tell me how do you configure the demo MDA with multi channel ? Cause I end up with an image set which has 20 images (10 x 2) but I only receive 10 events.

nicost:
Slight improvement, legend is now for one channel, but data from second channel still picked up:

image

nicost:
MDA settings that I use (and that generate events for each image that is inserted into the Datastore):

image

tronar:
I don't see data from second channel there. The code returns early if current channel is above the number of channels plotted, I don't see how you might be getting channel 2 data in the plot.
On the other hand, here's my MDA config, a 2 channel (1&2) plot (only 1 showing up) the final capture (showing 20 data points) and the title showing only 7 events were received...

mda
2channels
image
RT

tronar:
Oh, wait, I'm not using latest version of Micro-Manager. May be some recent code fixed that, I tested in a preset VM...

tronar:
Nope, same thing with latest nightly :(

nicost:
Copy exactly my MDA settings.  You are using the LowRes/HighRes presets.  An image taken with the LowRes preset has a different size than an image taken with the HighRes preset and can not be added to the same Datastore. Your log file is full or errors.  
There should be better user feedback (in the MM code).  Error reporting during MDA is difficult because you do not want to overload the user with annoying dialog boxes and it is important to continue acquisition.  I will look into adding alerts (although not many people seem to notice those), and possibly a warning/explanation of what went wrong at the end.

Once you do this, you will get the notifications from the second channel and see what goes wrong.   The logic you are using seems quite complex, and to be making assumptions that will not be always correct.  I have not thought about it too deeply, but I would probably keep a list of measurements (keyed by frame number), and once all the required data have been received (i.e. data for only channel 1, or data for both channel 1 and 2) update the plot.

tronar:
Ok, I understand now why I was not getting events (although I don't understand the result of my previous try, where I was getting interleaved channel 1 and channel 2 events, but I digress).

I would contend that my logic is not complex, but understanding my code is (w/o help, that is). Again, this started as a "I want to get this working, can I ?" experiment. I have no problem in, as someone I pay respect would say, add enough comments to keep the length of code constant :)

If plotting non related data (only channel 1 or both channels) code should work as is. If plotting related (ratio) then this will plot with the frequency of second channel using latest value of first, so I guess it's ok too.

In any case, now that I can play (test) multichannel, I'll see if I can add the option to substract a set of background reference points, and wait for real users feedback.

tronar:
image

Strokes seem like a nice touch...

nicost:
Yes!  Still hard to grasp the legend (but I guess you want to keep it compact).  

Maybe 1-c1, 1-c2, 2-c1, etc..?

tronar:
From what I'm told, ther could be tens of ROI points... I want to be obvious from the plotting,

nicost:
It isn't;)

tronar:
Ok, 1/c1 will be, you have the experience...

tronar:
Or may be just 1c1 1c2 2c1 ?

nicost:
Sure

tronar:
That's a 1 char update :)

tronar:
FTR, tried moving from a deprecated interface of ROI manager to the "new" one, and it does not return ROIs count: manager_.getCount() returns 0 even when a ROI is defined. So using again tone that I found in an example, based on manager_.getList().

tronar:
image
Well, this does what I was hoping. Naming some ROIs with bg prefix triggers background "equalization", i.e., mean of background intensities is substracted from other points values, and plotted in dotted grey.

tronar:
The plugin is receiving 2 events per update (dataStore and dataProvider) which seems not to be creating issues, but is clearly not ideal. I did not find any "id" to filter duplicates, but again I'm on the preventing side of things.
Made sure we do not left over plots dangling.

nicost:
Looks good!

Yes, looks like the onOverWrittenImage Even handler is not needed.  You should delete that one (I already forgot how and what, but this looks much cleaner in any case).

I see only two possible improvements:
- It would be nice to be able to handle MDAs with more than 2 channels.  Could still only plot 2 channels, but give the option to select those from whatever is found in the MDA data
- It would be nice to have an option to auto-start (i.e., no need to press the "plot" button).

Both can be added later, but always good to do things while they are fresh.  

Let me know when you want the code to be merged (I'll cleanup a bit more at that point).

tronar:
If you give me a hint on what you would consider for starting a new plot...
We need ROIs, so we already have a display. From the plugin side, I don't know if that same display is going to generate points (snaps) or there will be a new active windows the user wants to plot.
A clean way would be to mark from the source side that you want to plot, but that's too much change I guess. May be a new setting to "follow new windows" ? And presets for plot mode.

Multichannel would not be hard. May be make the ratio mode a preset, then just start as many channels as are there.

tronar:
Well, it seems I can not let go.
Created a new  auto start "option", and event handlers for live on, and new MDA.
Got some trouble there because it seems the event is kind of early (that may a wrong conclusion) so I deferred starting the plot to when the first image arrives. Seems to hold.

There is an issue, at least with DemoCamera, when using the ROI Manager, and is not related to this plugin but obviously hurts. It happens a lot that starting a second acquisition triggers an error. May be related to multichannel. 
Reproduce by :
-open snap
-open ROI manager
-setup a ROI
-run an MDA acquisition
-close / discard
-run another MDA acquisition
image
java.lang.IndexOutOfBoundsException: Index: 0, Size: 0 in Thread[AWT-EventQueue-0,6,main]
                                   [       ]   at java.util.ArrayList.rangeCheck(ArrayList.java:657)
                                   [       ]   at java.util.ArrayList.get(ArrayList.java:433)
                                   [       ]   at org.micromanager.acquisition.internal.MMAcquisition.<init>(MMAcquisition.java:218)
                                   [       ]   at org.micromanager.acquisition.internal.AcquisitionWrapperEngine.runAcquisition(AcquisitionWrapperEngine.java:146)
                                   [       ]   at org.micromanager.acquisition.internal.AcquisitionWrapperEngine.acquire(AcquisitionWrapperEngine.java:77)
                                   [       ]   at org.micromanager.internal.dialogs.AcqControlDlg.runAcquisition(AcqControlDlg.java:1359)
                                   [       ]   at org.micromanager.internal.dialogs.AcqControlDlg.lambda$createRunButtons$22(AcqControlDlg.java:627)
...
That one I leave for someone else (Nico ?)

nicost:
Made that into issue #817.  I'll look at the current plugin code soon.  Happy to see I planted a seed;)

nicost:
Merged in PR #839.  I added a few last fixes, and included your name in the copyright info.  You may need to delete your fork of the micro-manager repo and check out a new copy to get the full latest version (and if you want to make further changes, you may want to create a branch at that point).

tronar:
Thanks, Facendo disco :) (You learn by doing...)

PR #795, opened 2020-05-07
Add ApplicationSkinEvent that is posted when the skin mode is changed

mabruce:
ApplicationSkin#setSkin stated that the event would be posted, but the
event didn't actually exist.


nicost:
Looks fine to me, since it involves the api, I'll wait for @marktsuchida to decide

marktsuchida:
@mabruce Sorry for letting this sit for so long.

The only change that I would ask for is to make the event an interface rather than a class, with a corresponding implementation `DefaultApplicationSkinEvent` in `org.micromanager.events.intenal`. I know some other events are also classes, but those are bugs -- we want _nearly_ all API types to be interfaces because classes often restrict future evolution.

marktsuchida:
Thanks, Marc!

PR #796, opened 2020-05-11
Allow multiple PositionListDlg

nanthony21:
The current implementation of `PositionListDlg`  operates under the assumption that there will be only one instance of it. It fires `NewPositionListEvent`s to the MMStudio `EventManager` and it updates the `AcquisitionControlDlg` whenever a change is made to the `PositionList`.

However there are scenarios where someone may want to make use of one or multiple `PositionListDlg`s to generate or edit a `PositionList` without wanting to have any effect on the `EventManager` or the `AcquisitionControlDlg`

This PR removes the functionality that relies on only having one instance from the `PositionListDlg` and moves it to a new subclass called `MMPositionListDlg`

The end result is that the behavior of the base Micro-Manager program are completely unchanged while readability is improved. However developers wanting to extend Micro-Manager now have an easy-to-use UI for `PositionList`s that won't unexpectedly affect other aspects of the program.

It is worth noting that numerous other minor changes have been made just for the sake of code readability.

### DIFF
--- a/mmstudio/src/main/java/org/micromanager/internal/MMStudio.java
+++ b/mmstudio/src/main/java/org/micromanager/internal/MMStudio.java
@@ -100,7 +100,7 @@ import org.micromanager.internal.menus.MMMenuBar;
 import org.micromanager.internal.navigation.UiMovesStageManager;
 import org.micromanager.internal.pipelineinterface.PipelineFrame;
 import org.micromanager.internal.pluginmanagement.DefaultPluginManager;
-import org.micromanager.internal.positionlist.PositionListDlg;
+import org.micromanager.internal.positionlist.MMPositionListDlg;
 import org.micromanager.internal.propertymap.DefaultPropertyMap;
 import org.micromanager.internal.script.ScriptPanel;
 import org.micromanager.internal.utils.DaytimeNighttime;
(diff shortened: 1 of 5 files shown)


nicost:
Looks really good! 

Few things I would consider:
- Access modifiers of a number of variables in PositionListDlg are now "Default", which make them "private to all code in the package, whereas it seems they should be protected.
- Can the use of "this" in the PositionListDlg constructor be avoided (which was the reason to have the addListener function)?  Other than putting all these occurrences in a function and calling after construction, this could be done by using a static function to construct the object, but I would not knwo how to use that with inheritance.

nanthony21:
Thanks for taking a look, Nico.

I've set all the member variables of `PositionListDlg` to `protected` except for the few which generated an "Exporting non-public type through public API" warning, which have been set to `private`.

I have also reversed the commit which added new references to `this` to the constructor. However, it is worth noting that there are still two references to `this` that were already there previously. The issue of "leaking `this` in the constructor" is something I haven't thought about much, but I'm not sure I understand why it is problematic in this case since `this` is only being passed to private variables anyway.

nicost:
Great! Good to set access privileges as strict as possible.

The main issue with "leaking this in constructor" is that the object can become available to other code/threads before it is fully constructed, so it is worth avoiding whenever possible.  I started to follow @marktsuchida 's example of using a static function that creates the object and adds "this" after object construction.  I'll have a quick look to see if this can be done easily in this case.

PR #800, opened 2020-05-12
Pycro-manager Java side

henrypinkard:
Updates now that pycro-manager Java side + dependencies are now in external Jars. This is mostly updates in Magellan + a small bit of code to launch ZMQ server


henrypinkard:
@nicost in trying to test this locally, it seems that the pycro-manager-Java Jar on bintray may be out of date. Specifically, it appears to have the internal version number of 2.3, when the latest source code is 2.4 (https://github.com/micro-manager/pycro-manager/blob/50dd1ac218a66e1708f35723fead8eb6eca4902f/java/src/main/java/org/micromanager/internal/zmq/ZMQServer.java#L33)

You have a build script that uploads this to bintray manually, right?

henrypinkard:
Actually, I think this is not a problem with the Jar on Bintray, but rather the old code still existing in MMJ_.jar that I am testing on (just switched over to IntelliJ, still getting the hang of things)

henrypinkard:
Not yet ready to merge, still working out some changes

henrypinkard:
Should be ready now

nicost:
@henrypinkard Thanks!  

It looks like MMStudio.java has a lot of changes that are whitespace only.  Would be nice to separate those from real changes, as they make it hard to see what actually changed. 

White space changes are inconsistent.  Sometimes string concatenation + goes from the end of the line to the beginning, other times the other way around. Javadoc new line removal sometimes makes things harder to read, especially when the dot at the end of a sentence is removed.


Noticed that you are using the IJ classloader.  I forgot already how all that worked (except that it was very painful).  Did you check that this works outside of an IDE (i.e. when you run MM normally)?

Nice how you made ZMQServer agnostic of the class it serves!

nicost:
Awesome!

henrypinkard:
Not sure why those changes were there, but I reverted MMStudio and re added changes on top.

The IJ classloader was the solution you added before that made it able to find everything on the classpath. I wouldn't be surprised if something else crops up related to this in the future, but for now it works on both IDE and and installed version. There were also some hacks that I had to put into ZMQ server to make sure it found certain classes (specifically in Netbeans). Hopefully can now remove these in the future

nicost:
Exciting!  Should all be in tonight's build.  I'll download the binaries, and start playing with it.  Once we are sure this all works as intended, may be worth some announcements.

henrypinkard:
Agreed! Let me know how it goes

PR #808, opened 2020-05-19
updated magellan and library versions

henrypinkard:


nicost:
@henrypinkard: Can you precede your commit messages with the area it touches, for instance: 

Build: updated  AcqEngJ version
Magellan: refactored use of AcqEngJ datasink

That will make it so much easier to read through commit messages in the future.  Also, you can add multiple lines (for instance to write the new version of the library that you are using now)

henrypinkard:
Will do!

nicost:
running `ant - f buildscripts/fetchdeps.xml` now yields:

org.micro-manager.pycro-manager#PycroManagerJava;0.6.2: not found

Was Pycromanager 0.6.2 not uploaded yet?

henrypinkard:
Guess I forgot. It should be up now

nicost:
@henrypinkard also wondering if it is better to use Snapshot builds.  Those are always checked to see if there is a newer version around, and do not need adding new versions continuously

henrypinkard:
Maybe so if its not too hard to set up, though I _think_ I'm done for now

Issue #786, opened 2020-04-17
Replace use of MMFrame with WindowPositioning


nicost:
At first check how well the new class works, then add it to the api (i.e., make it not internal).

Issue #758, opened 2020-04-06
Can (parts) of internal.utils become part of the API?


nicost:
While annotating plugins with warnings against using internal packages (issue #372), it became clear that the use of internal packages is actually advisable.  Most of the plugins use packages from internal.utils for good reason.  Not using them would result in much less uniform UI behavior (undesirable for users), and/or massive code duplication.  @marktsuchida  how do we proceed?  

The most widely used internal.utils classes are:
MMFrame
MMDialog
FileDialogs
GUIUtils
ProgressBar
NumberUtils
TextUtils
JavaUtils
MathFunctions
AffineUtils

marktsuchida:
I think many parts of these can indeed be turned into APIs. Some of the grouping may need revision, some methods are inappropriate, etc., and it's mostly about sorting out details.

For example, `NumberUtils` is confusingly named. It's about Core device property values, nothing to do with numbers in general. The `AffineUtils` methods for conversion to/from `mmcorej.DoubleVector` better belong together with these than they do with the mathematical methods on affine transforms (which don't depend on `mmcorej`).

`TextUtils` is a heterogeneous mix of two obsolete file utilities (see `Files.readAllBytes` et al.), one string truncation method which is indeed useful but having more to do with GUI than text, and some number formatting stuff that really should be defined at the point of usage (or merged into `NumberUtils.doubleToDisplayString()`?). They clearly don't belong together.

`JavaUtils` looks like a dumping ground for methods that people didn't know where to put, and some of them have implementations that are highly suspect.

And so on. In short, some of the functionality definitely should be in the API but they are not currently written in the form of a well organized API. I think some amount of "miscellaneous utilities" classes are okay for internal-only use, but in the API at least each class should have clearly delineated responsibilities.

`MMFrame` and `MMDialog` are a slightly different story and can hopefully be retired in favor of `WindowPositioning`, but the latter might need some more testing and has some TODO comments. `MMFrame` and `MMDialog` are broken when you already need a frame or dialog to inherit from something else (or when you are given an existing `JFrame` instance created by someone else), and also encourages you to subclass the frame, which often leads to bad code organization except in the simplest cases.

I think the first steps are to narrow down the methods (or functionality) that we want in the API and to decide where they should go, given the overall organization of our API.

[...]

Issue #391, opened 2016-08-03
mm2: The functionality of MMFrame should be in the API


marktsuchida:
Currently it does this:
- Positioning the window according to its last saved position
- Offsetting said position when there are multiple windows
- Setting up the menu bar (OS X only)
- Saving window position when the window is closed

Plugins should be able to take advantage of this functionality. It should be possible to provide an API that doesn't require subclassing a particular class.

nicost:
@marktsuchida can you evaluate if we still want this (I vote yes), and how to go about it if we decide to add it to the API? I'll be happy to help implement and test.

marktsuchida:
I remember working on this and the code might still be in master. I had it so you don't have to inherit from `MMFrame`, which is too limiting in the absence of multiple inheritance.

I'll try and see if I can resurrect it.

marktsuchida:
See comment in #758 about `WindowPositioning`.

PR #815, opened 2020-05-25
Proof of concept: Dockable windows

nanthony21:
A typical session of Micro-Manager usually ends up with so many different windows opened that they can't all fit on the display without overlapping and it can become quite difficult to keep track of them, especially when other application windows are being switched to/from. This PR proposes solving this problem in the same way that most other complex programs solve it, by introducing a single main frame in which other frames are docked.

This PR was thrown together fairly quickly and introduces many bugs. Implementing this idea properly would be likely take a lot of work but I just wanted to put it up out there to get feedback on if this is a direction that Micro-Manager devs would be interested in moving towards.

This PR doesn't currently default to a very good initial layout but the idea is that a layout might look something like this:
Untitled

Users would be able to easily save and load custom layouts. By default there would be a central location for image display and, any new image displays that were opened would be tabbed on top of the existing display.

### DIFF
--- a/buildscripts/ivy.xml
+++ b/buildscripts/ivy.xml
@@ -39,6 +39,7 @@
 		<dependency org="org.jfree" name="jcommon" rev="1.0.24"/>
 		<dependency org="org.jfree" name="jfreechart" rev="1.5.0"/>
 		<dependency org="org.scijava" name="scijava-common" rev="2.77.0"/>
+		<dependency org="org.dockingframes" name="docking-frames-common" rev="1.1.1"/>
 		<!-- Magellan dependencies -->
 		<!-- https://mvnrepository.com/artifact/org.zeromq/jeromq -->
 		<dependency org="org.zeromq" name="jeromq" rev="0.5.1"/>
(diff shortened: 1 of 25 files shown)


nicost:
Very much agree that the many windows can be difficult to work with (especially on Windows), and I admire your attempt!  I do like something like the IntelliJ interface that quickly became intuitive for me.  

I have never worked with a UI build on top of docking-frames.  How do you like its use in practice? 

Is it still possible to work with fewer frames open (i.e. only the main frame and an image frame)?

How would plugins behave?

nanthony21:
From a programming perspective, getting things set up in "docking-frames" was very straight forward. However, from a user perspective I think that it looks somewhat clunky, especially if you pop one of the frames out of the main `JFrame`. Fortunately there are many different options for this type of functionality so I'll take a look at a few demo programs. https://stackoverflow.com/questions/304874/what-are-good-docking-frameworks-for-java-swing

If this is an idea that you are generally open to then maybe a good starting point would be for me to submit a PR that will default to grouping display windows and an "Inspector" into a single frame. This alone will do a lot to reduce clutter and then other frames can potentially be incorporated later.

Plugins are probably best being left alone. Similar to the existing `MMFrame` we could make an `MMDockableFrame` class that plugins can use if they want to.

nicost:
Would this still leave the possibility of having displays side-by-side?  That is important for people who want to compare channels and things like that (there is quite a bit of code involved in synchronizing displays).  

Mark and I are both interested, but major UI changes are tricky to do well, and it will be important for maintenance burden to decrease rather than increase.

Also, we are trying to remove MMFrame altogether (see issues #786, #758, #391) .  There may be other easy ways to achieve dockable windows.

nanthony21:
@nicost I'll make sure that any solution I propose gives the user the ability to "float" a display window as it's own independent frame as well as the ability to drag and drop multiple displays next to each for a flexible "split view".

I'll close this for now as it will probably be a while before I get around to this.

PR #822, opened 2020-06-02
AcqEngJ: to 0.5.1; Pycro-manager: 0.7.3; NDTiffStorage: 0.2.6

henrypinkard:


henrypinkard:
- AcqEngJ changes fixes bug that gave exception if no core-focus device
 - NDTiffStorage fixes strange bug that I don't fully understand where `ByteBuffer.rewind()` threw an exception only when using Jar but not when running from source in IntelliJ. Apparently this has [something to do with a version mismatch between JDK8 and JDK9](https://issues.apache.org/jira/browse/MRESOLVER-85), though the jar seemed to be compiled with JDK8 and ImageJ also seemed to running Java8
 - Pycro-manager updated to latest version of both dependencies

PR #823, opened 2020-06-04
add iterator interface for more MMCoreJ "vectors"

nanthony21:
The current version of MMCoreJ extends the autogenerated code for `StrVector`, `BoolVector`, and `UnsignedVector` so that they implement the `Iterable` interface.  This makes it much simpler to work with these classes in Java.

This PR simply applies the same treatment to `LongVector`, `DoubleVector`, and `CharVector`.

If this looks acceptable I have one question before merging. The autogenerated `get()` method of `LongVector` returns `int` rather than `long`. This means that we have to cast from `int` to `long` in order to act as an `Iterable<Long>`, is this ok or should `LongVector` actually implement `Iterable<Integer>`?

### DIFF
--- a/MMCoreJ_wrap/MMCoreJ.i
+++ b/MMCoreJ_wrap/MMCoreJ.i
@@ -911,6 +911,144 @@
 // instantiate STL mappings
 
 namespace std {
+	%typemap(javaimports) vector<char> %{
+		import java.lang.Iterable;
+		import java.util.Iterator;
+		import java.util.NoSuchElementException;
+		import java.lang.UnsupportedOperationException;
+	%}
+
+   %typemap(javainterfaces) vector<char> %{ Iterable<Character>%}
+
+   %typemap(javacode) vector<char> %{
+      public Iterator<Character> iterator() {
+         return new Iterator<Character>() {
+
+            private int i_=0;
+
+            public boolean hasNext() {
+               return (i_<size());
+            }
+
+            public Character next() throws NoSuchElementException {
+               if (hasNext()) {
+                  ++i_;
+                  return get(i_-1);
+               } else {
+                  throw new NoSuchElementException();
+               }
+            }
+
+            public void remove() throws UnsupportedOperationException {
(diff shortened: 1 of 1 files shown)


marktsuchida:
@nanthony21 Sorry for the late reply, and thanks for doing this -- it's been on my wish list for a long time!

Regarding your question about `LongVector`, note that `long` is 32-bit (i.e. same as `int`) in x86 C++ and Windows x86-64 C++, whereas it is 64-bit in Java and x86-64 Unix C++. I have no idea why the decision was made to use `long` instead of `int` for certain MMCore interfaces -- perhaps it was conventional for some people. To be honest, there are probably a bunch of platform-specific bugs in our code due to this mismatch, although they will rarely be encountered, because most values never reach the numerical limits.

In other words, it is probably not safe to exceed 32-bit limits even when using `LongVector` on 64-bit Unix -- and definitely problematic on Windows. These interfaces were not designed to expect values outside of the 32-bit range.

So I don't think there is any 100% ideal answer here, but I do think it might be slightly cleaner for `LongVector` to implement `Iterator<Integer>`. If mapping to `Iterable<Long>`, it would be best to throw an (unchecked) exception when someone attempts to store a value that doesn't fit in 32 bits. Casting from `int` to `long` is not a problem.

One thing we should do is discourage future use of `LongVector` in the MMCore interface (fortunately it looks like it is barely used).

Regarding the overall change itself, I will need to get back to you after checking how the Java interface actually changes, as I can't tell this from just looking at the Swig code -- although I don't expect any problems.

nanthony21:
Thanks for the feedback @marktsuchida. I've changed `vector<long>` so that it maps to `Iterable<Integer>` and added a brief comment explaining why.

The only place I see `vector<long>` actually being used is for  `MMCore::getAvailableDeviceTypes`, so probably not likely to cause issues either way.

marktsuchida:
@nanthony21 Many apologies for taking so long to merge this, and thank you so much for contributing!

PR #853, opened 2020-06-24
1.Compatible with the camera

fandayu:
1.Compatible with the camera(Dhyana 401D, Dhyana 95V2)


nicost:
Merged into the 1.4 branch.  In the future, can you work with the master branch instead of mm2?  

Your code changes are pushed into the subversion repository, which is used to build MM1.4.  From there, it goes automatically into the svn-mirror branch and then is cherry-pick into the master branch.  The mm2 branch is no longer used and will be removed soon.

Thanks!

PR #870, opened 2020-07-06
When MMstudio is started with a profile name startup argument, make sure to apply the correct skin for the profile.

nanthony21:
This PR applies a treatment similar to 9ee975aaeadbccd2924152faf8f6d90617bed880 for when MMStudio is started using a `profileNameAutoStart` startup argument.

Previously user profiles that wanted to use the "Day" skin would cause the UI to load improperly when the startup argument was used, now it works as expected.


nicost:
Thanks!

PR #874, opened 2020-07-09
Sequentially number displays rather than using the hashcode

nanthony21:
This PR implements something that was suggested in a `TODO` comment. Rather that giving each DisplayController a unique ID with a lengthy hash code, we now use a counter to make sure that each object has a unique number during runtime. The end result is that display window titles and other places that refer to the display controller will be more human readable.

For example when you snap an image you will see "Preview-1 (100%)" rather than "Preview-86839238756021256 (100%)"


nicost:
I really like this, as the current window naming is super difficult to work with as a used.  @marktsuchida  OK to merge?

marktsuchida:
Thanks, @nanthony21!

Issue #834, opened 2020-06-10
set preferred save mode in SequenceSettings


impact27:
When running an acquisition with the acquisition manager, there is an option to use `org.micromanager.acquisition.SequenceSettings` with `runAcquisitionWithSettings`. However, there is no settings to say if the file should be saved as a multipage tiff ot as multiple files. This is one of the only settings from the acquisition pane that is not accessible. To change it, one must access some internal class:
```java
import org.micromanager.data.Datastore;
import org.micromanager.data.internal.DefaultDatastore;
DefaultDatastore.setPreferredSaveMode(mmStudio_,
    Datastore.SaveMode.SINGLEPLANE_TIFF_SERIES);
```
It would be nice to have that as an acquisition setting.

nicost:
SequenceSettings now lets you specify the desired save mode.  Added in PR #882. Please re-open if it does not work, or needs further additions.

Issue #802, opened 2020-05-14
AcquisitionWrapperEngine saves invalid `SequenceSettings`.


nanthony21:
Possibly related to #793

`AcquisitionWrapperEngine` does not load `SequenceSettings` in the same way that is saves them. This appears to be largely due to the fact that `SequenceSettings` do not have boolean fields to mirror the boolean variables of the `AcquisitionWrapperEngine` such as `useFrames_`, `useSlices_`, etc. 

For example, if `useFrames_` is `false` then `AcquisitionWrapperEngine:getSequenceSettings` will set the `numFrames` field of the `SequenceSettings` to 0. However, if those same settings are loaded by `AcquisitionWrapperEngine:setSequenceSettings`, it will set `useFrames_` to `true` and will set `numFrames` to 0, which is an invalid setting which will result in an error when the acquisition is run.

nanthony21:
The solution I would propose would be to change `SequenceSettings` so that they fully match the parameters of the `AcquisitionWrapperEngine`. I'm not sure what other effects this might have though.

nicost:
This is implemented in PR #882, and seems to work as intended now.  There may be bugs/edge cases, so feel free to re-open if there are still issues.

Issue #793, opened 2020-05-07
MDA dialog does not properly update


nanthony21:
This is true when setting acquisition settings through Studio.acquisitions().setAcquisitionSettings() as well as when settings are loaded from file in the GUI. The MDA dialog does not get property updated to reflect the actual settings even though the proper settings do appear to be loaded to the acquisition engine. 

I may have missed a few things but the following components do not update to represent the loaded settings: the "Channels" checkbox, the "Save Images" checkbox, and the "Name Prefix" textfield.

This appears to be a problem with AcqControlDlg.updateGuiContents()

nicost:
Should work correctly after PR #882.  Please re-open if this is not true.

PR #882, opened 2020-07-13
Refctoring of Sequence Settings, AcqControlDlg, and AcquisitionWrapperEngine

### LINKED ISSUE #872: Loading acquisition settings does not restore root/prefix.
Also see #793 #802 #834 

I have set up a script to acquire three sequential sets of images. The simplest way I found is to sequentially load and run three Acquisition Setting Files. The script (below) works fine with the exception that all the files are save with the same prefix, despite each acquisition settings file having the following entries though they differ in their prefix.

    “save”: true,
    “root”: “D:\Temp\0630_Testing\”,
    “prefix”: “Pre”,

Is there anything else that needs to be done to activate correct file naming? Perhaps this is relevant, manually loading the acquisition settings files also does not update the file path/name. Perhaps this is a bug that has since been fixed, as my version is not particularly new.
Thoughts?

Acquisition Script:

// file locations
String rootDirName = “D:/Temp/MMTesting/”;
String acqPre = “AcqSettingsPre.txt”;
String acqActivate = “AcqSettingsActivate.txt”;
String acqPost = “AcqSettingsPost.txt”;

//Assign mmc to the Core
mmc = mm.getCore();

//parameters
long waitMs = 1000;

mm.acquisitions().loadAcquisition(rootDirName+acqPre);
mm.acquisitions().runAcquisition();
mmc.sleep(waitMs);
mm.acquisitions().loadAcquisition(rootDirName+acqActivate);
mm.acquisitions().runAcquisition();
mmc.sleep(waitMs);
mm.acquisitions().loadAcquisition(rootDirName+acqPost);
mm.acquisitions().runAcquisition();

nicost:
SequenceSettings now contain everything(?) needed in AcqControl Dlg to restore display.  Converted to using Builder pattern  Separates all data used into SequenceSettings.  WrapperEngine code is simplified and more manageable (or so I think). Closes #872 #793 #802 #834


marktsuchida:
Awesome, this is so much cleaner, and stays compatible to the extent possible!

PR #883, opened 2020-07-14
HardwareFocusExtender: Add comments, fix potential future bug, add debug logging.

nanthony21:
This PR contains minor changes to the HardwareFocusExtender code that are unlikely to effect usage.

1. Comments are added to further explain variables, return values, etc.
2. Debug level logging is added so that performance of the extender can potentially be analyzed in retrospect.
3. An potential error is avoided that would occur if a hardware focus device other that the Nikon TI PFS were to have a property called `Status`.

### DIFF
--- a/autofocus/src/main/java/org/micromanager/autofocus/HardwareFocusExtender.java
+++ b/autofocus/src/main/java/org/micromanager/autofocus/HardwareFocusExtender.java
@@ -49,6 +49,8 @@ import org.scijava.plugin.SciJavaPlugin;
 public class HardwareFocusExtender extends AutofocusBase implements AutofocusPlugin, SciJavaPlugin {
 
    private static final String AF_DEVICE_NAME = "HardwareFocusExtender";
+   
+   //Settings property names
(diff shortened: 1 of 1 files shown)


nicost:
Thanks Nick!  I started the HardwareFocusExtender years ago, but never finished it (and remember it working incorrectly quite often).  Guess you ran into the same problem.  Would be super nice to get to a usable solution.  I am afraid that I can not be of much help due to lack of access to a microscope.  Also sounds like your tickets for the Ti2 are related.

nanthony21:
Hi Nico,
For the most part the HardwareFocusExtender has worked very well for me, but the lack of logging means it was hard to know what happened in the event that it failed. Thanks for making it in the first place.

The Nikon PFS is unfortunately not reliable enough on its own to support some of the longer imaging sessions so I have been trying to see how I can supplement it with software.

nicost:
Very curious where you end up with that.  Would be great if you can document your experiences somewhere.

PR #884, opened 2020-07-14
WIP: DisplayUIController API cleanup

nanthony21:
The management of display windows is distributed accross multiple classes including:
 - `DisplayController` 
 - `DisplayUIController`
 - `ImageJBridge`

All of which are > 1000 lines long and contain circular references which means that making any changes to this portion of the code without causing issues is extremely challenging.

This PR attempts to make this code more understandable by renaming the `DisplayUIController` class to `DefaultDisplayUIController` and by splitting the public methods of the `DefaultDisplayUIController` class into two new interfaces, `DisplayUIController`, and `ImageJBridgeParent`. this allows many circular references to be removed.

 - `DisplayController` is now only dependent on the `DisplayUIController` interface and doesn't directly refer to `DefaultDisplayUIController`.
 - `ImageJBridge` is now only dependent on the `ImageJBridgeParent` interface.

TODO:
 - There is some overlap in the methods defined in the two interfaces. e.g `getFrame`, `setVisible`, and `toFront`. I'm not sure how to best deal with this.
 - The comments and `MustCallOnEDT` annotations in the new interfaces match what was present in the original class. However many of the methods are uncommented and many methods that probably should be annotated are not.
 - Many comments in `DisplayController` , `DisplayUIController`, `ImageJBridge` appear to be outdated and no longer make sense.

I'm hoping to get some feedback and/or help on cleaning up this section of code and making it more readable.


marktsuchida:
@nanthony21 I'm not against refactoring this code (there is certainly a lot of room for improvement!), but (as you pointed out yourself) I don't feel that I can review it and be confident about the changes unless some explicit rules are established and followed about how the code is being transformed (and exceptions to such rules documented with enough detail). There are a lot of subtleties that have to do with ImageJ1 internals here, that are not obvious from how the code looks, so (sadly) I don't see how nontrivial changes can be made without a deep understanding of those interactions (something I would need to spend a few days re-learning if I were to do major refactoring on this code).

I also would rather not have two new interfaces here, because we are 100% sure that no other sane class will ever implement them (`ImageJBridge` is a helper class for `DisplayUIController`, which in turn belongs solely to `DisplayController`; neither of them are anywhere near being reusable components). I think it would be much more useful (for code maintainability) to split the complicated classes into smaller, well factored component classes that are easier to understand in relative isolation; some of them may also turn out to be reusable in other contexts.

I do sympathize with the idea of clarifying the interactions between the classes. Perhaps an `ImageJBridgeListener` (or nested as `ImageJBridge.Listener`) interface, containing only the notification methods, *could* have made some sense if `ImageJBridge` itself were better organized (which probably means being a collection of classes with better names). Under the current situation, though, fewer files, names, and lines of code seems like a greater advantage to me.

nanthony21:
@marktsuchida thanks for the feedback. One thing I don't think I understood when I did this is that compatibility with the ImageJBridge is vital part of compatibility with the Display Controller.

I'll read through this code a bit more and give this another shot sometime.

marktsuchida:
@nanthony21 No problem!

I'm not sure what you mean by compatibility (these are `internal` classes, after all, that are not intended to be used by any other code), but perhaps the class names are not helping here. At first (early in 2.0 development) there was `DisplayWindow`, which was the _only_ viewer. Later `DisplayWindow` was made a particular implementation of `DataViewer`, so that we could have alternative viewers (although this is only a very crude way of enabling that, meaning that writing a new viewer requires reinventing tons of stuff). Slightly later, I split out parts of `DisplayWindow` (or its implementation `DefaultDisplayWindow`, which was getting way too complex) into separate classes like `DispalyController`, `DisplayUIController`, `ImageJBridge` etc.; the latter in particular was intended to provide a single gateway into ImageJ1 viewer code, because our previous code was becoming (even more) unmaintainable due to direct calls into ImageJ1 scattered around various parts of the viewer system.

So `DisplayWindow` is the name of a particular data viewer -- a 2D one that is based on integration with the ImageJ1 GUI -- and the `DisplayWindow` API is a facade that hides a great deal of internal code behind itself. `DisplayController` et al. are implementation details of `DisplayWindow` (but no other viewer), and since it is unlikely that we will ever want to have another viewer based in IJ1 technology, they were not designed as reusable components. That point might be slightly debatable, but in any case I decided not to attempt the much more challenging task of further decoupling `ImageJBridge` from our viewer, making it reusable on its own.

Of course cleaning up internal-only code is quite worthwhile, but I think in this specific case that would require finding better ways of factoring the code into (hopefully) smaller classes so that the interactions between them can be more readily understood or described.

On a somewhat related topic (in that it is about `DisplayWindow` internals), it would be ideal if various generalizable aspects of viewers (histogram scaling and other interactions with the inspector, access to the underlying data via things like the gear menu items, animation during acquisition and replay, linking parameters with other viewers, etc.) were designed as modules with APIs rather than being internal and specific to `DisplayWindow`, so that people can write new viewers without reinventing (or forgoing) all those features. I expect designing such APIs will be quite challenging, but could be quite useful in the long term.

nanthony21:
@marktsuchida 

My original intention was to work towards creating a new implementation for `DisplayUIController` so that the UI would be different while still working with the histogram and other related functionality.

It sounds like `DisplayUIController` is not the correct place to do this. I'm going to close this for now. At some point I will plan to look at the `DataViewer` API and see where it can be expanded on.

PR #933, opened 2020-09-05
Studio: Handle more strange forms of binning property

nicost:
coming from camera device adapters.

The "Binning" property was meant to provide numbers, i.e. 1, 2, 4, etc.. Hamamatsu provided 1x1, 2x2, and we ended up parsing that string in the Java layer.  While testing the Spinnaker device adapter, I discoveerd that the Binning property provided the value "Mode0", which of course leads to Exceptions.  This PR works around this by testing for sane values, and if they are insane to set binning to 1.  Of course, it is desirable to fix this at the root (i.e. in the device adapter), so maybe this change is not desired.  Most importantly, there should be some kind of mechanism to make it impossible to write a device adapter that delivers anything but a number.  Maybe there should be a SetBinning() and GetBinning function in the api, and we should no longer use a property for binning?

### DIFF
--- a/mmstudio/src/main/java/org/micromanager/data/internal/PropertyKey.java
+++ b/mmstudio/src/main/java/org/micromanager/data/internal/PropertyKey.java
@@ -168,8 +168,12 @@ public enum PropertyKey {
             dest.putInteger(key(), jp.getAsInt());
          }
          if (jp.isString()) { // "1x1", "2x2", ...
-            dest.putInteger(key(),
-                  Integer.parseInt(jp.getAsString().split("x", 2)[0]));
+            String token = jp.getAsString().split("x", 2)[0];
+            if (token.length() < 3) {
+               dest.putInteger(key(), Integer.parseInt(token));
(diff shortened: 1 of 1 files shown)


marktsuchida:
@nicost I agree with the latter half of your comment, but I don't have a problem with your code that effectively ignores unexpected strings (much better than misinterpreting them!).

Introducing the use of `GetBinning` and `SetBinning` might require moving similar logic into MMCore, at least as a fallback, unless we update every device adapter. It might still be good if future device adapters could be written without the `Binning` property.

marktsuchida (approved):

nicost:
Yes, those functions should be present in the lower as well as upper APIs of MMCore.  I don't think that it would be a huge amount of work to add them to all device adapters that we have in our repository.  The hardest part will be backwards compatibility.  Ideally, we would retire the Binning Property everywhere, but that would be bad for existing configuration files.  Maintaining both will be a bit of a pain.  If we do not have a ticket o this issue yet, we should add it.

Issue #915, opened 2020-08-26
MDA channel settings not handling null correctly


marktsuchida:
I'm seeing a `NullPointerException` during startup (after loading config) caused by the `channels` field in `SequenceSettings` being set to a list containing one or more `null`s. This happens even if no config is loaded.

```
2020-08-26T12:22:08.974147 tid42592 [IFO,App] Thread 18 (AWT-EventQueue-0) terminated with uncaught exception
2020-08-26T12:22:08.974147 tid42592 [IFO,App] java.lang.NullPointerException
2020-08-26T12:22:08.974147 tid42592 [IFO,App]   at org.micromanager.acquisition.internal.AcquisitionWrapperEngine.getNumChannels(AcquisitionWrapperEngine.java:209)
2020-08-26T12:22:08.974147 tid42592 [IFO,App]   at org.micromanager.acquisition.internal.AcquisitionWrapperEngine.getVerboseSummary(AcquisitionWrapperEngine.java:563)
2020-08-26T12:22:08.974147 tid42592 [IFO,App]   at org.micromanager.internal.dialogs.AcqControlDlg.tableChanged(AcqControlDlg.java:1624)
2020-08-26T12:22:08.974147 tid42592 [IFO,App]   at javax.swing.table.AbstractTableModel.fireTableChanged(AbstractTableModel.java:296)
2020-08-26T12:22:08.974147 tid42592 [IFO,App]   at javax.swing.table.AbstractTableModel.fireTableStructureChanged(AbstractTableModel.java:216)
2020-08-26T12:22:08.974147 tid42592 [IFO,App]   at org.micromanager.internal.dialogs.AcqControlDlg.lambda$updateGUIFromSequenceSettings$32(AcqControlDlg.java:1121)
2020-08-26T12:22:08.974147 tid42592 [IFO,App]   at java.awt.event.InvocationEvent.dispatch$$$capture(InvocationEvent.java:311)
2020-08-26T12:22:08.974147 tid42592 [IFO,App]   at java.awt.event.InvocationEvent.dispatch(InvocationEvent.java)
[...]
```

The direct cause, in my case, seems to be that such a `null`-containing list is getting restored from the user profile, having been saved previously (possibly related to my investigation of #891).

We need to document the legal states of the `channels` field (i.e. whether or not `null`s are allowed when there are unchecked channels in the MDA dialog), and make sure all code follows it.

[...]

PR #935, opened 2020-09-09
Ensure that SequenceSettings.channel can never have entries that are null

nicost:
Also avoid the use of private members of SequenceSettings in the code.  It would now be possible (at the risk of breaking code that is not in our repo) to make all Deprecated members private.

I believe this takes care of issue #915

### DIFF
--- a/mmstudio/src/main/java/org/micromanager/acquisition/SequenceSettings.java
+++ b/mmstudio/src/main/java/org/micromanager/acquisition/SequenceSettings.java
@@ -80,7 +80,17 @@ public class SequenceSettings {
          useCustomIntervals = use; return this; }
       public Builder customIntervalsMs(ArrayList<Double> c) {
          customIntervalsMs = c; return this; }
-      public Builder channels(ArrayList<ChannelSpec> c) {channels = c; return this;}
+      public Builder channels(ArrayList<ChannelSpec> c) {
+         // avoid inserting null channels
+         channels = new ArrayList<>();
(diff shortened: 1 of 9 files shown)


mmstudio/src/main/java/org/micromanager/acquisition/SequenceSettings.java
```diff
@@ -211,10 +221,11 @@ public SequenceSettings build() {
    public ArrayList<Double> customIntervalsMs = null;
    /**
     * an array of ChannelSpec settings (one for each channel)
+    * no member of the array should ever by null
     * @deprecated use Builder and channels() instead
     */
    @Deprecated
-   public ArrayList<ChannelSpec> channels = new ArrayList<>();
+   private ArrayList<ChannelSpec> channels = new ArrayList<>();
```
marktsuchida: Was this perhaps left `private` by accident, given your description (and the remaining `@Deprecate`)?

marktsuchida:
Thanks, @nicost . I've linked the issue because we can be reasonably sure that avoiding direct access to the deprecated fields will prevent ever introducing `null`s. Everything looks great except for my one question above.

mmstudio/src/main/java/org/micromanager/acquisition/SequenceSettings.java
nicost: Yes!  Excellent spot.  Reverted in an extra commit.

PR #945, opened 2020-09-15
Replace "–" with "-".

zhangxiaosa:
Replace "–" with "-", because the former doesn't belong to ascii code.

Other non-ascii characters that appear in source code(excluding name and long comments in foreign languages such as Chinese):

�
micro-manager\DeviceAdapters\CoboltOfficial\LaserShutterProperty.cpp:
   31: //                specified in owner�s manual may result in exposure to hazardous radiation and

micro-manager\DeviceAdapters\CoboltOfficial\LaserShutterProperty.h:
   31: //                specified in owner�s manual may result in exposure to hazardous radiation and

micro-manager\DeviceAdapters\HydraLMT200\ITKHydra.cpp:
    9: // AUTHOR:        Steven Fletcher, derived from Corvus adapter written by Johan Henriksson, mahogny@areta.org, derived from M�rzhauser adapter  
  346:    // Accel (Acceleration (in m/s�)
  397:  * Returns current position in �m.

…
micro-manager\DeviceAdapters\dc1394\dc1394.h:
   61: // 00 0A 47 …. Node_Vendor_Id  

±
micro-manager\DeviceAdapters\ITC18\MMITC18.cpp: 
   65: const char* g_ITC18_AD_RANGE_10VString = "±10 Volt";
   66: const char* g_ITC18_AD_RANGE_5VString = "±5 Volt";
   67: const char* g_ITC18_AD_RANGE_2VString = "±2 Volt";
   68: const char* g_ITC18_AD_RANGE_1VString = "±1 Volt"; 

¡¯
micro-manager\DeviceAdapters\Mightex_BLS\Mightex_BLSDriver_SDK.h:
    2: #define  MAX_PULSE_COUNT     21   // For SX Modules, it¡¯s 3 instead of 128.

“ and ”
micro-manager\DeviceAdapters\Omicron\OmicronDeviceDriver.cpp:
   16: //[DllImport(“msvcrt.dll”, CallingConvention = CallingConvention.Cdecl)]

‘ and ’
micro-manager\DeviceAdapters\PCO_Generic\lib\pco_generic\sc2_SDKStructures.h:
  414:   WORD  wSize;                         // Sizeof ‘this’ (for future enhancements)
  449:   WORD              wSize;             // Sizeof ‘this’ (for future enhancements)

’
micro-manager\DeviceAdapters\ZeissCAN\ZeissCAN.cpp:
  617:       // ensure we’ve been provided with a valid serial port device name

† and ‡
[...]

Issue #669, opened 2019-03-09
[MM2] error: unmappable character for encoding ASCII


cErikson:
I'm running into this error a lot during compiling. Usually when every there is a mu symbol in a source file. 
I've managed to get a work around by setting `JAVA_TOOL_OPTIONS=-Dfile.encoding=UTF8` and adding `-Dfile.encoding=UTF8` to ANTFLAGS

marktsuchida:
This is why it is good practice to escape non-ASCII characters in Java source....

The fix for this will be to set up an _automatic_ scan for non-ASCII characters in source code; otherwise it's bound to keep happening. Until then I'm keeping this open.

marktsuchida:
@zhangxiaosa has kindly created a list: #945.

PR #949, opened 2020-09-17
Remove or escape non-ascii chars in Java source

marktsuchida:
This fixes the last 4 files listed in #945 by @zhangxiaosa.


marktsuchida:
See also: #669.

Issue #818, opened 2020-05-29
Pixel Size Calibration silently accepts presets with identical property values


nicost:
Simplest way to reproduce this using the demo configuration:
- Remove All Pixel Size configurations
- Add a config with only the DObjective label, set it to 10x.
- Create a new config, change the name but leave the DObjective set to 10x. Micro-Manager accepts this setting, but will display 0 nm/px in the main display.  Only one of these two identical configs will be matched.  

It surprises me that the Core allows this setting, and consider that the main bug.

marktsuchida:
Fixed, thanks @zhangxiaosa!

nicost:
Great that this is fixed in the UI, however, the Core should never allow for this, so this should be fixed there as well.

marktsuchida:
Let's make that a separate issue (I'll create it), because the normal config groups have the same problem, not just the pixel size ones.

nicost:
Wow, did not realize that.  Thanks!

marktsuchida:
@nicost Yeah, I was a little surprised, too, but it is also complicated by how `CMMCore::defineConfig()` works, because each property value is set as a separate function call.

PR #951, opened 2020-09-19
Fix issue #818

zhangxiaosa:
Logic has been added to check against identical properties combination in Pixel Size Calibration.

### DIFF
--- a/mmstudio/src/main/java/org/micromanager/internal/dialogs/PixelPresetEditor.java
+++ b/mmstudio/src/main/java/org/micromanager/internal/dialogs/PixelPresetEditor.java
@@ -25,14 +25,11 @@ import java.awt.HeadlessException;
 import java.awt.geom.AffineTransform;
 import java.text.ParseException;
 import javax.swing.JOptionPane;
+import mmcorej.Configuration;
 import mmcorej.DoubleVector;
 import org.micromanager.Studio;
 import org.micromanager.internal.MMStudio;
-import org.micromanager.internal.utils.AffineUtils;
-import org.micromanager.internal.utils.NumberUtils;
-import org.micromanager.internal.utils.PropertyItem;
-import org.micromanager.internal.utils.PropertyTableData;
-import org.micromanager.internal.utils.ReportingUtils;
+import org.micromanager.internal.utils.*;
 
 /**
  *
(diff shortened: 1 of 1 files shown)


mmstudio/src/main/java/org/micromanager/internal/dialogs/PixelPresetEditor.java
```diff
@@ -154,15 +184,28 @@ public boolean writeGroup(String newName) {
          }
 
          core_.definePixelSizeConfig(newName);
-
-         for (PropertyItem item : data_.getProperties()) {
-            if (item.confInclude) {
-               core_.definePixelSizeConfig(newName, item.device, item.name, item.getValueInCoreFormat());              
+         // if no duplicate presets will be created, then allow this preset to be added
+         if (!same) {
```
marktsuchida: I think this check is not necessary because if `same` were set to `true` above, we would have returned from this method and never reach here.

mmstudio/src/main/java/org/micromanager/internal/dialogs/PixelPresetEditor.java
```diff
@@ -154,15 +184,28 @@ public boolean writeGroup(String newName) {
...
+         // if fails to edit an existing preset due to the duplicate presets, keep the old preset
+         else if(!newItem_) {
+            for (PropertyItem item : data_.getProperties()) {
+               if (item.confInclude) {
+                  core_.definePixelSizeConfig(newName, item.device, item.name, oldPreset.getSetting(item.device, item.name).getPropertyValue());
+               }
             }
+            pixelSize_ = pixelSizeField_.getText();
+            core_.setPixelSizeUm(newName, NumberUtils.displayStringToDouble(pixelSize_));
+            core_.setPixelSizeAffine(newName, affineEditorPanel_.getAffineTransform() );;
```
marktsuchida: Note that these lines are nearly identical to the lines above. It is best to avoid repeating code like this, because it makes the code harder to maintain (for example, multiple places will need to be updated when making a change, and it is easy to forget to do so).

In this case, however, I would avoid these lines entirely, instead giving control back to the user: after editing an existing preset A and making its properties identical to preset B, the user should stay in the preset editor after the error message instead of canceling the edit. They can then explicitly cancel, or chose different property values.

marktsuchida (changes requested):
Great start! As I wrote in another comment, it will be better not to close the preset editor in the case where an existing property is edited to make it identical to another one. Otherwise this is looking good!

mmstudio/src/main/java/org/micromanager/internal/dialogs/PixelPresetEditor.java
marktsuchida: I missed the `continue`!

marktsuchida (approved):

PR #963, opened 2020-09-25
Fix issue#953 by checking the connection.

zhangxiaosa:
A check on connection to the device is added before doing any initialization. 
If no connection is detected, the program will return an error code immediately.

### DIFF
--- a/DeviceAdapters/Aladdin/Aladdin.cpp
+++ b/DeviceAdapters/Aladdin/Aladdin.cpp
@@ -132,17 +132,20 @@ int AladdinController::Initialize()
 {
    this->LogMessage("AladdinController::Initialize()");
 
-   GeneratePropertyVolume();
-   GeneratePropertyDiameter();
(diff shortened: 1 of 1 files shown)


marktsuchida:
@zhangxiaosa Looks good, thanks!

There is still the possibility that some other device is connected to the serial port, in which case `Purge` would not trigger an error but subsequent commands may crash. However, I can see that fixing that would not be easy without being able to test with the actual device, so let's leave it here.

@nicost Is it safe to merge this on Git? (In which case, can you set the SVN Aladdin directory to be read-only?) Or would it be better if I commit this via SVN?

nicost:
Much easier to commit to svn.  If we want to move to git for C++ code, we should do that for all code at once, rather than one at a time.

marktsuchida:
@nicost Ok, will do.

nicost:
Committed to svn in revision 17408

PR #967, opened 2020-09-30
Allow log manager to deal with all Throwables instead of just Exceptions

nanthony21:
The `LogManager` and `ReportingUtils` are written such that they only work with `Exceptions`. However they only actually use methods from `Throwable`, a base class of `Exception`.

This means that the `LogManager` is unable to deal with other `Throwable`s such as `Error`s.

Even though non-`Exception` throwables are less common there is no reason no to allow the `LogManager` to deal with them.

### DIFF
--- a/mmstudio/src/main/java/org/micromanager/LogManager.java
+++ b/mmstudio/src/main/java/org/micromanager/LogManager.java
@@ -55,13 +55,13 @@ public interface LogManager {
     * @param e - Java exception to be logged
     * @param msg - message to be shown
     */
-   public void logError(Exception e, String msg);
+   public void logError(Throwable e, String msg);
 
    /**
     * Writes a stacktrace to the Micro-Manager log.
     * @param e - Java exception to be logged
     */
-   public void logError(Exception e);
+   public void logError(Throwable e);
 
    /**
     * Writes an error to the Micro-Manager log (same as logMessage).
@@ -74,13 +74,13 @@ public interface LogManager {
     * @param e - Java exception to be shown and logged
     * @param msg - Error message to be shown and logged
     */
-   public void showError(Exception e, String msg);
+   public void showError(Throwable e, String msg);
 
    /**
     * Shows and logs a Java exception.
-    * @param e - Java excpetion to be shown and logged
+    * @param e - Java exception to be shown and logged
     */
-   public void showError(Exception e);
+   public void showError(Throwable e);
 
    /**
     * Shows an error message in the UI and logs to the Micro-Manager log.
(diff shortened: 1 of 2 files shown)


marktsuchida:
@nanthony21 Interesting suggestion. Would you mind commenting on the use case for this?

`Error`s are only used to report unrecoverable conditions; the thread will crash, and the stack trace will be logged (because MMStudio sets up an uncaught exception handler that does so). In the vast majority of cases, it is desirable to let the thread crash upon an `Error`, because there is (usually) no correct way to recover and continuing to run only makes diagnosis more difficult. To put this another way, our code should not catch non-`Exception` `Throwable`s except in very special cases, such as when the error needs to be propagated by another mechanism.

Could it perhaps be that you are dealing with some other API that passes around `Throwable`s (as opposed to catching `Throwable` yourself)?

I think changing `LogManager` to accept `Throwable` is technically okay, but I do worry a little that it could encourage code that casually swallows unrecoverable `Error`s. (And we need every chance to prevent such code, because said swallowing will rarely be tested.) Given that 99% of correct uses for `LogManager` should be with `Exception` only, there is some advantage to having the parameter limited to that. So I would like to consider the actual use cases and make sure this is the best solution.

nanthony21:
@marktsuchida you are right the the instances where this change would actually be used are very few.

The situation where this arose for me was from these lines in a plugin that I am working on.

```
try {
   future.get(); //If an exception was thrown in the thread this will cause it to be thrown here as an ExecutionException.
} catch (ExecutionException ee) {
   studio_.logs().logError((Exception) ee.getCause());
}
```

In normal usage `ee.getCause()` would only be an `Exception`. However , due to a classpath difference between debugging environment and normal micro-manager runtime environment I had a situation where `ee.getCause()` was returning `ClassDefNotFoundError`. However instead of getting a log of that error I only ended up with a log of the `ClassCastException` from trying to cast to `Exception`.

I don't think that this change would encourage sloppy handling of `Error`s, but I also recognize that The benefit offered by this PR is minimal. I'll leave it up to your judgement.

marktsuchida:
Ah! It was `ExecutionException` that I was trying to think of when I asked if this came from the use of some API.

I think in this specific case it is preferable to work around, e.g. by passing `ee` directly to `logError`.

If you want to be even more technically correct, you could do something like
```
    Throwable t = ee.getCause();
    if (t instanceof Exception) {
        studio_.logs().logError((Exception) t);
        // Handle recoverable error here (e.g. show error message)
    }
    else {
        throw new RuntimeException("Fatal error in xyz", t); // Crash on Error
    }
```

This is cumbersome, but I don't think there is much that can be done to simplify it, given the JDK API.

PR #972, opened 2020-10-02
add documentation to device adapter and PycromanagerJava to 0.8.21

henrypinkard:


nicost:
Hi @henrypinkard !

The general workflow is still that changes in the C++ layer (Device adapter and Core) go into the subversion repo first, which then is (manually) synchronized with the 2.0 source on github.  I think that you still have commit access to svn, if not, let me know and I'll commit your Teensy changes.  If you can take that out of this PR, then it is easy for us to apply.  Thanks!

henrypinkard:
I actually don't have any of this setup on my current computer. That would be great if you could add this over, thanks. And I'd like to add this into the build as well. Does that happen through svn also?

I'll close this and reopen another one with just ivy change

nicost:
Hi @henrypinkard !

Committed to svn (which is a lot easier to setup Pycharm with Jupyter;). I don't think this firmware code makes it into the build at all.  The MM source code repo is more a convenient download side (and nicely in sync with the device adapter code).  It will sync to the 2.0 source soon.

henrypinkard:
@nicost Thanks!

PR #974, opened 2020-10-04
Fix issue955.

zhangxiaosa:
Replace 2*2 affine transformation table with 3*3.

### DIFF
--- a/mmstudio/src/main/java/org/micromanager/internal/dialogs/AffineEditorPanel.java
+++ b/mmstudio/src/main/java/org/micromanager/internal/dialogs/AffineEditorPanel.java
@@ -1,7 +1,7 @@
 
 package org.micromanager.internal.dialogs;
 
-import java.awt.Component;
+import java.awt.*;
 import java.awt.event.ActionEvent;
 import java.awt.geom.AffineTransform;
 import java.text.NumberFormat;
@@ -65,6 +65,7 @@ public class AffineEditorPanel extends JPanel {
       table.setModel(atm_);
       table.putClientProperty("terminateEditOnFocusLost", Boolean.TRUE);
       table.setCellSelectionEnabled(true);
+      table.setBorder(BorderFactory.createLineBorder(Color.GRAY,1));
       
       for (int col = 0; col < table.getColumnModel().getColumnCount(); col++) {
          table.getColumnModel().getColumn(col).setMaxWidth(75);
@@ -72,10 +73,9 @@ public class AffineEditorPanel extends JPanel {
       super.add( new JLabel("<html><center>Affine transforms define the relation between <br> " +
               "camera and stage movement</center></html>"), " span 2, center, wrap") ;
       table.setFillsViewportHeight(true);
-      
-      super.add(table.getTableHeader(), "flowy, split 2, shrink 100");
+
       super.add(table);
-      
+
       JButton calcButton = new JButton("Calculate");
       calcButton.addActionListener((ActionEvent e) -> {
          calculate();
@@ -190,12 +190,6 @@ public class AffineEditorPanel extends JPanel {
       
       public void setAffineTransform(DoubleVector aft) {
          affineTransform_ = copyDoubleVector(aft);
-         // ensure that there is no translation component in the affine transform
-         // setting this accidentally can lead to bad stuff.
-         // Not sure if this should be in the core instead
-         affineTransform_.set(2, 0.0);
-         affineTransform_.set(5, 0.0);
-         
          fireTableDataChanged();
       }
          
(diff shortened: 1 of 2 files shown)


marktsuchida:
Thanks, @zhangxiaosa!

We should ensure that the automatic Pixel Calibrator does not generate non-zero translations before merging this, because code that forces that has been removed from the pixel size preset editor. In the meantime, @nicost: how do you feel about this GUI? We kept the table (instead of replacing with labels and text fields) because @zhangxiaosa found that it was not hard to disable the 3rd row -- but I suggested we remove the header (which doesn't really convey anything).

nicost:
Can you post a screenshot?  I agree with removing the header.  There was code in there previously to avoid setting translation to anything but 0, but I think that code was in the wrong place (inside the table if my memory is correct, rather than in the automatic Pixel Calibrator code).

marktsuchida:
@nicost Here's a screenshot. Yes, the code to ensure translation is [0 0] was in the wrong place, and this PR (rightly) deletes it. Therefore I would like to look at the Pixel Calibrator code and ensure it always generates zero translations before merging this PR.
Screenshot 2020-10-07 164442

nicost:
Looks good!  Relevant code is in org.micromanager.internal.pixelcalibrator.PixelCalibratorDialog.promptToSaveResult(). 

Add the lines that set the translation part of the affine transform (called 'result') to zero and that should be all.

zhangxiaosa:
> Looks good! Relevant code is in org.micromanager.internal.pixelcalibrator.PixelCalibratorDialog.promptToSaveResult().
> 
> Add the lines that set the translation part of the affine transform (called 'result') to zero and that should be all.

Thanks that's helpful. But if I set translation to zero, how should I set the `type` and `state` of `AffineTransform` class? There is no method to set translation to zero, can I implement one in `AffineTransform` class?

nicost:
Same as the code previously in the AffineEditorTable did:

```
affineTransform_.set(2, 0.0);
affineTransform_.set(5, 0.0);
```

where affineTransform_ is called `result` in PixelCalibratorDialog.promptToSaveResult().

zhangxiaosa:
> Same as the code previously in the AffineEditorTable did:
> 
> ```
> affineTransform_.set(2, 0.0);
> affineTransform_.set(5, 0.0);
> ```
> 
> where affineTransform_ is called `result` in PixelCalibratorDialog.promptToSaveResult().

We can't do this because `affineTransform_` is `DoubleVector`, whie `result` is `AffineTransform`, which doesn't allow to change a specific element. It only has methods like `setToIdentity`, `setTotranslation` and `setToRotation`.

nicost:
Ah, yes.  Can you use:

```      result.setToTranslation(0.0, 0.0); ```

on something like line 256 of PixelCalibratorDialog.
That should ensure that the x and y translation components are always set to 0.

**Comment by zhangxiaosa** [comment]
> Ah, yes. Can you use:
> 
> ` result.setToTranslation(0.0, 0.0);`
> 
> on something like line 256 of PixelCalibratorDialog.
> That should ensure that the x and y translation components are always set to 0.

I'm not sure if we can use `result.setToTranslation(0.0, 0.0);` because it will set the transformation back to identity transformation, so it may change another 4 elements. I don't know if it is correct to do this before `pixelSizeProvider_.setAffineTransform(result);` in org.micromanager.internal.pixelcalibrator.PixelCalibratorDialog.promptToSaveResult().

```
if (response == JOptionPane.YES_OPTION) {
   if (pixelSizeProvider_ != null) {
      pixelSizeProvider_.setPixelSize(pixelSize);
      // result.setToTranslation(0.0, 0.0);??
      pixelSizeProvider_.setAffineTransform(result);
   }
}
```

**Comment by nicost** [comment]
I doubt that setToTranslation will set the transform to identity (that would be quite unhelpful;)

The Javadocs say: `Sets this transform to a translation transformation.`

I tested it, and it works exactly as expected (i.e., only sets the Translation component).  

Adding 

`      result.setToTranslation(0.0, 0.0);`

after the null check (i.e. on line 256) will do the right thing.

**Comment by zhangxiaosa** [comment]
Thanks. It is added.

**Comment by marktsuchida** [comment]
Sorry to take so long to get back to this. Unfortunately @zhangxiaosa is right about `setToTranslation()`: it does reset the 2x2 part to the identity matrix (hence `setToTranslation` rather than `setTranslationTo`).

```
bsh % import java.awt.geom.AffineTransform;
bsh % t = AffineTransform.getRotateInstance(1.57);
<AffineTransform[[7.96326710733E-4, -0.999999682931835, 0.0], [0.999999682931835, 7.96326710733E-4, 0.0]]>
bsh % t.setToTranslation(0.5, 0.5);
bsh % print(t);
AffineTransform[[1.0, 0.0, 0.5], [0.0, 1.0, 0.5]]
```

So I think a combination of `getMatrix()` and `setTransform()` (or just create a new instance using the constructor taking `double[]`) needs to be used.

```
import java.awt.geom.AffineTransform;
t = AffineTransform.getTranslateInstance(0.2, 0.3);

double[] m = new double[6];
t.getMatrix(m);
m[4] = 0.0;
m[5] = 0.0;

tt = new AffineTransform(m);
print(tt);
```

zhangxiaosa:
@marktsuchida Yes It makes sense to me and I have updated it.

nicost:
Wow. Thanks!  Surely got confused while testing!  How unhelpful of Java!

PR #975, opened 2020-10-06
Fix issue#965.

zhangxiaosa:
Fix the bug that color is not updated before the channel is added to `channel_`.
Fix the logic to process click on `edit` button(the color bar in the 6th column) and `ok` button.

### DIFF
--- a/mmstudio/src/main/java/org/micromanager/internal/dialogs/ChannelTableModel.java
+++ b/mmstudio/src/main/java/org/micromanager/internal/dialogs/ChannelTableModel.java
@@ -163,6 +163,7 @@ public final class ChannelTableModel extends AbstractTableModel  {
       }
       channel = cb.color((Color) value).build();
       channels_.set(row, channel);
+      storeChannels();
 
       this.fireTableChanged(new TableModelEvent(this));
    }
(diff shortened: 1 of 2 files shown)


marktsuchida:
@zhangxiaosa Thanks! It appears to work correctly, _except_ that the choice of color is lost after restarting Micro-Manager. (Compare this to how all the other settings in the Multi-D Acquisition dialog are saved between sessions.)

Could you take a look to see if you can fix that, too? It might help to look for where the default (initial) colors come from when the MDA dialog is created. If the fix is not simple, let me know; in that case I'll go ahead and merge this PR and we can work on it as a separate issue.

@nicost Pinging you in case you have any helpful comments, since I think you were working on this code most recently.

nicost:
There is a ChannelTableModel.storeChannels method that should do the job.  It may be good to call it every time anything changes in the channel definitions (i.e. at the end of the ChannelTableModel.setValueAt method). I suspect that will do the job.

zhangxiaosa:
I simply added storechannels() at the end of setvalueat(), it works.

nicost:
Awesome!  Thanks Max!

PR #983, opened 2020-10-17
Fix da

zhangxiaosa:
Ask the core for the device pointer each time we access the device, instead of saving the pointer for future use. The latter is dangerous because the pointer may have been deleted.

### DIFF
--- a/DeviceAdapters/Utilities/Utilities.cpp
+++ b/DeviceAdapters/Utilities/Utilities.cpp
@@ -159,7 +159,6 @@ MultiShutter::MultiShutter() :
 
    for (int i = 0; i < nrPhysicalShutters_; i++) {
       usedShutters_.push_back(g_Undefined);
-      physicalShutters_.push_back(0);
    }
 }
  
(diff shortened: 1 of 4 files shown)


nicost:
Hi Max,

Changing the device interface (MMDevice.h) is quite drastic, and I am not sure why these interface functions should no longer be const.  Having them be const helps avoid side-effects.  Which member variables do you need to change within these functions?

zhangxiaosa:
Hi Nico,
This is a bad implementation. The reason I remove const from methods like `GetImageWidth()` is that is need to call a non const method `GetDevice()`, which is not allowed.
But maybe it's better to make `GetDevice()` const and I have tried this just now, it can work.

marktsuchida:
@zhangxiaosa The code changes look good, and I like your solution of adding `const` to `GetDevice` (which is appropriate and also backward-compatible, because any existing code calling `GetDevice` will continue to compile). (You are right that it is questionable that `GetImageWidth` etc. are `const`, but we don't want to change that.)

I'd like to do a little manual testing before merging this, since it's an important module, but so far I don't expect any changes to be needed.

nicost:
Committed in svn commit 17409.  Will trickle through to github soon.

PR #984, opened 2020-10-18
"AcqEngJ" rev="0.7.4",  "PycroManagerJava" rev="0.9.3"

henrypinkard:


buildscripts/ivy.xml
```diff
@@ -1,4 +1,4 @@
-<ivy-module version="2.0" xmlns:e="http://ant.apache.org/ivy/extra">
+ <ivy-module version="2.0" xmlns:e="http://ant.apache.org/ivy/extra">
```
marktsuchida: Sorry this feels like nitpicking, but would appreciate if you could remove this spurious space change.

henrypinkard: done

Issue #1023, opened 2020-11-18
MultiStagePosition has a static function with a MultiStagePosition instance as an argument


nicost:
org.micromanager.MultiStagePosition contains the static function:

`public static void goToPosition(MultiStagePosition msp, CMMCore core_) throws Exception`

It seems strange not use:

`public void goToPosition(CMMCore core) throws Exception `

instead.

In addition, the function is inefficient as it moves each stage, waits for it to stop moving, and then moves the other.  Also, the `core_` argument should be `core` according to the style guide.

marktsuchida:
Agreed. This should be removed from `MultiStagePosition` entirely if possible, making the class not dependent on MMCore and closer to a clean pure-data object. Not yet sure exactly where to place it, although it might be good to consider a stage control utility class.

Also agree that movements should be fully parallel.

The method is currently used by the GUI actions to move the stage, in the Position List and Create Grid dialogs.

nicost:
Have a look at PR#1025, it may not be completely what you have in mind, but a step in the right direction, and would allow our GUI to call this better implementation.  Or we can wait until you have a good idea where the goToPosition method should move.

marktsuchida:
Yeah, sorry I commented before seeing #1025; also added a comment there.

PR #1025, opened 2020-11-18
Studio: adds goToPosition as a member function in MultiStageP. Fixes issue #1023

nicost:
Deprecates the static function with same name. Also first sends all stages to move , and then waits for all of them to finish, rather then moving them one by one.

Fixes issue #1023

### DIFF
--- a/mmstudio/src/main/java/org/micromanager/MultiStagePosition.java
+++ b/mmstudio/src/main/java/org/micromanager/MultiStagePosition.java
@@ -242,11 +242,39 @@ public final class MultiStagePosition {
    }
 
    /**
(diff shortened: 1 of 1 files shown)


marktsuchida (changes requested):
As I commented on #1023, I think we should take this opportunity to move this method somewhere else (possibly a `StageManager`?), in the interest of separating data and device control. It would be nice to make the change only once, so would like to keep this pending until we decide on a final location. Please feel free to go ahead and create such a class.

The method code itself looks good, except that the deprecated static method should be changed to call the new non-static method, to avoid code duplication.

PR #1034, opened 2020-12-04
Save/Load overlay settings from user profile

nanthony21:
This PR changes the way that overlay settings are saved so that they can be successfully reloaded from the user profile. This makes overlays behave like most other aspects of Micro-Manager where preferences are automatically saved and recalled.

### DIFF
--- a/mmstudio/src/main/java/org/micromanager/display/inspector/internal/panels/overlays/OverlaysInspectorPanelController.java
+++ b/mmstudio/src/main/java/org/micromanager/display/inspector/internal/panels/overlays/OverlaysInspectorPanelController.java
@@ -110,36 +110,36 @@ public final class OverlaysInspectorPanelController
    }
 
    private void loadSettings(Iterable<OverlayPlugin> plugins) {
-    //Load the overlays from the profile.
-    List<PropertyMap> settings = profile_.getSettings(this.getClass()).getPropertyMapList(PROPERTYMAPKEY, (PropertyMap[]) null);
-    if (settings == null) {
-        return;
-    }
-    for (PropertyMap pMap : settings) {
-       for (OverlayPlugin p : plugins) { // We must loop through overlay plugins to determine if they are a match for this setting.
-          Overlay o = p.createOverlay();
-          if (pMap.getString(TITLEPMAPKEY, "loadFailed").equals(o.getTitle())) {  // Checking against Overlay 'Title; is the best way we have to link settings with an overlay.
-              PropertyMap config = pMap.getPropertyMap(CONFIGPMAPKEY, null);
-              o.setConfiguration(config);
-              o.setVisible(pMap.getBoolean(VISIBLEPMAPKEY, false));
-              viewer_.addOverlay(o);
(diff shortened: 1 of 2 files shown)


nicost:
This is wonderful!  Had been annoyed by the same thing, but never got around doing anything.  Will have a quick look this weekend and merge.

nanthony21:
@nicost one thing I'll mention that is a little messy about this is that in order to load the overlay settings we have to loop through all available overlay plugins until we find one that creates an overlay with a matching title to the title saved in the profile. However, I don't see a better way to do this without changing the Overlay / OverlayPlugin API.

nicost:
What API addition would help?  We can ask @marktsuchida if we can add the needed code

nanthony21:
I think we would just need for each instance of `Overlay` to have a reference back to the `OverlayPlugin` class that created it.

That way the profile could just save the name of the correct `OverlayPlugin` for each `Overlay` rather than the current approach which amounts to saving `Overlay.getTitle()` to the profile and then during the load step having to do:
```
for  (OverlayPlugin plugin : plugins) {
    Overlay o = plugin.createOverlay();
    if (o.getTitle().equals(profileSavedTitle) {
       //Load overlay settings
    } else {
        //Discard this overlay instance and try the next plugin
    }
}
```

nicost:
Hi @nanthony21,

Checked out your branch and played around with it. I love the intended behavior, this is absolutely how things should be!

However, I did uncover a bit of a bug.  When I snap an image, add a scalebar, then run an MDA, it nicely also adds a scalebar to that MDA.  When I then click in the Snapped image window, an extra overlay panel appears in the inspector, and after each switch between windows there is a new one.  I can see this may be difficult to deal with since multiple instances of an overlay should be possible.  Since you may still remember the relevant code, could you have a look?

Also, I don't understand what API change you propose.  Which class/interface needs an additional method, and what should it look like?

nanthony21:
Hi @nicost ,

Thanks for finding that issue, it should be fixed now.


The API change is not strictly necessary. The reason to make a change is that there is currently no way to find out which `OverlayPlugin` created an `Overlay`. This means that in order to instantiate an `Overlay` from saved settings we need to loop through all available `OverlayPlugins` just to see if it creates an `Overlay` that is a match for the settings. What I would propose is adding a method to the `Overlay` class that is along the lines of: `public Class getOverlayPlugin();`

nicost:
Cool!  I am going to merge the PR, but still find some funny things with the behavior.  I noted that when a DisplayViewer becomes inactive, the Overlays disappear.  Not sure if that is new or old behavior, but it is kind of strange.  Also, as a user I expect the Overlays to be tied to the display settings for a DisplayViewer, i.e. when I change the Overlays in a specific viewer, and then switch to another Viewer, I do not expect my changes to follow with me, but to see the Overlays as I left them in that Viewer.  

I think the same problem applies to other aspects of the InspectorPanel, so may not be worth looking into now, but certainly should be overhauled at one point and the behavior should be made consistent and pleasant;)

PR #1047, opened 2020-12-10
Improve inspector panel and overlay settings.

nanthony21:
This PR makes the following changes:

1: Previously all UI content of the `Inspector` panel would be removed and completely reconstructed each time the `Inspector` was  attached to a new `DataViewer`. This led to some confusing behavior where code in the `detachDataViewer` method of `AbstractInspectorPanelController` implementations would be ineffective since a completely new instance of every `InspectorPanelController` was being created for every new attachment event. Now each `InspectorPanelController` is only constructed once during the construction of a new `Inspector` panel. The only notable difference to a user is that the `Inspector` panel will now show a section for every available `InspectorPanelController` even if  `InspectorPanelPlugin.isApplicableToDataViewer` returns `false`. However, the panel will be collapsed and disabled in this case.

2: The `detachDataViewer` method of each `InspectorPanelController` was being called multiple times for each attachment to a new `DataViewer`. The code has been simplified so that it is now called only once.

3: Overlay settings are now saved and loaded based on the name of the `DataProvider` attached to the `DataViewer`. This means that settings for a "Preview" window will be separate from overlay settings for an "Album" window, etc.

### DIFF
--- a/mmstudio/src/main/java/org/micromanager/display/inspector/AbstractInspectorPanelController.java
+++ b/mmstudio/src/main/java/org/micromanager/display/inspector/AbstractInspectorPanelController.java
@@ -95,7 +95,7 @@ public abstract class AbstractInspectorPanelController
    public abstract void attachDataViewer(DataViewer viewer);
 
    /**
-    * Detach from any currently attached data viewer.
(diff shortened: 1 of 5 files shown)


nicost:
Great improvement!  Works well in my (limited) tests.  Thanks so much!

PR #1048, opened 2020-12-11
Extends ThorlabsElliptec to ELL17 and ELL20

jdeschamps:
* Adds ELL17/ELL20 device
* Fix errors in ELL6 and ELL9 devices


nicost:
Committed to subversion, will make it into 2.0 soon.

PR #1061, opened 2020-12-22
Split MMStudio into separate classes

nanthony21:
The MMStudio class is very long and handles many seemingly unrelated functionalities. There is still a lot more that can be done but this PR improves the situation by separating all code related to the `Application` or `CompatiblityInterface` interfaces into separate dedicated classes.


nicost:
This looks really good Nick!  @marktsuchida, I will merge by 12/23 unless you tell me not to.

nanthony21:
Thanks @nicost . 
I'm interested in doing some more to clean up the MMStudio class. One thing I noticed is that there are many public methods in MMStudio that are not part of an interface and do not have any usages within the MMStudio project. Do you know if it is ok to remove these?

nicost:
I only found:  

public boolean getHideMDADisplayOption()

as an unused function.  Did you find more?

nanthony21:
You're right. Maybe that was on a different test branch where I had already done some refactoring of MMStudio.

nicost:
When running this branch, I get:

java.lang.NullPointerException in Thread[AWT-EventQueue-0,6,main]
                                   [       ]   at org.micromanager.internal.MMStudio.logStartupProperties(MMStudio.java:1355)

and I get many of these:

2020-12-23T09:47:27.076824 tid17088 [IFO,App] Unable to instantiate class for class='src.main.java.out.production.main.org.micromanager.display.inspector.internal.panels.intensity.IntensityInspectorPanelPlugin', name='Intensity Scaling', description='View and adjust intensity scaling', priority=10000.0, enabled=true, pluginType=InspectorPanelPlugin
                                    [       ] org.scijava.InstantiableException: Class not found: src.main.java.out.production.main.org.micromanager.display.inspector.internal.panels.intensity.IntensityInspectorPanelPlugin in Thread[Plugin loading thread,6,main]
                                    [       ]   at org.scijava.plugin.PluginInfo.loadClass(PluginInfo.java:292)

which all may be cause by the first error.

Also, there are a bunch of unused imports left in MMStudio.class.  All hopefully not hard to fix, but could you have a look first?

Thanks @nanthony21 !

nanthony21:
@nicost Oops, not sure how that happened. The null pointer error and unused imports are fixed now.

I haven't been able to reproduce the issue with loading the inspector panel plugin though, the "Intensity Scaling" plugin is working fine for me.

nicost:
Looks good! Also not sure what happens during that plugin loading (the plugin loading mechanism is so utterly complicated that I wonder if anyone understands how it works, that SciJava stuff may be nice for some, but I rather work with code that is easy to follow in an IDE).  The Inspector works for me, and I see the same errors on the main branch, so those are very likely unrelated, and hopefully only an annoyance on my development environment.

nanthony21:
I've run into a similar sounding issue many times in Netbeans and IntelliJ where plugins that are defined in the source code rather than in a JAR file do not get found at runtime. I figured out that I could fix it in Netbeans by rearranging the order of "Project Properties -> Libraries -> Run -> Classpath". I never did figure it out for IntelliJ though.

nicost:
Just noted that the Conifguration Settings are no longer updated after this PR was merged.  I think this is because of:

 @Override
   public void refreshGUI() {
      ((MMStudio) studio_).updateGUI(false);
   }

   @Override
   public void refreshGUIFromCache() {
      ((MMStudio) studio_).updateGUI(true);
   }

Note the signatures of updateGUI:

public void updateGUI(boolean updateConfigPadStructure) 
public void updateGUI(boolean updateConfigPadStructure, boolean fromCache),

so, I think that the above code should be changed to:

 @Override
   public void refreshGUI() {
      ((MMStudio) studio_).updateGUI(true, false);
   }

   @Override
   public void refreshGUIFromCache() {
      ((MMStudio) studio_).updateGUI(true, true);
   }


Let me know if that looks right.

nanthony21:
Hi @nicost ,

I think you're right. Looking at the methods before this PR we have:

```
   @Override	
   public void refreshGUI() {	
      updateGUI(true);	
   }	

   @Override	
   public void refreshGUIFromCache() {	
      updateGUI(true, true);	
   }
```

Which is effectively the same as what you have suggested as a correction.

I had originally refactored the `updateGui` method to remove the `updateConfigPadStructure` argument leaving only a single signature `public void updateGui(boolean fromCache)`. This mistake must have been left over when I reversed that change.

PR #1063, opened 2020-12-23
Delegate functionality from MMStudio to helper classes.

nanthony21:
This PR introduces two new classes which take over functionality that was previously performed directly by MMStudio. There should be no changes to the functionality of Micro-Manager, the goal of this PR is just to make MMStudio more organized and readable.

`MMCache` is essentially just a rename of the `StaticInfo` class and replaces these public methods of `MMStudio`:
```updateXYPos
updateXYPosRelative
updateZPos
updateZPosRelative
updateXYStagePosition
getCachedXPosition
getCachedYPosition
getCachedZPosition
getCachedBitDepth
getCachedPixelSizeUm
getCachedPixelSizeAffine
```

The `MMSettings` class is currently an internal class of `MMStudio` and is responsible for setting and retrieving top-level application preferences and replaces these public methods of `MMStudio`:
```
getShouldDeleteOldCoreLogs
setShouldDeleteOldCoreLogs
getShouldRunZMQServer
setShouldRunZMQServer
getCoreLogLifetimeDays
setCoreLogLifetimeDays
getCircularBufferSize
setCircularBufferSize
```

These two classes can be accessed from an instance of MMStudio using the `cache()` and `settings()` methods.

### DIFF
--- a/mmstudio/src/main/java/org/micromanager/acquisition/internal/DefaultAcquisitionManager.java
+++ b/mmstudio/src/main/java/org/micromanager/acquisition/internal/DefaultAcquisitionManager.java
@@ -277,12 +277,12 @@ public final class DefaultAcquisitionManager implements AcquisitionManager {
       Metadata.Builder result = image.getMetadata().copyBuilderWithNewUUID()
          .camera(camera)
          .receivedTime(DATE_FORMATTER.format(new Date()))
-         .pixelSizeUm(mmstudio.cache().getCachedPixelSizeUm())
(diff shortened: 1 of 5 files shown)


nicost:
Looks good!  Wouldn't this also be a good time to remove all the static variables in what is now called MMCache?  I do not see a good reason these should be static (but I may be overlooking the reason), and if anyone would ever want to run two Studio instances in one JVM, this is surely going to be a problem. I guess all these variables should be private and have getters.

nanthony21:
Good idea. I've converted all the `public static` variables to `private` and added getters where needed.

PR #1067, opened 2020-12-26
Move UI code in MMStudio to UI Manager

nanthony21:
This PR moves all GUI related code in `MMStudio` to a new `MMUIManager` class

An instance of the UI manager can be accessed using the public `MMStudio.uiManager()` method.

The following methods have been moved:

```
public void showPipelineFrame() 
public MMMenuBar getMMMenubar()
public void promptToSaveConfigPresets() 
public void createPropertyEditor() 
public void createCalibrationListDlg() 
public CalibrationListDlg getCalibrationListDlg()
private void createScriptPanel()   
public PipelineFrame getPipelineFrame() 
public boolean isClickToMoveEnabled() 
public void markCurrentPosition() 
private void configureBinningCombo() throws Exception
public void initializeGUI()
public void updateGUI(boolean updateConfigPadStructure)
public void updateGUI(boolean updateConfigPadStructure, boolean fromCache)
public void enableRoiButtons(final boolean enabled)
```


nicost:
Sorry for the delay!  

I tried running the code, but acquisitions failed.  The acquisition engine calls the function MMStudio.getVersion() (and possible others), that is no longer present, hence fails.  I tried changing the acquisition engine code to find that function in its new place, but I have no real understanding of Clojure or even how/when that code is compiled, so did not get far.  Unless you can fix the engine as well, we should leave the functions it needs in MMStudio (and annotate them as deprecated).  

I got some acquisitions to work by adding the following code to MMStudio:

 /**
    * Used by Clojure acquisition engine only
    * @return
    */
   @Deprecated
   public String getVersion() {
      return compat().getVersion();
   }
   /**
    * Used by Clojure acquisition engine only
    * @deprecated Do not use
    */
   @Deprecated
   public void enableRoiButtons(final boolean enabled) {
      ui_.frame().enableRoiButtons(enabled);
   }

Not elegant, so if you have a better solution...

This will go away once we replace the Clojure engine with the Java engine that Henry wrote (another project that has been stalled for way too long).

Let me know how you want to proceed.

nicost:
Just realized that the getVersion member function was moved already in a previous PR, so that acquisitions can not have worked since that code was merged...  Even more urgent to come up with a solution.

nanthony21:
@nicost , thanks for catching that. Should be fixed by PR #1071.

I ran a quick test of the MDA using all 4 dimensions and I didn't see any other issues pop up. Do you know if there are instructions on how to run the unit tests? It would be nice to be able to use these to avoid such situations.

nicost:
I guess the call to MMStudio.enableRoiButtons in acq_engine.clj (line 705) will also need to be changed?  Unit tests would indeed be very nice, but test code has not been maintained well ( in fact, it would need work to set up a decent structure), and are not run in any organized fashion.

nanthony21:
Acquisition engine is fixed now.

PR #1071, opened 2021-01-03
Fix Call to non-api method of MMStudio in Clojure Acquisition Engine

nanthony21:
After PR #1061 the acquisition engine stopped working since one of the public method of MMStudio that it had been using was no longer present. This PR fixes the clojure code to make the correct call to the equivalent API method.


nicost:
Wow!  You can deal with Clojure!  Watch out, I may ask you for help in the future;)  Thanks!

nanthony21:
Haha, well it took me about 10 minutes of googling just to modify this one line so I don't know how helpful I'll be.

nicost:
You got much farther than I ever did! Awesome!

PR #1076, opened 2021-01-07
Move ROI functionality from MMStudio to a dedicated ROI manager

nanthony21:
This PR seeks to more closely align the  `MMStudio` class with the "Single-responsibility principle" by moving public methods used by the `MainFrame` ROI buttons from `MMStudio` to a new class which can be accessed through the new `MMStudio.roiManager()` method.

This PR also makes some formatting changes to the `MMStudio` constructor to make it a bit easier to read by condensing into fewer lines. (I think an 80 character line-length limit is pretty dated at this point)


nicost:
I am one of those who does not like lines that are too long...  I guess I am dated;)  However, no need to change for me, and the changes look good to me.  I'll let Mark have a look but merge by the afternoon 1/12/21/ unless he would like changes.

PR #1078, opened 2021-01-09
Reduce querying of property values when changing a config group.

nanthony21:
When a user changes a config group the `ConfigGroupPad.setValueAt` method first queries all properties associated with any config group in order to see if the user's action has caused any other config groups to change, it then calls the `studio.app().updateGUI()` method which involved a full update of the cached property values.

This PR changes `ConfigGroupPad.setValueAt` so that it starts by explicitly requesting an update of the property cache. It then uses the cached values for both updating the GUI and for checking if config groups have changed. This means that the same properties don't need to be requested from the hardware twice.


marktsuchida:
This is great work! Thanks.

PR #1079, opened 2021-01-11
Reduce redundant function calls.

nanthony21:
This PR contains 3 independent changes that all have to do with avoiding duplicate calls certain functions. Each of the 3 changes are contained in their own commit so it is easy to view each one separately.

1: The `MMStudio.onPropertiesChanged` event handler initiates a refresh of the GUI which involved updating the property cache. However, `CoreEventCallback.onPropertiesChanged` already updates the property cache, there is no reason to do it twice.

2: The `PropertyEditor.onPropertiesChanged` event handler calls `PropertyEditor.refresh` . However,  `MMStudio.onPropertiesChanged` also indirectly calls `PropertyEditor.refresh` through a call to `refreshGUI`. There is no reason to refresh the Property Editor twice.

3: `PropertyEditor.setValueAt` calls `PropertyEditor.refresh` and then immediately calls `studio_.app().refreshGUIFromCache` which results in a second call to `PropertyEditor.refresh`. There is no reason to refresh the Property Editor twice.

### DIFF
--- a/mmstudio/src/main/java/org/micromanager/internal/MMStudio.java
+++ b/mmstudio/src/main/java/org/micromanager/internal/MMStudio.java
@@ -777,7 +777,7 @@ public final class MMStudio implements Studio {
 
    @Subscribe
    public void onPropertiesChanged(PropertiesChangedEvent event) {
-      ui_.updateGUI(true);
+      ui_.updateGUI(true, true);
    }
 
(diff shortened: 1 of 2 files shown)


nicost:
Looks good to me, if Mark does not merge tomorrow, I'll do so.

PR #1089, opened 2021-01-28
Support for OBIS Lasers without Remote used in Omicron LightHUBs

Tim-Omicron:
As the device Adapter for OBIS Lasers only works when the Laser Remote is connected, but OBIS Lasers built in the Omicron LightHUB have no remote, i added a device adapter Version for OBIS Lasers which can be used in this case.

I already commited this to the 1.4 subversion, as i didn't know about this repository. 
Let me know if i can do anything better in the future.

### DIFF
--- a/DeviceAdapters/Omicron/CoherentOBISDirect.cpp
+++ b/DeviceAdapters/Omicron/CoherentOBISDirect.cpp
@@ -1,0 +1,555 @@
+///////////////////////////////////////////////////////////////////////////////
+// FILE:          CoherentOBISDirect.cpp
+// PROJECT:       Micro-Manager
+// SUBSYSTEM:     DeviceAdapters
+//-----------------------------------------------------------------------------
+// DESCRIPTION:   CoherentObis controller adapter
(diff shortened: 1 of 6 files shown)


nicost:
Thanks Tim!  For the time being no need to do a PR for code in the subversion repo here. The subversion repo is semi-automatically included here, i.e. the subversion repo is the sole source for everything C++ in the MM source code.  In fact, it is a lot more work for me to deal with PRs here for C++ code since I have to backport them to the subversion repo first. This may change in the future, but we have been saying that for a long time already;)

nicost:
I can not locate your email @Tim-Omicron, but I just looked at a crash on shutdown of Micro-Manager caused by the Omicron device adapter.  I traced the problem to an attempt to access location 0x00000000 in the function ExXShutdownDLL().  Not calling that function fixes the problem, and I do not see any downside (other than that it feels wrong not to do so;). I pushed the change to svn (which will propagate to githb soon).   Please update your copy of the code from the svn repo, and feel free to provide a better fix.

Issue #1094, opened 2021-01-30
Invalid OME-XML produced by MicroManager


JacksonMaxfield:
Hello all!

I maintain a project called [aicsimageio](https://github.com/AllenCellModeling/aicsimageio) which for the context of this issue simply is: "a pure python library for microscopy image file format reading."

We have an `OmeTiffReader` and in the process of upgrading to our new 4.0 API we made the decision to adopt a library that specifically validates that any XML provided in is _valid_ OME XML (to the XSD).

We had an issue posted earlier today that tested a micromanager produced OME-TIFF and saw that it resulted in an `XMLSchemaValidationError`. Full issue here: https://github.com/AllenCellModeling/aicsimageio/issues/195

The shorthand, is that it looks like the `InstrumentID` in OME-XML produced by micromanager isn't valid to the OME XSD.
I tracked down what I think is the [single line culprit](https://github.com/micro-manager/micro-manager/blob/bc3f1d42ca3f18d721b7c268064436600bb3ea8c/mmstudio/src/main/java/org/micromanager/data/internal/multipagetiff/OMEMetadata.java#L175).

I _think_ this should be:
```java
String instrumentID = MetadataTools.createLSID("Instrument", 0);
``` 

as specified in the [bioformats API](https://downloads.openmicroscopy.org/bio-formats/5.0.6/api/loci/formats/MetadataTools.html#createLSID(java.lang.String,%20int...)).

Sidenote, there is only one other reference to `createLSID` and it is used for the [camera](https://github.com/micro-manager/micro-manager/blob/bc3f1d42ca3f18d721b7c268064436600bb3ea8c/mmstudio/src/main/java/org/micromanager/data/internal/multipagetiff/OMEMetadata.java#L394) I am unsure if that may need to be updated too.

It would be great to support micromanager OME-TIFFs in `aicsimageio` so hoping this spec compliance can be resolved :slightly_smiling_face:

nicost:
Thanks for looking into this @JacksonMaxfield !  I can't say that I understand the OME documentation very much (what is an LSID, what is an Instrument, and why should it have an ID with one or more indices, but the API allows for zero indices, which then generates and invalid data structure? Weird...), but made some changes as you suggested.  Can you have a quick look, and if you think this looks OK, I can push it into the main branch and it will appear in the next nightly build.  Completely agree that it will be nice if this is correct, and if there are more suggestions to make the OME-XML that the MM code generates more useful, I'll be happy to incorporate them.

[...]

PR #1095, opened 2021-01-30
Studio: attempt to create valid OME-XML. Suggested by Jackson Maxfield Brown

nicost:
Closes issue #1094

### DIFF
--- a/mmstudio/src/main/java/org/micromanager/data/internal/multipagetiff/OMEMetadata.java
+++ b/mmstudio/src/main/java/org/micromanager/data/internal/multipagetiff/OMEMetadata.java
@@ -172,7 +172,7 @@ public final class OMEMetadata {
       }
       metadata_.setStageLabelName(positionName, seriesIndex);
 
-      String instrumentID = MetadataTools.createLSID("Microscope", 0);
+      String instrumentID = MetadataTools.createLSID("Instrument", 0);
       metadata_.setInstrumentID(instrumentID, 0);
       // link Instrument and Image
       metadata_.setImageInstrumentRef(instrumentID, seriesIndex);
@@ -391,7 +391,7 @@ public final class OMEMetadata {
 
       for (int i = 0; i < cameras.size(); i++) {
          String camera = cameras.get(i);
-         String detectorID = MetadataTools.createLSID(camera, i);
+         String detectorID = MetadataTools.createLSID("Detector", i);
 
          //Instrument index, detector index
          metadata_.setDetectorID(detectorID, 0, i);
--- a/mmstudio/src/main/java/org/micromanager/data/internal/multipagetiff/OMEMetadata.java
+++ b/mmstudio/src/main/java/org/micromanager/data/internal/multipagetiff/OMEMetadata.java
@@ -172,7 +172,7 @@ public final class OMEMetadata {
       }
       metadata_.setStageLabelName(positionName, seriesIndex);
 
-      String instrumentID = MetadataTools.createLSID("Microscope");
+      String instrumentID = MetadataTools.createLSID("Microscope", 0);
       metadata_.setInstrumentID(instrumentID, 0);
       // link Instrument and Image
       metadata_.setImageInstrumentRef(instrumentID, seriesIndex);
@@ -391,7 +391,7 @@ public final class OMEMetadata {
 
       for (int i = 0; i < cameras.size(); i++) {
          String camera = cameras.get(i);
-         String detectorID = MetadataTools.createLSID(camera);
+         String detectorID = MetadataTools.createLSID(camera, i);
 
          //Instrument index, detector index
          metadata_.setDetectorID(detectorID, 0, i);


mmstudio/src/main/java/org/micromanager/data/internal/multipagetiff/OMEMetadata.java
```diff
@@ -172,7 +172,7 @@ private void startSeriesMetadata(int seriesIndex, String baseFileName) {
       }
       metadata_.setStageLabelName(positionName, seriesIndex);
 
-      String instrumentID = MetadataTools.createLSID("Microscope");
+      String instrumentID = MetadataTools.createLSID("Microscope", 0);
```
evamaxfield: I think this specifically needs to be "Instrument" instead of "Microscope"

mmstudio/src/main/java/org/micromanager/data/internal/multipagetiff/OMEMetadata.java
```diff
@@ -391,7 +391,7 @@ private void setOMEDetectorMetadata(Metadata metadata) throws JSONException {
 
       for (int i = 0; i < cameras.size(); i++) {
          String camera = cameras.get(i);
-         String detectorID = MetadataTools.createLSID(camera);
+         String detectorID = MetadataTools.createLSID(camera, i);
```
evamaxfield: Does this get saved out as "Camera:{i}" or what is the actual string attached to this?

evamaxfield: I ask because I think this also need to be "Detector:{i}"

JacksonMaxfield (commented):
Any chance you could generate a fake OME TIFF? I can run the validation and see if it passes.

mmstudio/src/main/java/org/micromanager/data/internal/multipagetiff/OMEMetadata.java
nicost: Is there any documentation on keywords that are essential?  OME metadata have been opaque/baroque/verbose/incomprehensible to me for the last 20 years;)

mmstudio/src/main/java/org/micromanager/data/internal/multipagetiff/OMEMetadata.java
nicost: No, this results in "WhateverNameTheEndUserHaveToThisCamer:{i}". 
I am a bit confused why there are no specific functions to generate these IDs if it is required to provide a given keyword.  This now puts the onus on the poor user of this API to know what keyword is required to generate valid metadata.  Sorry, venting... But it would be nice if you could point me to human understandable documentation.
Happy to change to whatever you think is required.  I'll change according to your suggestions and make a test set available.

mmstudio/src/main/java/org/micromanager/data/internal/multipagetiff/OMEMetadata.java
evamaxfield: I usually just use the XSD: https://www.openmicroscopy.org/Schemas/Documentation/Generated/OME-2016-06/ome_xsd.html#Instrument_ID

The "Pattern" is a regex match: `((urn:lsid:([\w\-\.]+\.[\w\-\.]+)+:\S+:\S+)|(\S+:\S+)) & ((urn:lsid:([\w\-\.]+\.[\w\-\.]+)+:Instrument:\S+)|(Instrument:\S+))`

Relevant bit here is "Instrument:{any-non-white-space-character}"

mmstudio/src/main/java/org/micromanager/data/internal/multipagetiff/OMEMetadata.java
evamaxfield: Same comment as above. Here is the "Dectector ID" Reference: https://www.openmicroscopy.org/Schemas/Documentation/Generated/OME-2016-06/ome_xsd.html#DetectorID

Hah no need to apologize. You are talking to a person who had to write an [entire function to clean up old OME XML mistakes](https://github.com/AllenCellModeling/aicsimageio/blob/main/aicsimageio/metadata/utils.py#L59) that were written into our version 3 library -- I understand the gripes of metadata confusion. I am just glad that we started validating our metadata now 😂

nicost: DataSet1 (using Microscope and cameraName): https://valelab4.ucsf.edu/~nstuurman/images/MMDataSet1/
DataSet2 (using Instrument and Detector): https://valelab4.ucsf.edu/~nstuurman/images/MMDataSet2/
Curious to hear what you find.  And thanks for your efforts!

evamaxfield: Caught me at just the right moment!

Tested both! Dataset2 (using Instrument and Detector) parses as valid OME and read in just fine in `aicsimageio`.

So I am going to say that does the trick. I also applied the change to a branch of mine as functionality to read the previously generated files: [my cleaning function here](https://github.com/AllenCellModeling/aicsimageio/blob/bugfix/catch-all-errors-in-ome-tiff/aicsimageio/metadata/utils.py#L145)

So both forward and backwards compatibility looks good to me! :tada:

evamaxfield: Oh woops. Forgot to say, I tried DataSet1 (using Microscope:{i} and Camera:{i}) and it failed to parse as OME. So the Dataset2 route is definitely the way to go.

JacksonMaxfield (approved):
Thanks so much!

nicost:
Thanks for all your work on this @JacksonMaxfield !

PR #1098, opened 2021-02-04
Clean OughtaFocus

nanthony21:
This PR makes no changes to the functionality of the OughtaFocus plugin but refactors it to make it more readable, better documented, and to pave the way for future improvements.

Rather than having a single class that is >1000 lines long the plugin has now been split up into 4 classes which each have a single responsibility:
 - `OughtaFocus`: Implements the `AutofocusPlugin` interface.
 - `internal.oughtafocus.ImgSharpnessAnalysis`: Provides several static methods for evaluating the focus of an `ImageProcessor`. The  `Method` enumerator and the `setMethod` and `compute` instance methods can be used to conveniently switch computation methods.
 - `internal.oughtafocus.BrentFocuser`: Controls a Z-Stage and a Camera to optimize focus position using the Brent Method. The function used to evaluate the sharpness of each image is provided to the constructor. The algorithm will attempt to maximize the value of this function.
 - `internal.oughtafocus.FHT_NoScaling`: This class was previously an internal class to `OughtaFocus`. It has been moved to its own file for improved readability of the code. It is used for the FFT mode of `ImgSharpnessAnalysis`.

PR #1099, opened 2021-02-04
Fix oughtafocus build

nanthony21:
Fixes a build error with PR  #1098.

Just needed to add the new source directory to the build.xml


nicost:
Thanks!

PR #1110, opened 2021-02-16
Move examples to common folder

nanthony21:
This PR move the `scripts` folder and the `Tutorial` folder under a common `examples` folder. It also renames the folders and adds a short readme.txt to explain the contents.


nicost:
This change will also need changes in the build.  Specifically, in the ant build (windows), the variable mm.scripts.srcdir will need to be changed (I am not sure where this one is set), or possible the file buildscripts/installer_Windows.iss. I am not exactly sure how these are copied into the Mac build, but they are...

It would certainly be simpler at this point in time to delete the Tutorial directory, and keep scripts where it is, so that we don't have to figure out the scripts generating the installers;)

nanthony21:
Makes sense. I've reverted the original commit and followed your suggestion of simply deleting `Tutorial`.

PR #1113, opened 2021-02-17
Clean folder structure WIP: Needs testing on mac build.

nanthony21:
This PR makes 3 changes to help clean up the folder structure.
- Remove the `TestDeviceAdapters` folder which doesn't appear to be used anywhere.
- Remove the `drivers` folder containing files that should be sourced elsewhere.
- Move the `MacInstaller` folder from the main directory to the `buildscripts` folder. Needs test on mac build.


nicost:
Is the idea to move the drivers folder to the 3rdparty repo?  I don't want them to get lost, i.e. they are still useful to some people.  Also will need to check the links to them from their documentation pages on micro-manager.org.

nanthony21:
I wasn't sure what the best place to move the drivers to is. I think it would be best for the "3rdPartyPublic" repository to only contain things that are needed to build the project.

Is it possible to host these on the website without having them in a source repository?

With a quick google search I wasn't able to find a mention of these drivers on micro-manager.org

nicost:
The drivers are referenced from the device pages on micro-manager.org (for instance: https://micro-manager.org/wiki/K8055).  I see that they are distributed as zip files.  However, I would not want to loose the ability to re-generate the drivers by loosing track of these files.  They have no use outside of Micro-Manager (i.e., they are intimately associated with the respective device adapters).

nanthony21:
I see. For now I will remove the deletion of `drivers` from this PR. The best thing to do with them can be figured out later.

I don't have a way of testing the build on Mac. Would it be possible to have this PR tested on Mac so we know that the move of `MacInstaller` hasn't caused an issue.

nicost:
Agreed.  Lots of historical artifacts on the way (you should see my basement;).  We'll see what the Mac build machine comes up with.

nicost:
Opened an issue for the drivers folder in case we'll forget: #1115

PR #1116, opened 2021-02-17
Remove portdebian folder

nanthony21:
This PR removes the `portdebian` folder which doesn't appear to be used.


marktsuchida:
Looks good to me. If/when an updated debian package is created (and maintained) in the future, it should be its own project (repo) that uses `micro-manager` as a dependency.
There is no reason to keep this folder in the current tree, as we can always access it in the history.

PR #1117, opened 2021-02-18
Allow building on newer version of Visual Studio

nanthony21:
After Visual Studio 2010 the location of the .bat file used to activate the Visual Studio build environment changed.

This PR adds support to the Ant build scripts for the new way to locate and run the .bat file.

Now the build script will use `VSWhere.exe` to identify the most recent version of VisualStudio that is found.

If this fails then the script will fall back to the old method of finding the VS2010 .bat file.

I have tested using `ant stage -Dmm.build.failonerror` to build using VS2019 and everything appears to have worked. I have not been able to make sure that this script still works properly on computers that only have VS2010 installed.


nicost:
It may be easiest to test this on the build machine;)  I'll merge the commit and see if the build machine complains (in which case I may revert). Only downside I see is that people with both VS versions installed now will all of a sudden build with another VS, but that is unlikely to be a big deal.

nicost:
It may be easiest to test this on the build machine;)  I'll merge the commit and see if the build machine complains (in which case I may revert). Only downside I see is that people with both VS versions installed now will all of a sudden build with another VS, but that is unlikely to be a big deal.

nicost:
First test result: It uses VS2010 on my laptop, although it also has VS2015 installed.  Maybe VSWhere was not yet included in VS2015? In any case, seems safe for VS2010 only environments. Thanks @nanthony21!

PR #1121, opened 2021-02-21
fix references to 3rdparty public

nanthony21:
This PR fixes various parts of the codebase that assume the 3rdPartyPublic repository is located at "../3rdPartyPublic" rather than using the Ant and VS environment variables.

This will make is easier to rename or move the public repository in the future.


nanthony21:
This will be addressed in a larger  future PR

PR #1124, opened 2021-02-25
UserProfile management bug fix

### LINKED ISSUE #1123: Multiple instances of the same UserProfile can be created resulting in settings not being properly saved to file.
This is most obvious with the UserProfile `READ_ONLY` setting which is set in the intro-dialog. Since the `UserProfileManager` stores an instance of the the profile in memory as soon as a new profile is selected, it doesn't realize that the underlying file has been updated when `UserProfileAdmin` sets the read only property.

`UserProfileManager` is deprecated anyway, so the best solution is probably to go ahead and remove this deprecated class.

nanthony21:
This PR fixes #1123 

This should also help to improve the readability of UserProfile related code (A lot more could be still be done)

The functionality of the deprecated `UserProfileManager` class has been rolled into the `UserProfileAdmin`. Having only one class that is responsible for managing the context of which user profile is selected helps to prevent confusing behavior and make the code easier to understand.

In addition a section of MMStudio that tried to close the UserProfile twice has been fixed.


nicost:
Thanks Nick!

PR #1132, opened 2021-03-05
Sharpness Inspector Plugin

nanthony21:
This PR adds an inspector plugin called "Image Sharpness" which provides a real-time plot of image sharpness within an ROI. The x axis can be set to `Time` or to `Z` in which case the current Z position of the default Z stage is used. This only works well if the ZStage provides consistent updates of it's position.

In addition to real-time analysis of the image you can use the `Scan` button to sweep through a range of Z positions and plot the sharpness as a function of Z.

The methods of evaluating sharpness are the same as those used by the OughtaFocus Plugin. In order to facilitate this, the sharpness analysis code previously found in `org.micromanager.autofocus.internal.oughtafocus` is now in the libraries folder and ends up in `org.micromanager.imageprocessing`.

Known issue: The axis labels and tick marks of the plot are black and cannot be seen in the `Night` color scheme.


nicost:
Thanks Nick!

PR #1139, opened 2021-03-07
mmCoreAndDevices submodule

nanthony21:
This PR removes all C++ components of the Micro-Manager project from this repository and instead includes this [repository](https://github.com/micro-manager/mmCoreAndDevices) as a submodule.

#### Primary Motivations:
 - The C++ components of the project are in no way dependent on the other components and are sufficiently complicated that it makes sense for them to be managed in their own isolated repository. Projects which are dependent on the C++ components can then update to a new stable version of the `mmCoreAndDevices` repository as needed.
 - Projects which want to use the C++ components but have no interest in the Java and Clojure components are currently required to clone all of that extra content anyway and then use the Java-based Ant build system to compile. The C++ components should have their own independent C++ focused build system which can then be called upon by Ant, Python, another C++ focused build system, etc.
 - The current flat directory structure of the repository makes it difficult to understand for people new to the project. This PR will lead to a more hierarchical structure where independent components have their own subdirectories.

#### Other notable improvements:
 - Adds a build variable to `MMCommon.props` that refers to the location of the `3rdParty` repository of dependencies that cannot be shared publicly.
 - All visual studio projects and C++ source code that assume a fixed relative path to the MMDevice, 3rdPartyPublic, or 3rdParty directories are replaced by references to the build variables in the `MMCommon.props` files. This means that if these relative paths need to be changed in the future there will only be one file to change rather than hundreds.

Things to be done before merging this:
 - [x] Get the Mac and Linux Autotools build system working.
 - [x] Make sure the `mmCoreAndDevices` repository is up to date with any recent C++ changes made in the `Micro-Manager` repository.

#### Instructions to test this PR:
This PR makes use of Git [submodules](https://git-scm.com/book/en/v2/Git-Tools-Submodules). After cloning the repository or checking out the branch as you normally would you will need to take an additional step to make sure that the submodule is properly configured in your local copy. The easiest way to do this that will cover all bases is with the command: `git submodule update --init --recursive`. Please refer to the [CMMCore](https://github.com/micro-manager/CMMCore) repository for additional information. If using Git Bash you can simply do: `cd mmCoreAndDevices` and Git will automatically recognize that it is working in the "mmCoreAndDevices" repository rather than the "Micro-Manager" repository, you can then switch branches, check out a different commit, etc.

#### Instructions get get the source code once this PR is merged.

1. Clone the repository: `git clone https://github.com/micro-manager/micro-manager.git`
2. Move Git bash into the repository: `cd micro-manager`
3. Make sure any submodules are also cloned: `git submodule update --init --recursive`

If you are working only with the source code that is publicly available then that's all. In order to get the private source code continue:

4. Move Git bash into the `mmCoreAndDevices` submodule: `cd mmCoreAndDevices`
5. Change to the `privateMain` branch: `git checkout privateMain`
6. `privateMain` has a submodule that `main` does not. Make sure that any new submodules are in a consistent state: `git submodule update --init --recursive`

### Building
While some aspects of the underlying build scripts have been modified, the process that a developer will use to build the entire project remains unchanged. This is true for the Windows Ant script as well as the Autotools scripts used on Linux and Mac.


nanthony21:
@marktsuchida could I get your thoughts on this? I think it would be a really good direction to take things and would help move us towards other improvements of the build system. One major thing I'm not sure about is how it will work with the current Git-SVN synchronization.

marktsuchida:
Thanks for working on this, @nanthony21 -- it's definitely the direction things need to go in.

I know this is a complex task, but here are some thoughts that immediately came to mind:

- I suspect you will find that making the autotools build work with this (in a non-kludgy way) is quite complex. One idea (not saying it's the only way) would be to make it a more progressive move by breaking up the migration.
    - For example, `MMDevice` is a simple library that has no dependencies (inside or outside of `micro-manager`). If you were to extract just `MMDevice` as its own git repo, it should be fairly straightforward. This can be cleaned up, and (importantly) versioned releases can be made. Once that is done, `MMCore` and every device adapter can be converted to depend on an externally built `MMDevice` (which will be linked like any other normal library -- no need for it to be a submodule). Even if we were to stop after this step, there is already a huge benefit: developers of new device adapters can create a simple project that just depends on the independent `MMDevice` -- which is a lot easier to understand than the current situation.
    - Of course this would mean that the `MMDevice` project would need to be built separately from `micro-manager`, just like `Boost` currently needs to be built before building `micro-manager`. But this can be dealt with (as I previously mentioned) by having a separate repository (perhaps `micro-manager-distro`) whose job is to build all dependencies, followed by `micro-manager`.
    - The other projects, such as `MMCore` and `MMCoreJ`, can be progressively separated out in a similar manner.
    - Because the only connection between the C++ modules and MMStudio is `MMCoreJ`, `MMCoreJ` could then be published as a Maven package (even without the DLLs). Then the remaining `micro-manager` project, which is by then Java-only, simply needs to add `MMCoreJ` to its `ivy.xml` (better yet, migrate to an official `pom.xml`), which is much simpler than trying to build the C++ modules as git submodules.
    - Perhaps what I'm trying to propose here is not so much the order of splitting repositories, but to try to avoid using git submodules to express module dependencies. Git submodules are a good workaround for situations where your dependencies are not clearly versioned, or are somehow overly coupled to your main code, or require local patching that cannot go into the upstream project. But we don't have those problems with `MMDevice` and `MMCore`: both have carefully managed version numbers (70 and 10.1.1 currently) that would work well if used to create releases. By consuming these modules as external dependencies rather than "vendored" submodules, the build system also gets much simpler, because we no longer need a single build system that deals with multiple languages. Matching the desired `MMDevice`, `MMCore`, and `MMStudio` versions, and building them all in one go, can be easily accomplished in the `micro-manager-distro` project.
    - If the idea of `micro-manager` becoming something that no longer builds a full micro-manager distribution bothers you, then the idea of a separate `micro-manager-distro` repo could be scrapped in favor of retaining and enhancing the scripts in `buildscripts/nightly`, which currently have exactly the same role (Boost on macOS, for example, is built by a script there, and the same could be done on Windows). Eventually `MMStudio` and its plugins may also graduate to separate git repos (they will be clean Maven projects with no autotools or Ant mess; how nice would that be!), leaving _only_ those distribution build scripts in `micro-manager`.
    - I didn't talk about the device adapters above -- they could be placed in their own (single) repo, or in individual repos. I think individual repos makes more sense in the longer term (and is required for closed-source device adapters), but that requires having standards in place (and automatically enforced) for how they are organized, and it will still make sense to have an umbrella repo that has all device adapters as submodules, so that crosscutting changes can be made. (Creating releases for each device adapter would be a nightmare.) For the near term, I think a single repository with all device adapters makes perfect sense.

- Is there a reason for the name `CMMCore` for the C++ repository? (I don't have a better idea, but it could be a little confusing because `CMMCore` is the name of the main class in the `MMCore` library (following, for no good reason, an [archaic practice](https://stackoverflow.com/questions/2292468/class-names-that-start-with-c)), whereas the `CMMCore` repository is the combination of `MMDevice` and `MMCore` libraries plus the device adapters.

- Maybe consider converting this PR to a "draft" rather than naming it "WIP" -- since GitHub has that feature.

nanthony21:
Thanks @marktsuchida 
 - You're right, I recently started on the autotools side of things and have found it extremely frustrating. I like what you propose for the future structure of repositories. For now, I am hoping to just get this to a point where it can be merged without making any more disruptive changes than are absolutely necessary. Primarily because I don't want the current situation where there are multiple repositories hosting the same C++ code to go on for too long. Then the process of creating independent build systems for each component of the project can begin.
 - Yes, the naming was just the first thing that came to mind. I've thought about it a bit and couldn't come up with anything very good. What do you think about `mmCoreAndDevices`. This would be the repository name as well as the name of the folder when the Micro-Manager repository is checked out.

nanthony21:
This PR is now working on Windows and Linux. The main remaining task is to test it on Mac. If anyone who has access to a Mac would be willing to give it a try I would be appreciative.

nanthony21:
My understanding is that this has now been successfully tested on Mac. I have also updated the C++ repository to match the current status of the SVN repository.

PR #1143, opened 2021-03-09
Fix oughta focus build

nanthony21:
This PR fixes an issue where Oughta Focus plugin was not building as a result using a file structure that is not supported by the autofocus Ant build scripts.

In addition the "libraries" are now included in the autofocus build scripts so that autofocus plugins can reference code in these libraries.


nanthony21:
Note: this reversal of a precious change to the ouggtafocus plugin was necessary because the autofocus plugin buildscript assumes that each plugin consists of only a single file under the main source directory. 

It might make sense to at some point move all of the autofocus plugins to the "plugins" directory. I don't see a reason why they need a separate build process from the other plugins. This would allow for greater flexibility in how autofocus plugins can be structured.

PR #1144, opened 2021-03-09
Speed up image metadata processing

tomhanak:
Replaced stream operations by `std::string` concatenation, avoided unnecessary new instances and clones of `MetadataTag` classes.
Based on measurements during live acquisition this reduces `Metadata::Serialize`+`Restore` time to a half.

I'm not sure how exactly `std::stream` separates strings and numbers. This patch works with strings only and ensures each part is followed by new line. Deserialization then uses `std::readline` for all. I hope this doesn't break anything on java side.

### DIFF
--- a/MMDevice/ImageMetadata.h
+++ b/MMDevice/ImageMetadata.h
@@ -95,13 +95,13 @@ public:
    const std::string& GetName() const {return name_;}
    const std::string GetQualifiedName() const
    {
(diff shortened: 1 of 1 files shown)


nicost:
I tested with the DemoCamera, and metadata looks alright (but hard to tell if details changed).  Pushing it into svn, and a bit later today into github.  Thanks!

PR #1157, opened 2021-03-20
Fix pymmcore error messages

tlambert03:
This fixes https://github.com/micro-manager/pymmcore/issues/15, wherein the message from `CMMCore` exceptions are not making it to the python exception (the code [here](https://github.com/micro-manager/micro-manager/blob/e6203c39e7ac9fd5f94a760cccc2c042115fedb3/MMCorePy_wrap/MMCorePy.i#L223-L255) is supposed to do this, but it doesn't work).  I also made it so that various error codes map to the corresponding python error type (instead of always being a `RuntimeError`).

One issue I ran into here is the use exception specifications in function signatures (with the `throw` keyword, like `bool CMMCore::deviceBusy(const char* label) throw (CMMError)` ... turns out swig [handles this specially](http://www.swig.org/Doc1.3/Typemaps.html#throws_typemap), and the [standard `%exception` directive](http://www.swig.org/Doc4.0/SWIGDocumentation.html#Customization_exception) will not work here.

cc @nicost @marktsuchida

### DIFF
--- a/MMCorePy_wrap/MMCorePy.i
+++ b/MMCorePy_wrap/MMCorePy.i
@@ -246,13 +246,13 @@ PyObject *setSLMImage_pywrap(const char* slmLabel, char *pixels, int receivedLen
         case MMERR_InvalidStageDevice:
         case MMERR_InvalidStateDevice:
         case MMERR_InvalidXYStageDevice:
+        case MMERR_NoConfigGroup:
+        case MMERR_NoConfiguration:
         case MMERR_PropertyNotInCache:
         case MMERR_SetPropertyFailed:
         case MMERR_UnexpectedDevice:
         case MMERR_UnknownModule:
             SWIG_exception(SWIG_ValueError, e.what());
-        case MMERR_NoConfigGroup:
-        case MMERR_NoConfiguration:
         case MMERR_NullPointerException:
             SWIG_exception(SWIG_NullReferenceError, e.what());
         case MMERR_FileOpenFailed:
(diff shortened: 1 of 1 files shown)


marktsuchida:
Cool! Mostly looks good to me, except that I think some of the mappings of MM error codes to the Python exception types are not appropriate. For example, `MMERR_NoConfigGroup` should be `ValueError`, and so forth. One would need to study each error code (which, unfortunately, is only defined by its current usage) and the intended use for each Python exception type to get it right.

Or, they could all be made to raise a vanilla `Exception` for now. If a given error raises `Exception` now it can later be refined to raise `ValueError`, but if it raises `ValueError` now we cannot later change it to `KeyError` without breaking user code that tries to catch `ValueError`.

tlambert03:
Yep, I figured you might want to discuss the mappings.  I did my best to look through each error code, figuring you'd have suggestions if you wanted to change them.  I'm happy to change anything.  Did you want to tell me others you want changed? should I just change `MMERR_NoConfigGroup` and `MMERR_NoConfiguration` to `ValueError`?  or would you prefer just using `RuntimeError` for everything (the swig default).

tlambert03:
@marktsuchida, I changed the two config errors.  Please let me know if you want to use the current mappings, suggest additional changes, or use RuntimeError for everything.

tlambert03:
> Or, they could all be made to raise a vanilla Exception for now.

(it's not clear to me which of the `SWIG_ExceptionTypes` will convert to the base python `Exception`. let me know if you know how to achieve that.

marktsuchida:
Thanks @tlambert03 and sorry for the delay. I might wait for #1139 to be merged and apply your changes afterwards, but will get to this soon.

nicost:
@marktsuchida, since #1139 was merged a while back, is it useful to look into this now?

nicost:
@marktsuchida can you please, please, pretty please look into this?

marktsuchida:
Moved to micro-manager/mmCoreAndDevices#9. Apologies for the long delay.

Issue #1017, opened 2020-11-18
Coords axes policy needs clarification


marktsuchida:
- When (if ever) do zeros get added for TPZC; can we stop doing so?
- Where are non-TPZC axes allowed (everywhere except file-backed Datastores?)
- Are axes ordered or not? Are comparison operations etc. well-defined and self-consistent?
- How are missing axes treated in `DataProvider` and `Datastore` methods?

See also #619.

nicost:
> When (if ever) do zeros get added for TPZC; can we stop doing so?

Not sure what this means.  For each and every instance of Coords all four prime axes need to be set to 0 by the calling code, or bad things will happen (such as not displaying the image with that Coord when added to a Datastore) as described in #619.  Not adding zeroes is currently not an option, but it would be great if the code would be fixed so that it would be possible.  What is the planned approach to improve on the current code?

marktsuchida:
Precisely. It is clear that in the long term, everything should just work without having to manually add zeroes. For that, the issues in #619 (and possibly others) need to be solved. If that is not easy in the short term (and therefore we need to keep the rule of users always having to add the 4 axes), at the very least we need a transition plan so that the restriction can be lifted in the future without breaking code written under the current restrictions. So either fixing #619 (very unlikely) or clearly documenting the current rules and future plans is what this issue is about.

This may not just be about the datastore code, but _might_ potentially involve the mathematical model of what a `Coords` is: for example, the definition of equality between two `Coords` values. Or maybe we need to introduce a new comparison operation ("are these two `Coords` equal given these particular axes?"). Alternatively, it might be as simple as documenting limitations of `Datastore` (rather than `Coords` in general). I intend to find the simplest future-proof thing we can do for the moment, but leaving these issues vague seems like a bad idea.

nicost:
Great!  I like the idea of introducing a comparison operator.  Maybe it is as simple as dropping all zero axes, and only then testing for equality.

marktsuchida:
The remaining issue has been restated as #1210.

PR #1178, opened 2021-04-05
Coords cleanup

### LINKED ISSUE #619: Confusing behavior when adding new images to Datastores
When constructing a Datastore that only varies along a single axis (e.g. a single position, single single slice, single channel, time lapse), the Datastore will allow you to add images with Coords that only have one axis specified, but it won't correctly display them. This is because the HashMap that stores images based on their Coords does not consider Z0, P0, C0, T1 equal to Coords with only T1. So the following code won't work:

  Coords.CoordsBuilder builder = studio.data().getCoordsBuilder();
  builder.time(1);
  datastore_.putImage(images.get(0).copyAtCoords(builder.build()));

But this code will:

  Coords.CoordsBuilder builder = studio.data().getCoordsBuilder();
  builder.time(1);
  builder.channel(0);
  builder.stagePosition(0);
  builder.z(0);
  datastore_.putImage(images.get(0).copyAtCoords(builder.build()));

Doesn't seem like this behavior is incorrect, but perhaps Datastore.putImage could more strictly enforce that the Coords it is passed be fully specified?

nicost:
This PR deals with the long standing issue that Coords with an axis set to zero are not identical to Coords that lack the axis altogether.  I know that Mark though differently about this, but to me, this made no logical sense for a system that is supposed to handle arbitrary axes.  The CoordsBuilder now drops all axes that were set to zero, asking for the value of a missing axis in a Coord will return zero (rather than -1), and the code was updated to work with this new scheme.

In addition, missing images are again filled in.  It is still desirable to add a user interface so that the user can switch this feature on or off. 

Closes #619 and addresses part of #1017


nanthony21:
I think this looks pretty good. I added a few comments for things that I wasn't sure about. I have a plugin I use where I remember running into something similar to #619 . I'll plan to give is a try tomorrow and see if this PR allows me to simplify the `Coord` handling of that plugin.

nanthony21:
This appears to work fine for me which includes testing a plugin that opens and adds images to its own display windows.

One change I notice is that if I "Snap" an image using the demo configuration it is now represented in the histogram as if it were a multichannel image. For example, rather than the histogram being colored white with a label that says "default", it is colored orange and has a label of "DAPI".

However, the display windows opened by my plugin still are shown in the histogram using white and "default".

nicost:
The idea is that when you have the ChannelGroup set to Channel, and the Channel Preset to DAPI, and the Inspector Windows to Composite mode, that the image is shown in the color of that channel.  When you change the Inspector WIndow mode to Grayscale, it will change to Gray (and that persists through closing and re-opening).   Seems the correct behavior to me (and it was intended to do so before as well, not sure if there was a bug there, plus there are still issues with persistence of display settings).

nanthony21:
What I am seeing is that even though the inspector is set to its default of "Grayscale" and the image is displayed as grayscale, the histogram is being shown in orange as if it were set to "Composite".

I tried switching to composite and then back to grayscale and while the image color changed as expected the histogram stays in composite mode.

It seems like a pretty minor thing but I just notice that it is different from before.

nicost:
I think this must have been similar before, but probably never showed itself.  I guess this should be addresses in the Inspector code (i.e., not color the histogram when display mode is grayscale.  I noticed this too, but not completely sure if the current behavior is nice or annoying;)

marktsuchida:
I actually like the simplicity of equating zeros to not having the axes. I probably didn't advocate for it because I felt that it would be a problem if `Coords` could not be printed/displayed to the user correctly without additional information (i.e. the axes of interest and their order), but maybe it is simpler if we go the extreme of keeping `Coords` minimal and handling the list-of-axes separately. (Maybe there is a place for a `CoordinateSpace` type which would be an _ordered_ list of axes.)

There is no way I can see through all potential problems, but here are a few things that I noticed:

- Speaking of going to the extreme, I think it would reduce the chances of hidden bugs if `DefaultCoords` is modified so that it never stores the axes with 0 upon creation. Then, you no longer need `copyRemovingZeroAxes()`. Even better, the axes could be sorted lexicographically upon creation, so that `equals()` and `hashCode()` don't need to sort. This doesn't just simplify the code, but also completely eliminates a class of possible bugs caused by behavior that is affected by order or zero-vs-absent.

- I don't think `isSuperspaceCoordsOf()` and `isSubspaceCoordsOf()` have any usable meaning now that `Coords` by itself has no idea what space it belongs to. So those methods need to be deleted. (I think you already eliminated at least some of their usage.)

- What does `DataProvider.getMaxIndices()` now return? I think its definition is now problematic because the former `-1` vs `0` distinction is now gone (not sure if it was 100% clean previously either). In theory it could return `null` when the data provider is empty versus return an empty `Coords` when there is just one image (with empty/all-zero `Coords`), though that would probably break a lot of code by causing NPEs. Maybe best to replace it with something like `getNextIndices()` which would return a `Coords` with past-end indices? (For a data provider containing [(T=0), (T=1)] it would return (T=2).)

nicost:
Great comments @marktsuchida!

> Speaking of going to the extreme, I think it would reduce the chances of hidden bugs if DefaultCoords is modified so that it never stores the axes with 0 upon creation. Then, you no longer need copyRemovingZeroAxes(). Even better, the axes could be sorted lexicographically upon creation, so that equals() and hashCode() don't need to sort. This doesn't just simplify the code, but also completely eliminates a class of possible bugs caused by behavior that is affected by order or zero-vs-absent.

Although possibly not obvious, DefaultCoords already never stored aces with 0 (the builder drops them, and objects can only be created through the builder). Sorting in the constructor only is a great idea.  Implemented.  Removed copyRemovingZeroAxes (which was a private function).

> I don't think isSuperspaceCoordsOf() and isSubspaceCoordsOf() have any usable meaning now that Coords by itself has no idea what space it belongs to. So those methods need to be deleted. (I think you already eliminated at least some of their usage.)

Deprecated them and removed their use in our code.  

> What does DataProvider.getMaxIndices() now return? I think its definition is now problematic because the former -1 vs 0 distinction is now gone (not sure if it was 100% clean previously either). In theory it could return null when the data provider is empty versus return an empty Coords when there is just one image (with empty/all-zero Coords), though that would probably break a lot of code by causing NPEs. Maybe best to replace it with something like getNextIndices() which would return a Coords with past-end indices? (For a data provider containing [(T=0), (T=1)] it would return (T=2).)

For a single image, it returns an empty Coords.  Same for a DataProvider that has no images. I am not sure that getNextIndices() would help, because what would it return for a single image (without knowledge of which axes the data set should contain)?  The calling code can figure out what situation is the case by using DataProvider.getNumImages(), but I agree that is not elegant.  We could make a special "NoImage" Coords, but that opens up potential other issues.  Suggestions are welcome.

marktsuchida:
@nicost You are absolutely right about `getNextIndices()` being undefined for the single image case....

I still think it is in our interest to remove `getMaxIndices()`, because it will likely invite people to unwittingly write code that has bugs around edge cases.

I did a _very_ quick grep of the uses of `getMaxIndices()`, and it looks like the majority of them just subsequently get the max index for a particular axis (either hard-coded, or looping over each axis). So at least for those uses, a combination of `Set<String> getCoordsAxes()` (which returns the union of all axes used in the dataset) and `int getNextIndex(String axis)` is sufficient (and simplifies user code!). `getNextIndex()` can return `0` when the axis is not used, and `1` if the axis is used but all images currently stored have a `0` for the axis.

A few of the other uses appeared buggy (e.g. equating the "max indices" to some idea of the "last image").

At least a couple uses (in Duplicator and ZProjector) looked like they needed a proper code review, but probably can be written using `getCoordsAxes` and `getNextIndex`.

For some cases (display window?) it might be useful to have an `Image getImageWithMaxIndices()`, which can legitimately return `null` when the dataset is empty.

By the way, I would suggest just deleting the superspace/subspace methods (rather than deprecating them), because I can't imagine any existing code that uses them (if such code even exists) would continue to work correctly with the new definition of `Coords`.

nicost:
> By the way, I would suggest just deleting the superspace/subspace methods (rather than deprecating them), because I can't imagine any existing code that uses them (if such code even exists) would continue to work correctly with the new definition of Coords.

Done.  I like removing unused functions;).

nicost:
>  did a very quick grep of the uses of getMaxIndices(), and it looks like the majority of them just subsequently get the max index for a particular axis (either hard-coded, or looping over each axis). So at least for those uses, a combination of Set<String> getCoordsAxes() (which returns the union of all axes used in the dataset) and int getNextIndex(String axis) is sufficient (and simplifies user code!). getNextIndex() can return 0 when the axis is not used, and 1 if the axis is used but all images currently stored have a 0 for the axis.

So, do I understand correctly that you propose to deprecate DataProvider.getMaxIndices() and introduce DataProvider.getNextIndex(String axis)?

nicost:
Alright.  Deprecated `DataProvider.getMaxIndices()` and `DataProvider.getAxisLength(String axis)`, the latter directly replaced by `DataProvider.getNextIndex(String axis)`.  I left the Storage Interface unchanged.  Tested (but impossible to test all behaviors).  Seems OK.

Ready to merge?

marktsuchida:
Yeah, I think this is good to merge.

I didn't see the `getAxisLength`, but I agree with your assessment about the naming.

The Storage methods should probably follow the same pattern, but that is less urgent because they are not used directly in user code very often (I hope).

Issue #1012, opened 2020-11-18
Handling of display closing, data saving, acquisition cancellation needs improvement


marktsuchida:
- Existing mechanisms involved in this include the "managed datastore" mechanism and `DataViewerListener`.
- `DataViewerListener` at the very least needs to be renamed (it's not what people know as a listener in Java).
- The logic of determining if a `DataViewer` may close should be moved out of individual `DataViewer` implementations and into `DataManager` or some other appropriate place; otherwise we have many copies of similar code handling "priorities"
- Otherwise, the overall idea of what `DataViewerListener` does is quite valid; only code organization needs to change.

PR #1179, opened 2021-04-06
Builder cleanup

### LINKED ISSUE #1021: All API (non-internal) types must be interfaces, not classes, except in special cases
All exceptions to this must be documented. Non-internal classes whose Javadoc doesn't state a valid reason are bugs.

Some valid exceptions are:
- Certain bag-of-static-methods classes that will always be stateless (e.g. `org.micromanager.data.Coordinates`)
- Exception classes and other types that Java or the JDK requires to be a class
- Abstract classes to be subclassed by user code when implementing SPIs
- Very simple pure data types

Most of the current violations are event classes; see #1018.

nicost:
Closes #1021 with the exception of issue #1012 mentioned in there.

Note this causes binary incompatibility with existing plugins (that expect classes rather than interfaces.  Recompiling those plugins (no code change needed) fixes that.  Alternative (deprecating and renaming these events) would be ugly.

### DIFF
--- a/mmstudio/src/main/java/org/micromanager/acquisition/AcquisitionManager.java
+++ b/mmstudio/src/main/java/org/micromanager/acquisition/AcquisitionManager.java
@@ -35,6 +35,19 @@ import org.micromanager.data.SummaryMetadata;
  * or Studio.getAcquisitionManager().
  */
 public interface AcquisitionManager {
+
+   /**
+    * Provides an empty Sequence Settings Builder
+    *
+    */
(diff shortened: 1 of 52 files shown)


marktsuchida (approved):
I don't have any problem with this, but if you feel uncomfortable with breaking existing processors, it can be left an abstract class. `Processor` is an SPI, which is one of the cases where classes, rather than interfaces, are acceptable, as I wrote in #1021 (although ideally it would have been named `AbstractProcessor`). But I'm also okay with the interface version.

PR #1188, opened 2021-04-14
Fixed integration of Boost.Filesystem.

tomhanak:
The ```ax_boost_filesystem.m4``` has been originally downloaded from https://www.gnu.org/software/autoconf-archive/ax_boost_filesystem.html.

Similar patch has been applied to ```mmCoreAndDevices``` submodule yet where the effect is limited to public device adapters.
Changes in this patch allow integration of Boost.Filesystem in ```MMCore*```.

@nicost Could please move these changes also to SVN?
Precisely:
- copy ```ax_boost_filesystem.m4``` to ```<root>/m4/```
- add ```AX_BOOST_FILESYSTEM``` to:
  - ```<root>/configure.ac```
  - ```<root>/DeviceAdapters/configure.ac```


nicost:
Ported to svn as well. Thanks!

PR #1191, opened 2021-04-15
Bump pycro-manager libraries to add core logging to ZMQ server

henrypinkard:
"AcqEngJ" rev="0.12.1"
"NDViewer" rev="0.5.1"
"PycroManagerJava" rev="0.21.2"

### DIFF
--- a/buildscripts/ivy.xml
+++ b/buildscripts/ivy.xml
@@ -42,16 +42,16 @@
 		<!-- Magellan dependencies -->
 		<!-- https://mvnrepository.com/artifact/org.zeromq/jeromq -->
 		<dependency org="org.zeromq" name="jeromq" rev="0.5.1"/>
-		<dependency org="org.micro-manager.acqengj" name="AcqEngJ" rev="0.12.0">
+		<dependency org="org.micro-manager.acqengj" name="AcqEngJ" rev="0.12.1">
          <exclude org="org.micro-manager.mmcorej" name="MMCoreJ"/>
       </dependency>
-		<dependency org="org.micro-manager.ndviewer" name="NDViewer" rev="0.5.0">
+		<dependency org="org.micro-manager.ndviewer" name="NDViewer" rev="0.5.1">
          <exclude org="org.micro-manager.mmcorej" name="MMCoreJ"/>
       </dependency>
 		<dependency org="org.micro-manager.ndtiffstorage" name="NDTiffStorage" rev="2.4.0">
          <exclude org="org.micro-manager.mmcorej" name="MMCoreJ"/>
       </dependency>
-		<dependency org="org.micro-manager.pycro-manager" name="PycroManagerJava" rev="0.21.0">
+		<dependency org="org.micro-manager.pycro-manager" name="PycroManagerJava" rev="0.21.2">
          <exclude org="org.micro-manager.mmcorej" name="MMCoreJ"/>
       </dependency>
 
--- a/mmstudio/src/main/java/org/micromanager/internal/MMStudio.java
+++ b/mmstudio/src/main/java/org/micromanager/internal/MMStudio.java
@@ -99,6 +99,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.UUID;
+import java.util.function.Consumer;
 import java.util.function.Function;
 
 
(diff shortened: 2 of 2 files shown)


nicost:
After this PR my ID shows an error in MMStudio line 563.  Something to the effect that it was unable to find a constructor for ZMQServer matching this signature.  Strangely, the project builds fine, so it seems mainly the IDE (IntelliJ) that is confused by something.

Weirdly, inspecting PycroManager there are two classes for ZMQServer:
org.micromanager.internal.zwq.ZMQServer
org.micromanager.internal.zmq.ZMQServer(1)
The latter seems to have an internal class also called ZMQServer.  

Unzipping the PycroManager 0.21.2 jar only shows one ZMQServer class, but it is called:
  inflating: org/micromanager/internal/zmq/ZMQServer (1).class
There are two ZMQUtility classes:
  inflating: org/micromanager/internal/zmq/ZMQUtil.class
  inflating: org/micromanager/internal/zmq/ZMQUtil (1).class

It is all a bit confusing.  

As an aside, would it not make sense to at least make starting the ZQM server part of the API, i.e. not be in an internal package?

henrypinkard:
> After this PR my IDE shows an error in MMStudio line 563. Something to the effect that it was unable to find a constructor for ZMQServer matching this signature. Strangely, the project builds fine, so it seems mainly the IDE (IntelliJ) that is confused by something.
>
> Weirdly, inspecting PycroManager there are two classes for ZMQServer:
> org.micromanager.internal.zwq.ZMQServer
> org.micromanager.internal.zmq.ZMQServer(1)
> The latter seems to have an internal class also called ZMQServer.
> 
> Unzipping the PycroManager 0.21.2 jar only shows one ZMQServer class, but it is called:
> inflating: org/micromanager/internal/zmq/ZMQServer (1).class
> There are two ZMQUtility classes:
> inflating: org/micromanager/internal/zmq/ZMQUtil.class
> inflating: org/micromanager/internal/zmq/ZMQUtil (1).class
> 
> It is all a bit confusing.

Weird. I changed the signature of the ZMQServer constructor in this PR, so that it accepts a logger that writes to corelog. Is it possible you don't have up to date copy of pycromanager-java.jar in your IDE? OR maybe duplicate copies (source code and Jar)? this might explain double constructors

There are also two constructors so maybe that explains this?
```
//this one used internally in ZMQ server
public ZMQServer() ;

//This one called by mmstudio
public ZMQServer(Collection<ClassLoader> cls, Function<Class, Object> classMapper,
                    String[] excludePaths, Consumer<String> debugLogger) throws URISyntaxException, UnsupportedEncodingException;
```
    
> As an aside, would it not make sense to at least make starting the ZQM server part of the API, i.e. not be in an internal package?

Maybe. Although you wouldn't be able to start it from Python (since you would need an instance of it to call the API in the first place). What do you think the advantages of doing this would be?

nicost:
> s it possible you don't have up to date copy of pycromanager-java.jar in your IDE? OR maybe duplicate copies (source code and Jar)? this might explain double constructors

I ran ant -f buildscripts/fetchdeps.xml and PycroManager is: PycroManagerJava-0.21.2.jar.  No other PycroManager (source or jar) anywhere near the project...  The decompiler in the IDE sees the 'old' constructor:
`public ZMQServer(Collection<ClassLoader> cls, Function<Class, Object> classMapper, String[] excludePaths);`
Removing PycroManager jar from the project gives errors for all zmq related functions, bringing it back in resolves all except for the error with the constructor.  It could be a bug in the IDE, but it certainly is annoying.

> Maybe. Although you wouldn't be able to start it from Python (since you would need an instance of it to call the API in the first place). What do you think the advantages of doing this would be?

Not sure how the starting from Python is related.  This is more a Java packaging/code organization issue.  It is nice to have functions that are for use by outside packages (such as starting the ZMQ server) available in the api, preferably in the form of an interface (to decouple the implementation and make it easy to change the implementation whenever that is needed).  MMStudio should not have to call internal functions of PycroManager.  So things like constructing the ZMQServer, and closing it, belong in an externally visible api and not as an internal function (outside code should never have to use internal code).

nicost:
The problem is with the two ZMQServer classes present in the jar.  One has the old signature, the other has the new signature.  The IDE picks up the old one, luckily, the build and runtime pick up the new one.  When I build the project myself (using `mvn clean, mvn compile`), I do not see that second ZMQServer class.  I suspect that there were some files cluttering the workspace while building, and that rebuilding (and re-publishing the cleaned up version) of PycroManager.jar will fix the problem.

PR #1194, opened 2021-04-15
Fix dev adapters install step

nanthony21:
This pull request removes an ant script that is no longer used because it has been moved to the `mmCoreAndDevices` submodule.

Many device adapters include a `build.xml` script that references this file by a relative path. It is used to copy additional dependencies to the `stage` folder. Since the `deviceadapters.xml` couldn't be found at that relative path to the device adapters this step was being skipped.

A future build system should be designed such that issues like this stop the build process with an error.

I am also updating our reference to the `mmCoreAndDevice` submodule to the most recent commit on `main`. However, since we should be running `git submodule update` with the `--remote` option before each nightly build, this shouldn't actually matter.


nicost:
It looks like we worked on the same problem but used different approaches.  I just updated the paths in all the build.xml files, so these two approaches surely clash.

nanthony21:
Sorry about that Nico, I should have let you know I was planning to do this. If you already have it working then we can go ahead and close this.

PR #1195, opened 2021-04-15
Make all Builders uniform

### LINKED ISSUE #1019: Builder APIs should be standardized
- The builder for `T` should be a nested interface `T.Builder` with a method `build()`.
- The method of `T` to obtain a copy builder should be named `copyBuilder()`, not `copy()` etc.
- Generic builder factory methods should be named `builder()` (if it is obvious for which type) or `tBuilder()`. Not `getTBuilder()`, `newTBuilder()` etc.
- Changes to conform to this rule should keep existing methods but deprecate them.

nicost:
Closes #1019

Only thing missing is access function for PropertyMaps.Builder().  I am not sure where that function should be.  The only logical place is the Studio interface itself. Suggestions?

PropertyMaps is a bit chaotic with the Builder extending a deprecated interface. Would be nice to clean this all up after the release.

### DIFF
--- a/mmstudio/src/main/java/org/micromanager/Application.java
+++ b/mmstudio/src/main/java/org/micromanager/Application.java
@@ -41,7 +41,7 @@ public interface Application {
     * poll each device in the system for its current state, which may take
     * significant time. Compare refreshGUIFromCache().
     */
-   public void refreshGUI();
+   void refreshGUI();
 
    /**
     * Updates the GUI so that its state reflects the current state of the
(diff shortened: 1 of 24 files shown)


nanthony21:
I know this would be a tremendous amount of work, but would it make sense to remove PropertyMaps altogether? They seem to me like a cumbersome extra layer that could just as easily be handled with JSON.

For example, see [here](https://github.com/micro-manager/pycro-manager/blob/master/pycromanager/utility/PropertyMap.py) for what is needed to make a PropertyMap easy to work with in Python.

nicost:
Going full circle!  These started out as JSON (for instance: https://valelab4.ucsf.edu/~MM/doc/mmstudio/org/micromanager/api/ScriptInterface.html#createAcquisition-org.json.JSONObject-boolean-boolean-), which created all kinds of interesting issues.  I am also not sure that creating our own dataFormat is the best solution, but that is the choice @marktsuchida made back in the day, and now is certainly not the time to change that.  Trying to get 2.0 finished here, following @marktsuchida's design.  Discussion about replacing PropertyMaps should be in its own issue with a milestone later than 2.0;)

nicost:
OK, added a PropertyManager that basically gives access to the static functions in PropertyMaps through an instance.  
@marktsuchida, I will merge this by 20140417 (i.e. today), 4pm unless you report any issues.

marktsuchida:
Please go ahead.

PR #1201, opened 2021-04-21
Throw exception when MultiPageTiff writer encounters data order that it can not save correctly.

### LINKED ISSUE #685: Datastore silently accepts images of different dimensions
Using the putImage function on a datastore and sending images of different sizes does not cause any error whatsoever. This can lead to difficult to trace failures. I think that it would be best for the putImage function to throw an exception.

Illustrated with the following script:

targetStore = mm.data().createRAMDatastore();
mm.displays().createDisplay(targetStore);

nrT = 3;
cb = Coordinates.builder();
cb.c(0).t(0).p(0).z(0);

for (int t =0; t < nrT; t++) {
	cb.t(t);
	imgs = mm.live().snap(true);
	img = imgs.get(0).copyAtCoords(cb.build());
	targetStore.putImage(img);
	xSize = Integer.parseInt(mmc.getProperty("Camera", "OnCameraCCDXSize"));
	xSize--;
	mmc.setProperty("Camera", "OnCameraCCDXSize", xSize);
}

### LINKED ISSUE #716: MultiPageTiffStorage needs data in specific order
The 2.0 API allows for data to be added to a Datastore in arbitrary order.  However, when choosing MultiPageTiffStorage as the Datastore, certain orders of addition of images lead to corrupt datasets.  The Beanshell script below illustrates this problem.  No exception is thrown by executing this code.  However, an Exception is caught many times on line 267 of org.micromanager.data.internal.multipagetiff.OMEMetadata.

Even though the example in this script is extreme, it quite often happens that data are missing, which may lead to data stored on disk without warning the user that reading the data back may be problematic.

```
import org.micromanager.data.Coordinates;

int positions = 7;
int slices = 6;
int tPoints = 4;
int channels = 3;

cb = Coordinates.builder();

cb.c(1).t(1).p(1).z(1);

store = mm.data().createMultipageTIFFDatastore(
      "C:\\tmp\\delme", true, false);

for (int p = positions; p > 0; p--) {
	for (int z = slices; z > 0; z--) {
		for (int t = tPoints; t > 0; t--) {
			for (int c = channels; c > 0; c--)  {
/*  this order works correctly
for (int p = 0; p < positions; p++) {
	for (int z = 0; z < slices; z++) {
		for (int t = 0; t < tPoints; t++) {
			for (int c = 0; c < channels; c++) {
*/
				cb.c(c).t(t).p(p).z(z);
				imgs = mm.live().snap(false);
				store.putImage(imgs.get(0).copyAtCoords(cb.build()));
			}
		}
	}
}

store.freeze();
mm.scripter().message("Done");
```

nicost:
Closes #716

### DIFF
--- a/mmstudio/src/main/java/org/micromanager/data/internal/multipagetiff/FileSet.java
+++ b/mmstudio/src/main/java/org/micromanager/data/internal/multipagetiff/FileSet.java
@@ -182,8 +182,12 @@ class FileSet {
          currentFrame_ = frame;
 
          omeMetadata_.addImageTagsToOME(img.getCoords(), img.getMetadata(),
-               ifdCount_, baseFilename_, currentTiffFilename_, currentTiffUUID_);
+                 ifdCount_, baseFilename_, currentTiffFilename_, currentTiffUUID_);
+      } catch (UnsupportedOperationException uoe) {
+         throw (uoe);
       } catch (Exception ex) {
+         // not sure why we continue when this fails, since there is also tiff related
+         // stuff that happens in the OME metadata code.
          ReportingUtils.logError(ex, "Problem writing OME metadata");
       }
 
--- a/mmstudio/src/main/java/org/micromanager/data/internal/multipagetiff/OMEMetadata.java
+++ b/mmstudio/src/main/java/org/micromanager/data/internal/multipagetiff/OMEMetadata.java
@@ -308,12 +308,17 @@ public final class OMEMetadata {
       int channel = coords.getChannel();
 
       // ifdCount is 0 when a new file started, tiff data plane count is 0 at a new position
-      metadata_.setTiffDataFirstZ(new NonNegativeInteger(slice), position, indices.tiffDataIndex_);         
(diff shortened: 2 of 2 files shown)


mmstudio/src/main/java/org/micromanager/data/internal/multipagetiff/FileSet.java
```diff
@@ -182,8 +182,12 @@ public void writeImage(final Image imgIn) throws IOException {
...
 
          omeMetadata_.addImageTagsToOME(img.getCoords(), img.getMetadata(),
-               ifdCount_, baseFilename_, currentTiffFilename_, currentTiffUUID_);
+                 ifdCount_, baseFilename_, currentTiffFilename_, currentTiffUUID_);
+      } catch (UnsupportedOperationException uoe) {
+         throw (uoe);
       } catch (Exception ex) {
+         // not sure why we continue when this fails, since there is also tiff related
+         // stuff that happens in the OME metadata code.
          ReportingUtils.logError(ex, "Problem writing OME metadata");
```
marktsuchida, line 191: My guess is just lazy coding (maybe it was too much work to propagate a checked exception?).

marktsuchida (approved):

PR #1202, opened 2021-04-21
Reformat all code and add GH actions linter

nanthony21:
Since this PR changes almost every line of every single file it will probably cause a lot of problems for existing pull requests and probably shouldn't be merged. But I just wanted to open it to bring up the discussion.

Currently much of the Java codebase uses non-uniform formatting style. There are many places where you can see the indent switching between 3 and 4 spaces. There are also many undocumented methods, variables that should be `final`, etc.

This PR does 2 things:
 - Introduces a GitHub action to run `super-linter` on all new pull requests to `master` as well as all pushes to other branches. Code style of all newly pushed code will be automatically checked and the PR will be flagged if there is an issue.
 - All existing code has been automatically reformatted to match Google's style guide using the "google-java-format" plugin for IntelliJ


nanthony21:
Note, that the linter will fail on this PR. That is because we have a huge amount of old code that is modified by this PR but which still is not fully compliant.

Normal pull requests (which don't modify every file) will only run the linter on those files that have been modified.

nicost:
Where can I find the definition of the style that is used here?  Note that there actually is a style guide: https://micro-manager.org/wiki/Micro-Manager_Coding_Style_and_Conventions.

nanthony21:
Thanks, @nicost. I actually didn't realize that there was already a  style guide defined.

Here is all of the info about the Google style guide. https://google.github.io/styleguide/javaguide.html. I'm actually not very familiar with the style (Personally I've always preferred 4 spaces), but it seems to be the only style that is supported by "Super-Linter" and it also seems to be the closest thing there is to a widely accepted standard.

nicost:
I am very reluctant to change style at this point in time.  Happy to use a linter that enforces the current style, but not a fan of changing style to match a linter.

nanthony21:
Makes sense. There's probably a way for us to specify a custom set of rules. I'll take a look at it at some point.

nicost:
To be honest, most of that Google style looks good to me, except for using only two spaces (rather than three) for each new line.  If you are used to four, that can not be nice for you either;)

nanthony21:
I don't really mind either way. The main reason I usually use 4 spaces is that is what most text editors / IDE start out configured as.

It also means I don't need to go reconfigure the editor if I switch between Java and Python for example.

PR #1203, opened 2021-04-23
Allow hiding of imageJ

nanthony21:
Using Micro-Manager usually involves having many separate windows open which can sometimes be a pain to manage if you are switching back and forth between multiple applications. This PR attempts to improve upon this by adding an option to hide the ImageJ window. While the ImageJ window does have some uses I find that most users never use it.

This PR adds an item to the "Window" menu that allows toggling the visibility of the ImageJ window. The visibility setting is saved/loaded to the user profile.

Currently, the default behavior in the absence a user profile setting will be to hide ImageJ


nicost:
The ImageJ window has the tools to set ROIs.  For a user to discover that they first need to make the ImageJ window visible to get to the ROI tools seems very confusing.  I also do not find the ImageJ window the most confusing, plus it is still kind of nice to be considered part of the ImageJ ecosystem.  So, at the very least I would like the default to be to show the ImageJ window.

nanthony21:
Makes sense. Now it will default to keeping the ImageJ window visible.

nicost:
It still feels a bit funny, since the application really is ImageJ, and Micro-Manager is an ImageJ plugin;).  I am fine with it though, if you think it is useful.  Another UI place would be in Tools > Options (that has several usability options), that seemed a bit more logical to me since the Windows menu only lists Dataviewer windows, i.e. it does not list the Inspector or any plugin windows, so it is not a very logical place to list the ImageJ window.

In addition, things may get weird on the Mac.  Whenever you select a DataViewer window, the Mac switches the menu to the ImageJ menu.  I am not sure what this will do when the ImageJ window is hidden.  @marktsuchida could you have a look at that?

nanthony21:
> For a user to discover that they first need to make the ImageJ window visible to get to the ROI tools seems very confusing.  

While I know that this PR doesn't help with the issue, I think this is already confusing. I only recently discovered that the ImageJ tools had an effect on display windows. Since the ImageJ window is the first one to appear, long before the first image is displayed, I had always assumed that it was just there as a side-effect of how Micro-Manager is launched.

I guess I'm only really aware of my own narrow use-case for Micro-Manager so I don't know what other peoples requirements are. Do people use Micro-Manager alongside other ImageJ plugins? To me, launching as an ImageJ plugin has always seemed like an odd way of doing things considering that it has it's own installer, installation directory, and bundled version of ImageJ.

nicost:
When we started Micro-Manager, the primary goal was to make an ImageJ plugin for image acquisition.  Full integration with ImageJ has always been difficult, and with 2.0 the data structures are further away from ImageJ then ever (ImageJ is only used for drawing and for several selection tools), which is a shame from the perspective of many users.  One of the goals of the Wisconsin group is to integrate MIcro-Manager better in the Fiji eco-system, making it easier to use Micro-Manager along with other Fiji plugins.

Irrespective of all that, the ImageJ UI is needed for its selection tools, zoom tool (zooming into specific regions), hand tool (allowing to move the stage while dragging the image), and probably a few more things.  I fully agree this is not obvious to new users, and welcome ideas how to organize this differently (we could include a tool bar with every data viewer for instance), but until we have a replacement, we need to keep the ImageJ window around at least to make those tools accessible.

I am OK with your PR if you feel strongly about it, but wonder if the convenience of skipping a single click to minimize this window outways the frustration of users unable to find the ImageJ window, having forgotten about the Window menu entry.

nanthony21:
I wouldn't say that I feel strongly about it, so we should add this if it'll cause problems. I just thought it made the program look a bit better.

If you have time, take a look at the most recent commit. I just added an example of how we could include the ROI tools in a more obvious way by adding them to the inspector panel.

I didn't find a way to include the startup macro tools but If the user opens the popup menu and clicks "Restore Startup Tools" then they will appear. They don't appear to have any effect anyway though. The documentation mentions an `init()` method which doesn't appear to be defined in our version of ImageJ, so maybe it's just a matter of updating to a newer version.

PR #1204, opened 2021-04-23
remove unused reference resolver from IVy

nanthony21:
Removes an unused reference resolver from Ivy settings that results in the `ant -f fetchdeps.xml` command reporting many errors.


nicost:
Did you check that it is actually not used?  I don't know how to force re-fetching all dependencies, but the surest check is to nuke your ivy cache (for me it is in ~/.ivy2/cache).

nanthony21:
Hi Nico, I deleted the entire ~/.ivy2 folder as well as the "dependencies" folder before running this.

I only tested by debugging `MMStudio`, but I think that if any declared dependencies weren't found then it should be been reported during `ant -f buildscripts/fetchdeps.xml` anyway, right?

nicost:
Awesome!  Yes, `ant -f buildscripts/fetchdeps.xml` should report missing dependencies.

PR #1206, opened 2021-04-26
Add automatic linting of pull requests

nanthony21:
This PR adds a Github workflow that will lint any new code in a Pull Request or that is directly pushed to a branch other than `master`

The workflow can also be triggered manually from the `Actions` tab. In this case the full repository will be checked rather than just the newly changed code.

The code style rules are defined in `.github/linters/sun_checks.xml` which is identical to the Google Java style guide except that the indentation rules have been swapped from multiples of 2 to multiples of 3.

In the future I might make sense to add an intellij project to source control which is preconfigured with the correct module structure and coding style.

See here for an example of the linting output when run manually on the full codebase: https://github.com/nanthony21/micro-manager/runs/2442521145?check_suite_focus=true

### DIFF
--- a/.github/linters/readme.md
+++ b/.github/linters/readme.md
@@ -3,6 +3,6 @@ a copy of the Google Java style guide rules that  have been modified from the
 original Google rules to match the rules specfied at https://micro-manager.org/wiki/Micro-Manager_Coding_Style_and_Conventions
 
 The following changes have been made to the original https://github.com/checkstyle/checkstyle/blob/master/src/main/resources/google_checks.xml :
- - All properties under the "Indentation" module have been changed to 3.
+ - All properties under the "Indentation" module have been changed to be multiples of 3 rather than multiples of 2.
  
  Information about these properties can be found here: https://checkstyle.sourceforge.io/config_misc.html
--- a/.github/linters/readme.md
+++ b/.github/linters/readme.md
@@ -1,0 +1,8 @@
+Super-Linter is hardcoded to look for a `sun_checks.xml` file in this directory for it's Java linting rules. These rules are actually
+a copy of the Google Java style guide rules that  have been modified from the 
+original Google rules to match the rules specfied at https://micro-manager.org/wiki/Micro-Manager_Coding_Style_and_Conventions
+
+The following changes have been made to the original https://github.com/checkstyle/checkstyle/blob/master/src/main/resources/google_checks.xml :
+ - All properties under the "Indentation" module have been changed to 3.
+ 
+ Information about these properties can be found here: https://checkstyle.sourceforge.io/config_misc.html
--- a/.github/linters/sun_checks.xml
+++ b/.github/linters/sun_checks.xml
@@ -19,7 +19,7 @@
 <module name = "Checker">
   <property name="charset" value="UTF-8"/>
 
-  <property name="severity" value="warning"/>
+  <property name="severity" value="error"/>
 
   <property name="fileExtensions" value="java, properties, xml"/>
   <!-- Excludes all 'module-info.java' files              -->
--- a/.github/linters/sun_checks.xml
+++ b/.github/linters/sun_checks.xml
@@ -1,0 +1,363 @@
+<?xml version="1.0"?>
+<!DOCTYPE module PUBLIC
+          "-//Checkstyle//DTD Checkstyle Configuration 1.3//EN"
+          "https://checkstyle.org/dtds/configuration_1_3.dtd">
+
+<!--
+    Checkstyle configuration that checks the Google coding conventions from Google Java Style
+    that can be found at https://google.github.io/styleguide/javaguide.html
+
+    Checkstyle is very configurable. Be sure to read the documentation at
+    http://checkstyle.org (or in your downloaded distribution).
+
+    To completely disable a check, just comment it out or delete it from the file.
+    To suppress certain violations please review suppression filters.
+
+    Authors: Max Vetrenko, Ruslan Diachenko, Roman Ivanov.
+ -->
+
+<module name = "Checker">
+  <property name="charset" value="UTF-8"/>
+
+  <property name="severity" value="warning"/>
+
+  <property name="fileExtensions" value="java, properties, xml"/>
+  <!-- Excludes all 'module-info.java' files              -->
+  <!-- See https://checkstyle.org/config_filefilters.html -->
+  <module name="BeforeExecutionExclusionFileFilter">
+    <property name="fileNamePattern" value="module\-info\.java$"/>
+  </module>
+  <!-- https://checkstyle.org/config_filters.html#SuppressionFilter -->
+  <module name="SuppressionFilter">
+    <property name="file" value="${org.checkstyle.google.suppressionfilter.config}"
+           default="checkstyle-suppressions.xml" />
+    <property name="optional" value="true"/>
+  </module>
+
+  <!-- Checks for whitespace                               -->
+  <!-- See http://checkstyle.org/config_whitespace.html -->
+  <module name="FileTabCharacter">
+    <property name="eachLine" value="true"/>
+  </module>
+
+  <module name="LineLength">
+    <property name="fileExtensions" value="java"/>
+    <property name="max" value="100"/>
+    <property name="ignorePattern" value="^package.*|^import.*|a href|href|http://|https://|ftp://"/>
+  </module>
+
+  <module name="TreeWalker">
+    <module name="OuterTypeFilename"/>
+    <module name="IllegalTokenText">
+      <property name="tokens" value="STRING_LITERAL, CHAR_LITERAL"/>
+      <property name="format"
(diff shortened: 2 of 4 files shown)


nanthony21:
It looks my example lint of the full codebase passed with no warnings which doesn't seem right. I know at least that some of the `libraries` have mismatched indentation. Let me take some time to figure that out.

nanthony21:
I have it working now. Example output of linting the entire codebase is here: https://github.com/nanthony21/micro-manager/runs/2454126500?check_suite_focus=true

nicost:
This looks great!  Can we apply it to the complete code base as well?  How can I see the output of some "corrected" files?

marktsuchida:
I'm all for this. Can this linter (when [run locally](https://github.com/github/super-linter/blob/master/docs/run-linter-locally.md)) also reformat the code? If so, would be great if you could reformat all the code after this PR is merged (aiming for a moment when there are no open PRs).

nanthony21:
@nicost , I currently have it set up so that when the action is automatically triggered by either a Pull Request or by pushing commits to a non-master branch only the newly changed files will be checked. This allows the process to run in just a few minutes and will give a simple "Pass/Fail" output for the PR even if other old code is not yet passing the linter.

However, you can also manually trigger the workflow by going to the "Actions" tab on Github. In this case all `*.java` files will be checked.

@marktsuchida , "Super-Linter" uses [CheckStyle](https://checkstyle.sourceforge.io/) to do its Java linting. As far as I'm aware it doesn't have any support for reformatting. However, we can add the same style guide to IntelliJ and then IntelliJ could be used to perform automatic reformatting. That will easily fix the indentation mismatches. However the linter is currently also checking things that are not so easy to do automatically (like javadoc comments). If we decide we don't care about some things then we can disable them in the linter's configuration file (.github/linters/sun_checks.xml). I think the main benefit here is in making sure that future pull requests have uniform style, going back and getting all the existing code to pass the linter will be a huge amount of work.

On a related note, IntelliJ [recommends](https://intellij-support.jetbrains.com/hc/en-us/articles/206544839-How-to-manage-projects-under-Version-Control-Systems) including the `.idea` and `.iml` project files in source control. What would you two think about adding a default IntelliJ project this repo and removing these files from `.gitignore`?
I've recently switched to using IntelliJ and it's great but I've ended up needing to configure the project a few times now which can be a real pain.

marktsuchida:
Of course IntelliJ recommends that, but that assumes that everybody working on the project uses IntelliJ, and that there is no other build system. So the things I worry about are:

- Given that IntelliJ (as well as other IDEs) can open a Maven project (i.e. `pom.xml`), and we definitely will want to have `pom.xml`s, would the IntelliJ project files end up being a redundant thing that we need to keep in sync? (Not to mention that it is already redundant with the Ant files.) Most of the work of keeping it in sync is setting up the dependencies, which very few people seem to be able to figure out how to do correctly.

- Are we going to officially say that everybody should use IntelliJ (I'm not particularly opposed to that, but there are still NetBeans `*.form` files in our repo), or is it safe for other people to work outside of IntelliJ even after the project files are added?

It would be nice if there were a command line tool that can reformat Java code. I searched a couple of years ago and found none that were satisfactory, except for `google-java-format`, which doesn't allow customization (which I agree is a good thing, as long as you don't have an existing codebase). If we are willing to move away from the highly unusual 3-space indentation (let's admit it, it's the largest cause of incorrectly indented code in our codebase), even just for Java code, then we will suddenly gain a command line tool as well as plugins for both IntelliJ and Eclipse. The IntelliJ plugin can presumably be used even if project files are not included in the repository.

(Note that having a command-line tool means that a git commit hook can be set up, which means that developers who enable the hook will get automatic formatting for every commit, whatever they used for an editor. Then the GitHub linter will only be a check for people who forgot to enable the hook.)

nanthony21:
I haven't used it, but it looks like the IntelliJ formatter can be called from the command line: https://www.jetbrains.com/help/idea/command-line-formatter.html 

In a previous PR (#1202) I had suggested moving to an indent of 2 so we would match the Google style but @nicost  was of the opinion that at this point we don't want to switch away from the 3-space indent that has been used for so long.

I'm not familiar with IntelliJ's handling of `pom.xml` files. I wouldn't suggest configuring IntelliJ to build JAR files or anything like that, but I think that having a project pre-configured to use the correct code style and have all of the basic dependencies/modules configured would be helpful since it is pretty confusing. I don't think that having these project files in source control should interfere with other IDE's so I don't think there would be any need to formally recommend using IntelliJ. You are right though that it is yet another set of files to keep updated.

nanthony21:
Ready to merge?

marktsuchida:
I think ok to merge. This is easy to switch off if it is later found to interfere with code contribution workflows (though I don't expect that).

Regarding the 3-space indents, I don't think it's a good idea to change the indentation rule either, unless there is a clear benefit and everything is planned out.

By the way, it might be useful to set up [editorconfig](https://editorconfig.org/) files.

As for IDEA project files: I think I wouldn't be opposed to adding them if
- the dependencies are set up using the same groups (compile, runtime, etc.) as we use in `ivy.xml` (this should take care of most of the issues with external dependencies) (you do need an additional group for `acqEngine` and `MMCoreJ`), and
- you come up with a simple script that can check that there is one IDEA project (`.iml`) per JAR (meaning, mostly, it checks the projects against the list of modules in the `libraries` and `plugins`, plus `mmstudio` and `autofocus`), and
- people can use the project files without modifying them locally (because they would then accidentally commit local changes, which we don't want)

If the check script is run using GitHub Actions, we'd be able to catch it whenever a PR would cause the project files to go out of sync, even though updating would still require manual intervention.

I currently have my projects set up like this (plugin and library projects have the prefixes to deal with identically named projects):
```
$ find . -name '*.iml'
./autofocus/autofocus.iml
./library-Big/library-Big.iml
./library-ExampleLibrary/library-ExampleLibrary.iml
./library-Gaussian/library-Gaussian.iml
./library-ImageProcessing/library-ImageProcessing.iml
./library-Projector/library-Projector.iml
./mmstudio.iml
./plugin-AcquireMultipleRegions/plugin-AcquireMultipleRegions.iml
./plugin-ASIdiSPIM/plugin-ASIdiSPIM.iml
./plugin-Assemble/plugin-Assemble.iml
./plugin-autolase/plugin-autolase.iml
./plugin-ChannelCorrector/plugin-ChannelCorrector.iml
./plugin-CRISP/plugin-CRISP.iml
./plugin-CRISPv2/plugin-CRISPv2.iml
./plugin-DemoCustomDisplay/plugin-DemoCustomDisplay.iml
./plugin-Duplicator/plugin-Duplicator.iml
./plugin-Emu/plugin-Emu.iml
./plugin-Example/plugin-Example.iml
./plugin-FrameCombiner/plugin-FrameCombiner.iml
./plugin-Gaussian/plugin-Gaussian.iml
./plugin-HCS/plugin-HCS.iml
./plugin-HelloWorld/plugin-HelloWorld.iml
./plugin-IlluminateLED/plugin-IlluminateLED.iml
./plugin-ImageFlipper/plugin-ImageFlipper.iml
./plugin-IntelligentAcquisition/plugin-IntelligentAcquisition.iml
./plugin-Magellan/plugin-Magellan.iml
./plugin-MMClearVolumePlugin/plugin-MMClearVolumePlugin.iml
./plugin-MultiAndor/plugin-MultiAndor.iml
./plugin-MultiChannelShading/plugin-MultiChannelShading.iml
./plugin-pgFocus/plugin-pgFocus.iml
./plugin-PipelineSaver/plugin-PipelineSaver.iml
./plugin-PointAndShootAnalysis/plugin-PointAndShootAnalysis.iml
./plugin-Projector/plugin-Projector.iml
./plugin-PTCTools/plugin-PTCTools.iml
./plugin-RandomizePositionNames/plugin-RandomizePositionNames.iml
./plugin-RatioImaging/plugin-RatioImaging.iml
./plugin-Recall/plugin-Recall.iml
./plugin-RTIntensities/plugin-RTIntensities.iml
./plugin-SequenceBufferMonitor/plugin-SequenceBufferMonitor.iml
./plugin-SlideExplorer/plugin-SlideExplorer.iml
./plugin-SnapOnLive/plugin-SnapOnLive.iml
./plugin-SplitView/plugin-SplitView.iml
./plugin-Tracker/plugin-Tracker.iml
./plugin-WhiteBalance/plugin-WhiteBalance.iml
./plugin-ZProjector/plugin-ZProjector.iml
```
And you can see that some of the recently added plugins are missing, which can be a problem when trying to inspect all uses of a particular class or method, for example.

I'm not sure how to deal with the fact that the location of the Ant-built `MMCoreJ.jar` and `MMAcqEngine.jar` differs between Windows and Unix. I'm also not sure how to provide a Run/Debug configuration that works for everybody (I personally set the working directory to an MM binary installation) -- and how to prevent people from committing their local modifications to such settings. (I suppose we could provide a project file generator instead of project files themselves, keeping the project files in `.gitignore`, but that might be a lot of work and complexity.)

PR #1222, opened 2021-05-03
Memory leak fixes

nicost:
Reduces memory consumption significantly, since many large objects are now garbage collected.  Still a long way to go.  Requesting reviews to get input on the best strategy to proceed.  Merging now, since we will need this for our experiment tomorrow;).

### DIFF
--- a/mmstudio/src/main/java/org/micromanager/acquisition/internal/AcquisitionWrapperEngine.java
+++ b/mmstudio/src/main/java/org/micromanager/acquisition/internal/AcquisitionWrapperEngine.java
@@ -797,6 +797,7 @@ public final class AcquisitionWrapperEngine implements AcquisitionEngine {
    public void onAcquisitionEnded(AcquisitionEndedEvent event) {
       curStore_ = null;
       curPipeline_ = null;
+      studio_.events().unregisterForEvents(this);
    }
 
    @Subscribe
--- a/mmstudio/src/main/java/org/micromanager/acquisition/internal/DefaultAcquisitionEndedEvent.java
+++ b/mmstudio/src/main/java/org/micromanager/acquisition/internal/DefaultAcquisitionEndedEvent.java
@@ -23,6 +23,12 @@ package org.micromanager.acquisition.internal;
 import org.micromanager.data.Datastore;
 import org.micromanager.acquisition.AcquisitionEndedEvent;
 
+/**
+ * This event signifies that an acquisition has been ended.
+ *
+ * This default implementation of this event is posted on the studio event bus,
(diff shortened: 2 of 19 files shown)


nanthony21:
Looks good. So is the main issue that objects don't unregister themselves from events or remove `ActionListener`s so they don't get GC'd?

Out of curiosity what is the purpose of removing the `final` qualifier from so many variables?

nicost:
To be honest, I am still digging.  I was previously under the impression that the Swing method "dispose()" would take care of making objects available for garbage collection, but that was my misconception.  So, we need to ensure that there are no references remaining to an object (or a group of objects that has references to each other).  The problem is figuring out what the linkage to the root of the application is.  The current PR removes references to big chunks of memory (especially the first image of every dataset was kept in memory), but there are still many, many objects accumulating, which eventually will make the application sluggish (just the garbage collection becomes unwieldy).  I'll continue working on this.

Great point about the `final ` qualifier, I dislike removing those.  However, to set those variables to null (to loose the reference to that object), they can not be final.  It is possible that many of these null assignments are not needed, at this point it is very difficult to find the reference linkage to the root of the application (help will be appreciated!).

nanthony21:
I didn't realize that Micro-Manager had this issue. We often do multi-day automated experiments saving ~5TB of data and never have issues. Though, I guess the computer being used has 32GB RAM.

Based on what I see online I get the impression that setting variables to `null` and removing action listeners shouldn't usually be needed.
https://coderanch.com/t/344003/java/call-remove-Listeners
https://stackoverflow.com/questions/2931170/does-variable-null-set-it-for-garbage-collection

Have you tried using something like [VisualVM](https://docs.oracle.com/javase/8/docs/technotes/guides/visualvm/profiler.html) to find what is using the most memory?

nicost:
The problem mainly manifests itself when doing many acquisitions (thousands).  There is a leak linear with the number of acquisitions, not so much the amount of data per acquisitions (which is why we could ignore this for so long).

Both IntelliJ and Netbeans (which seems to use VisualVM as its UI looks very similar) have build-in profilers/memory viewers.  Starting to get the hang of the IntelliJ one, as it shows numbers of objects, and lets you see what refers to those objects.  Wat really would help is a visual  graph of the relations between objects (and possibly automated detection of clouds of objects tied to the main application with only a very few references.  See the second answer in https://stackoverflow.com/questions/1910194/how-does-java-garbage-collection-work-with-circular-references.

nicost:
I think that I found the main culprit, backdoor references to DisplayController created in SaveButton and GearButton.  These use plugins that somehow keep the reference to DisplayController alive.  See PR #1224. I will go back and restore some of the code changes here that turn out to be not necessary.

PR #1241, opened 2021-05-21
Add `getByteArray` to `Image`

nanthony21:
This PR adds a method to `Image` that returns a `byte[]` array of the raw pixel data regardless of the pixel data type.

I use this method to efficiently transfer image data to Python using Py4J. More information [here](https://www.py4j.org/advanced_topics.html#byte-array-byte). A similar approach may be useful to Pycro-Manager if not already implemented.

This method could also be useful for other low-level processing of raw data.

The primary implementation is added to `BufferTools` which `DefaultImage` then calls upon.

### DIFF
--- a/mmstudio/src/main/java/org/micromanager/data/Image.java
+++ b/mmstudio/src/main/java/org/micromanager/data/Image.java
@@ -243,4 +243,12 @@ public interface Image {
     */
    Object getRawPixelsForComponent(int component);
 
+   /**
+    * Regardless of the the value of {@code #getBytesPerPixel} return an array of bytes containing
+    * raw data for the image. Useful for doing low-level handling of raw image data.
+    *
+    * @return The byte array containing raw data for this image.
+    */
+   byte[] getByteArray();
+
 }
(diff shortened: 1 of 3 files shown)


marktsuchida:
Good method to have. The documentation should be very clear that in some cases the returned `byte[]` is not a copy but the actual underlying buffer of the `Image`, and therefore should never be modified (the consequence being unpredictable data corruption). Either that or the implementation should always make a copy.

nanthony21:
Thanks @marktsuchida , I just made some changes so that a copy should always be returned and updated the javadoc to make that clear.

nanthony21:
@marktsuchida , Ok to merge?

marktsuchida:
@nanthony21 Looks good, but it is an API change, so it should wait until 2.0 is out unless there is a special reason?

nanthony21:
Ok, sounds good. No rush

nanthony21:
Just checking back in on this. Now that 2.0 is out are we ok to merge this?

PR #1244, opened 2021-06-02
Improves display performance under load.

nicost:
The display now stays active and responsive when data are acquired at 500 fps.  A couple of changes were made:
- An older hack to throttle the display in DisplayController.onNewImage was removed.
- Some calls to ExpandDisplayRangeCoalescentRunnable(position) were removed since they were redundant.
- The DisplayUIController.displayImages function now ignores requests to display a new image when a paint is still in progress.  This is debatable since we are using CoallescentRunnables to schedule displaying images, but apparently, these can be scheduled faster than the paint can be executed, leading to the old problem of the display choking up.  Ignoring requests is probably not always a good idea, but will only happen when the datarate is very high, when it is not really a problem.  Better solutions are welcome, but this one seems to work.
- Avoid setting the "NEW IMAGE" indicator visible/invisible too often (drawing just this indicator seems to slow down the display significantly).
- Added option to set Thread Priority to THreadFactory.  This did not make a significant difference, but it is nice to have the option, and should not harm anything (or make things much more complex).


mmstudio/src/main/java/org/micromanager/display/internal/animate/AnimationController.java
```diff
@@ -309,8 +311,8 @@ public void run() {
             @Override
             public void run() {
                synchronized (AnimationController.this) {
+                  // TODO: should this be done at all?
                   listeners_.fire().animationAcknowledgeDataPosition(newPosition);
-
```
marktsuchida: Not sure at this point, but I think originally this was part of the mechanism to prevent the last frame from being missing from the display window scrollbar range.

mmstudio/src/main/java/org/micromanager/display/internal/animate/AnimationController.java
```diff
@@ -89,9 +89,10 @@
 
    private final ScheduledExecutorService scheduler_ =
          Executors.newSingleThreadScheduledExecutor(ThreadFactoryFactory.
-               createThreadFactory("AnimationController"));
+               createThreadFactory("AnimationController", 4));
```
marktsuchida: Thread priority sounded nice to me, too, when I first saw it could be set, but the more you read about what it actually does (or doesn't) in each OS, the less attractive it seems.
See, e.g., https://vanillajava.blogspot.com/2011/09/why-thread-priority-rarely-matters.html.
I would vote for not introducing this, for simplicity's sake (a conscientious future developer might find it terrifying to modify this code if they wondered what behavior depends on the thread priority) -- since you say it doesn't make much of a difference.

mmstudio/src/main/java/org/micromanager/display/internal/animate/AnimationController.java
```diff
@@ -349,12 +351,14 @@ public void run() {
             public void run() {
                synchronized (AnimationController.this) {
                   listeners_.fire().animationAcknowledgeDataPosition(newPosition);
-                  if (newDisplayPosition != oldPosition) {
+                  if (!newDisplayPosition.equals(oldPosition)) {
```
marktsuchida, line 353: Nice catch!

mmstudio/src/main/java/org/micromanager/display/internal/displaywindow/DisplayController.java
```diff
@@ -526,8 +526,8 @@ public void run() {
    protected Coords handleDisplayPosition(Coords position) {
       perfMon_.sampleTimeInterval("Handle display position");
 
-      runnablePool_.invokeLaterWithCoalescence(
-            new ExpandDisplayRangeCoalescentRunnable(position));
+      //runnablePool_.invokeLaterWithCoalescence(
+      //      new ExpandDisplayRangeCoalescentRunnable(position));
```
marktsuchida: It would be nice to avoid commented code (old code is kept in git!).

In the original way I expected this code to work, removing this would be a bug (same as omitting the `animationAcknowledgeDataPosition()` firing elsewhere), but is there a mechanism that replaces this now? If so, the 2 lines should just be deleted. If updates to the display range (i.e. the display's model of the size of the datastore) are now handled without coalescent runnables, then it would probably be cleanest to remove all related code. If there is no such mechanism, then I'm pretty sure these lines are needed, though I'd be happy to be proven wrong.

marktsuchida (commented):
Here are my comments -- to the best of my understanding of how this code currently works.

mmstudio/src/main/java/org/micromanager/display/internal/displaywindow/DisplayController.java
nicost: Just tested, and leaving this code in slows down the display to the point of it becoming unworkable.  The display range is updated in the DisplayUIController.displayImage function.  Not sure about code paths that do not call displayImage (if there are such updates to the viewer without displaying an image).

mmstudio/src/main/java/org/micromanager/display/internal/animate/AnimationController.java
nicost: Yes, this may be part of one of the code paths updating the sliders when displayImage is not called.

PR #1245, opened 2021-06-03
Crisp plugin

bls337:
Hello Nico & everyone, this is my first pull request to Micro-Manger 2.0, hopefully I did everything right. Thanks!

This is my update to the CRISP plugin. This is Brandon from Applied Scientific Instrumentation.


marktsuchida:
@perceptron42 Thanks for this. It looks like this would replace @nicost's `CRISP` plugin, rather than ASI's existing `CRISPv2` plugin, leaving the latter in place -- perhaps not what you intended? Or is this supposed to replace both?

Also, it would be great if you could keep the `src/main/java` part of the path (before `com/asiimaging`) -- this is a standard directory layout for Java projects (not just ours), and by having source in that directory the build system will automatically pick it up.

perceptron42:
Thanks for the help. I changed the project directory structure. Yes, the idea is to replace both of the current CRISP plugins.

marktsuchida:
@bls337 Thanks for fixing. I added a commit to remove CRISPv2.

PR #1250, opened 2021-06-05
Adrresses slowness of display of  datasets > ~10,000 images

nicost:
When acquiring 50,000 images at 5 msec exposures, the display would start off nice and fast, but slow down over time.  This PR addresses that slowness, at least for single channel data.  Most of the gains seem to be made by optimizing the Storage function "getImagesIgnoringAxes".  This function would iterate through all Coords in the dataset, remove the requested axes and check for identity.  This becomes slow with large collections.  By keeping track of the axes in the dataset, and only doing the complete check when the "missing axes" are actually present in the dataset, this bottleneck was removed.  However, this approach does not work when the "missing axes" are present in the dataset (such as in multi-camera datasets), so it will be good to think about alternative optimizations (none of which are pretty).

In addition, some synchronization has been replaced with Atomic Booleans or ConurrentHashMaps, and additions were made so that perfMon can be set to null without crashing the application.

Display now stays fast even when acquiring 50,000 images (single channel) with the demo camera.

### DIFF
--- a/mmstudio/src/main/java/org/micromanager/data/internal/StorageRAM.java
+++ b/mmstudio/src/main/java/org/micromanager/data/internal/StorageRAM.java
@@ -21,9 +21,12 @@
 package org.micromanager.data.internal;
 
 import com.google.common.eventbus.Subscribe;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
 import java.util.Set;
 import java.util.TreeSet;
-import java.util.concurrent.ConcurrentHashMap;
 import org.micromanager.data.Coords;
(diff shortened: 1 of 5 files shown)


marktsuchida:
You can't replace `synchronized` with `ConcurrentHashMap`, because `synchronized` protects multiple accesses to the hash map as a whole. The same applies to atomic variables.

nicost:
Alright, reverted synchronizedHashMap, but kept the AtomicBoolean.  The AtomicBoolean in AnimatonController makes it possible to not synchronize the method "isAnimating()".  Since that is called often, this is a worthwhile optimziation.

PR #1258, opened 2021-06-16
Update README.md

nanthony21:
Fix the description of how to clone submodules. `--remote` was mentioned near the bottom of the readme but was left out from the top section. Until a better option that using the `--remote` option is implemented this will make the readme more accurate.


marktsuchida:
LGTM.

PR #1259, opened 2021-06-17
Fixes a few issues with the Display

nicost:
- Prevents a deadlock that could occur when closing a display that is animating.
- Prevents situations where the display has incorrect axes, by updating the axes when the display freezes.
- Fixes a bug where closing one display would effectively kill the linkmanager, making it impossible to link the remaining displays.

Also applied Checkstyle recommendations to DisplayController.


marktsuchida (approved):
Looks good to me!

PR #1265, opened 2021-06-23
Update to library with selectable port number

henrypinkard:


nicost:
Thanks Henry!  How does the user set the port number?

henrypinkard:
Its only in the pycro-manager headless launcher right now, not GUI. But would be easy enough to add in by using the new constructor:

https://github.com/henrypinkard/pycro-manager/blob/0e0e7212faef9aa90d7c6313e5ec99a1f17f2ecd/java/src/main/java/org/micromanager/internal/zmq/ZMQServer.java#L70

nicost:
I was thinking about adding it as a command line or JVM option (so that the user can launch an MM instance that will listen to the specified port), but will wait to see if there is a need.

henrypinkard:
That's already how it works when you launch headless from pycromanager--you can just pass the port as an arg:

https://github.com/micro-manager/pycro-manager/blob/c569415bde766c6735b0d75e3945662b4b6ab820/pycromanager/acquire.py#L58

PR #1267, opened 2021-06-28
Makes Annotations work how they presumably were meant to work.

nicost:
Both Single and MultiPageTiff storage now stores Summary and Per Image comments in a file "comments.txt" in the acquisition directory.  Even after opening a file, changes to the comments will be stored in that file.  One possible addition would be to preserve the original comments, but it is already a big step forwards that things actually work.  The code is not easy to understand, and probably could use some structural improvements, but that will happen another time.


marktsuchida (approved):
Looks good to me.

Regarding storing the original comments, the natural generalization would be to store a complete history of comment editing history. It could be argued that we should either not do it at all, or do it fully. But there are other tools out there for change tracking, and I don't think we want to reinvent ours. The remaining options are completely immutable (after acquisition start) vs mutable any time, but given how easy it is to start an acquisition without remembering to set the correct comment, I think the completely mutable behavior here is actually the best.

PR #1292, opened 2021-07-27
Add ChannelGroup setting to demo config

ianhi:
This adds an example to the demo config to make this more discoverable. This also helps when using the demo config to develop a GUI application. 

This is particularly important for people working with pymmcore(-plus), as when doing so you can end up working with config files not generated directly by micromanager. 

See also https://forum.image.sc/t/how-to-set-channelgroup-in-config-file/55746


marktsuchida:
Excellent.

PR #1302, opened 2021-08-16
Remove problem report sending function

marktsuchida:
We no longer maintain the server necessary for this, and prefer to provide support on open forums.

This PR removes the functionality to send a finished problem report directly; the ability to view the report and save as a file is unchanged.

The report format has been slightly changed to remove the username and host info in the header, although this does not guarantee against private information being included as part of the logging.


nicost:
Looks good to me.

PR #1314, opened 2021-08-30
Use multiple camera modes in Demo Config

ianhi:
This makes it clearer that something has happened when you switch channels. Unfortunately I had to duplicate noise as the camera only has three modes.


marktsuchida:
This seems like a useful idea, but I think it would be better to keep the camera mode independent of the fluorescence channel (it might be confusing if, for example, the user manually switched to gaussian noise but got something else after switching the channel).

How about changing the `StripeWidth` (and maybe also `RotateImages`) properties between the channels, so that a difference can be seen as long as the user stays in the (default) sine wave mode?

PR #1362, opened 2021-12-19
Fixes bug in Oughtafocus.  Improves GUI behavior of Autofocus.  Lots of cleanup.

nicost:


marktsuchida:
@nicost This broke the build, because the build scripts currently require that each autofocus plugin be a single `.java` file.
(Of course that's not ideal, but it's the way it is....)

Is it okay to move `class BrentFocusOptimizer` back into `OughtaFocus.java`, at least for the time being?

nicost:
Sure.  It just is a very ugly thing... Bit busy right now, so if you could do it, that would be great

Forgot about that build weirdness...

marktsuchida:
Sounds good, will do.

PR #1365, opened 2021-12-21
Specify the SVN revision of 3rdparty, 3rdpartypublic to use for build

marktsuchida:
This is analogous to the secret-device-adapters-commit file in the mmCoreAndDevices repository.

Having this file pins the SVN revision of 3rdparty and 3rdpartypublic to be used to build a particular commit of this repo, making it possible to trace the full input to the build (modulo build tools). One small step in the direction of reproducible builds.

It also makes it easier to manage the CI triggers, because it eliminates the need to rebuild micro-manager when external repos change.

The downside is that these files need to be updated manually when files are added to the SVN repos. We just need to be careful to check this when merging pull requests.

Cc: @nicost @nanthony21


nanthony21:
This seems to me like a worthwhile addition. @marktsuchida what are your thoughts on moving to Git-LFS in the future? Then we could use the submodule functionality rather than tracking this manually.

marktsuchida:
@nanthony21 I'm not sure exactly what you mean, because when it comes time to replace the SVN repos, the primary problem is not large individual files. If you mean to propose that we have a git version of 3rdparty or 3rdpartypublic, then, please, no! It makes absolutely no sense to put all dependencies in a single repository. It's the source of so may headaches.

I generally don't like the idea of using submodules for dependencies. For vendor SDKs, the default method should be to use the SDK installed on the system (in Program Files on Windows). The idea is the same as traditional Unix `configure` scripts. Otherwise it is too hard for many people to "just build for the device adapter using the SDK I have installed on my computer", which should be the most natural thing to do. Pinning the source of dependencies is important for CI and releases, but can be counterproductive for the default build system.

For CI, we can have private repos (one for each dependency!) that contain the necessary headers and libs (but _not_ anything not required for the build), and maybe Git-LFS will be needed for this (though I would hope very rarely, and quite probably never), but we can't use those as submodules without making it extremely inconvenient for people outside the team. Submodules of public repos should be public (as I think I already wrote elsewhere). Meson wrap files or some other mechanism (such as package managers, though I have different reservations for those in our specific context) seems better suited to the job.

Additionally, it is important to realize that many device adapter contributors will be beginning or occasional Git users. A dependency version management mechanism based on git submodules (aside from being potentially daunting) is just asking for trouble (and scales poorly from a maintainer's viewpoint). It will be though to even be sure that the same thing is being built by the contributor as being committed.

nanthony21:
Thanks Mark, that all makes sense.

PR #1368, opened 2021-12-23
Avoid shipping log4j

nicost:
IT departments start to disallow software that includes log4j.  This PR removes log4j, but does not do so cleanly, because the latest scifio release still brings in a dependency on log4j.  Hitting the missing functionality is extremely unlikely, as log4j was only used by a library by Herr Saalfeld that we were not using, but that was brought in with imglib2-roi.

### DIFF
--- a/buildscripts/ivy.xml
+++ b/buildscripts/ivy.xml
@@ -100,6 +100,9 @@
 
 		<!-- Run-time-only dependencies -->
 		<dependency conf="runtime" org="org.swinglabs" name="swingx" rev="0.9.5"/>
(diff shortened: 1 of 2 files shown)


nanthony21 (approved):

marktsuchida (approved):
Looks good!
It might be good to also edit the InnoSetup file to make sure log4j is actually removed by the Windows installer.

nicost:
I believe that the following lines in the .iss file take care of removing all stale jars:

[InstallDelete]
; Remove JARs that are no longer used or have been renamed
Type: filesandordirs; Name: "{app}\plugins\Micro-Manager"

PR #1370, opened 2022-01-12
Code cleanup

nicost:
Modified many files to make them compliant with our Checkstyle plugin.  However, I did not add Javadoc in places I did not know what to add or when it would take too much time to figure out meaningful comments. Since this touches on so many files, it would be great to have a quick look (and also to make sure it does not interfere with work that you may be doing).


marktsuchida (commented):
@nicost I'm not sure how to review so many changes, although most appear to be style-only changes, which are of course fine. Are there any substantial changes that I should look at, or new Javadoc comments that you want me to take a look at?

(Just as a suggestion for the future, it might help all of us to separate commits not by the file(s) they touch, but by the transformation they introduce -- especially so that automatic transformations are in separate commits from any manual edits/cleanup. That way _most_ of the automated transformations can be accepted without further human review -- or at least with more confidence that random errors are not introduced.)

nicost:
There are no automated changes here, this is all loving manual work;)  No need to review everything in detail, I mainly want you and Nick to be aware before merging so as not to interfere with what you may be doing.  There should be no functional changes (unless I goofed, which is well possible).

nanthony21:
Wow, that's a lot of changes! I reviewed the first ~1/5th of them and they all look fine to me. So assuming that the rest is more of the same then this is good by me.

marktsuchida:
Cool!

PR #1378, opened 2022-01-22
Make macOS x86_64 build work on Apple arm64

marktsuchida:
This is only useful for testing the build, for now.

The macOS nightly build script now works correctly when run on an Apple
arm64 computer, and produces x86_64 binaries.

The only change needed was figuring out how to cross-compile Boost,
which appears to be broken when using only command-line options to b2.
(I.e., the method we used to use for Intel 32/64-bit does not work.)

The method used here may become useful if we eventually decide to build
arm64 or universal2 binaries.

Actually running Micro-Manager on arm64 machines (via Rosetta 2 or
otherwise) is not addressed (the main issues are Gatekeeper path
randomization and, if running the x86_64 version, how to launch an
x86_64 JVM).


marktsuchida:
I need to test this back on an x86_64 machine before merging.

PR #1383, opened 2022-02-01
Added a basic Gamepad API

bls337:
Added a basic API to get and set the multiplier for buttons in the axis assignment table. Also fixed some issues with saving and setting values. The plugin needs to be open to gain access to the axis table so that API methods can work.


marktsuchida:
@bls337 It looks like this doesn't build, and there is no `GamepadAPI.java`. Any chance you forgot to add it to the commit?

bls337:
Oops! I just submitted another pull request with the file.

marktsuchida:
@bls337 Thanks!
(Ref: #1385)

PR #1389, opened 2022-02-04
Replace VC2008/2010 Redistributable with 2015-2022

### LINKED ISSUE #1377: vs2019 branch: Installer should bundle new VC++ Redistributable
Currently we ship the VC++ 2008 and 2010 Redistributables. On the `vs2019` branch this needs to be replaced with the VC++ 2015-2022 Redistributable.

See https://docs.microsoft.com/en-US/cpp/windows/latest-supported-vc-redist

https://github.com/micro-manager/micro-manager/blob/6a6f4ecbaa4d2c9e6476ef4779e7b249215b437a/buildscripts/installer_Windows.iss#L70-L76

marktsuchida:
Closes #1377.

Not tested yet: [instructions](https://docs.microsoft.com/en-US/cpp/windows/redistributing-visual-cpp-files?view=msvc-170#install-the-redistributable-packages) suggest that we might need to ignore errors in case a newer version is already installed.


marktsuchida:
When `vc_redist.x64.exe` is run (using `start /wait`), and a newer version is already installed, the exit code is non-zero (actually `1638`). However, our installer does not seem to have a problem with this (the exit code appears to be ingored).

There is a different issue, which is that if an _older_ version (e.g. the 2015-2019 version, as opposed to the 2015-2022 version that we are installing; or an older version of 2015-2022) is already installed, a restart is requested. In most of my testing, this occurred in the form of our installer prompting the user to (optionally) restart at the end, which is what we want. However, there was one instance in which it kicked off a reboot willy-nilly (I am unable to figure out the conditions to reproduce this).

Adding the `/norestart` option does not inhibit our installer from prompting to restart (presumably through registry keys indicating pending operations), so seems like a good idea, in the hope that it prevents an unconditional reboot.

Issue #1392, opened 2022-02-08
Build improvements master plan


henrypinkard:
I think it might be useful to have one issue where we can keep track of the many improvements to the build system, so that it is easier to keep on top of it and contribute. @marktsuchida @nanthony21 @nicost, does anyone want to summarize where we are headed?

FYI, I just (somewhat painfully) figured out how to automatically build and deploy pycro-manager-Java and its libraries using Github actions. Example [here](https://github.com/micro-manager/pycro-manager/blob/main/.github/workflows/maven_deployment.yml).

marktsuchida:
Yes.

marktsuchida:
Here is an overview of the short term goals. The overarching principle is to avoid changing more than one thing about the build at a time, as much as possible.

# Recently finished:
- Moved our automated builds (Jenkins) and download server to the cloud
- Dropped support for 32-bit x86 (both Windows and macOS)
- Updated macOS build to work with current Apple compilers

# Current (almost done):
- https://github.com/micro-manager/mmCoreAndDevices/issues/40

# Next:
- Switch our C++ standard from C++03 to C++14
  On Windows, this happens automatically by virtue of building with VS2019.
  Make `-std=c++14` the default for macOS and Linux.
  After this, people can contribute code that requires modern C++.

- Make our code compatible with C++17
  This mostly means replacing `std::auto_ptr` with `std::unique_ptr`.
  Updates should be made while keeping compatibility with C++14; no need for a
  branch.

- Provide code-signed installers (Windows, macOS) #396

# Later this year

We want to eliminate the hard-to-maintain dual build system (Visual Studio + GNU Autotools) and make it easier for people to get started with parts of the code and make contributions.

Below are the desired end results, **not** in order of implementation; the actual transition plan is complex (and partly TBD); separate issue(s) will be created as we go.

- Each module (device adapter, plugin, MMCore, MMDevice, MMStudio, etc.) should live in its own Git repository. Each repo should have a single license, version numbering (which can be date-based for modules not exposing an API), and CI.

  - This better matches the reality of our development: the majority of contributors work on one or a few modules, and for them, dealing with the complexity of the full MM build is counter-productive.

- Eliminate the subversion repositories `3rdparty` and `3rdpartypublic`.

  - We shouldn't have to provide binaries for open source C++ dependencies: they should be detected on the system (Unix) or else automatically built by our build system.

[...]

PR #1397, opened 2022-02-11
Update ClearVolume dependency which is now in Maven

fabio-echegaray:
Hello,

This PR has the following changes:

1. ClearVolume can be pulled from Maven entirely now, so I changed the relevant lines in ivy.xml.
2. The IconLoader project is not on Maven, but it has a GitHub repository at https://github.com/bulenkov/iconloader. This PR fetches the content of the GitHub repo and copies the jar from there; for this, it uses JGit (https://www.eclipse.org/jgit/).
3. The previous changes triggered some errors at compilation. Attending this, I added the Apache Commons dependencies which are required for the compilation of mmstudio (at least under Linux).

Is fetching code from GitHub something that is discouraged? If this is not welcomed, I can update the PR, let me know. Finally, I haven't tried compiling under Windows or MacOS.


marktsuchida:
It would be wonderful if you could put the different changes into separate PRs, or at least separate commits :)

- I guess this happened in parallel to my cleanup #1395, which already removed the old BinTray repos for ClearVolume. Versions did not need to change there; the versions we use are available from the image.net repo.

- There is already a line for `commons-lang3` 3.5 in `ivy.xml`. It is indeed a bug that mmstudio imports `org.apache.commons.lang`, so that code change is great. (`commons-lang` gets pulled as a transitive dependency, probably for one of the other things you changed.)

- The addition of the missing `linux-amd64` natives for JOGL and gluegen is also great.

- Regarding `iconloader`, I feel conflicted. In principle it is a good thing if our own build of iconloader is specified by code, as you have done here, but I'm not sure the added complexity and build dependencies are worth it, given that it still doesn't eliminate all dependencies in `3rdpartypublic`. It also breaks the assumption that `fetchdeps` does not build anything. One idea is for us to fork iconloader on GitHub (in the micro-manager org) and add a `pom.xml` that works with [JitPack](https://jitpack.io/) (I checked existing forks but the only one that has a `pom.xml` is outdated).

marktsuchida:
(Also one reason to handle iconloader the way I proposed is that we are hoping to stop using Ant (and Ivy) entirely in the future, switching to Maven proper: see #1392.)

fabio-echegaray:
Hello Mark,

Thank you for the feedback. I have attended the comments regarding Ivy dependencies and removed the integration with Git now. It is compiling on Linux correctly; however I have no direct way to compile it under Windows or MacOS yet.

I will give JitPack a go.

marktsuchida:
Perfect, thanks!

Issue #838, opened 2020-06-12
micromanager won't compile with swig 4.x


dschneiderch:
see #837 

to see your swig version:
`swig -version`

current work around is to install swig 3.0.12

On Ubuntu 20.04 LTS (default swig is 4.x) I was able to use:
```
sudo apt install swig3.0
sudo update-alternatives --install /usr/bin/swig swig /usr/bin/swig3.0 1
sudo update-alternatives --install /usr/bin/swig swig /usr/bin/swig4.0 2
sudo update-alternatives --config swig
```
then choose the version you want to use and attempt to compile micromanager.
Note, on ubuntu 18+ you need to hardcode jogl2.jar path in plugins/MMClearVolume/build.xml by replacing current contents with the following (see #708 )
```
<project name="MMClearVolume" basedir="." default="jar">
   <import file="../javapluginbuild.xml" />
   <path id="plugin.classpath">
           <pathelement location="/usr/share/java/jogl2.jar" />
   </path>
</project>
```

nicost:
I made some changes to MMCoreJ.i that still compile with swig 3, but I believe to also compile with Swig 4.  You can find the branch at github.com/nicost/micro-manager under the name swig4.  Or you can copy the file MMCoreJ.i (https://github.com/nicost/micro-manager/blob/swig4/MMCoreJ_wrap/MMCoreJ.i) into your current source code.  You may need to do a make clean in the upper build directory to ensure that swig will run again.  Curious to hear if that compiles OK with swig 4 generated code (if so, I'll push this into the main branch). I will also look at the ant issue in the CLearVolume plugin.

dschneiderch:
I copied the new file into my directory, switched back to swig4.0, ran make clean, then make clean, ./autogen.sh, make and get similar error.

```
Making all in mmstudio
make[1]: Entering directory '/home/raptor/Documents/GitHub/mm_scripts/micro-manager/mmstudio'
JAVA_HOME=/usr/lib/jvm/java-11-openjdk-amd64 /usr/bin/ant -Dmm.javacflags="-Xlint:all,-path,-serial" -logger org.apache.tools.ant.listener.SimpleBigProjectLogger jar
Buildfile: /home/raptor/Documents/GitHub/mm_scripts/micro-manager/mmstudio/build.xml

mmstudio.compile:
    [mkdir] Created dir: /home/raptor/Documents/GitHub/mm_scripts/micro-manager/mmstudio/build
[...]

PR #1398, opened 2022-02-11
Fix error in the signature call to DoubleVector constructor that arises when compiling with Swig 4.0.

fabio-echegaray:
This is a workaround related to issue #838 (and also issue 37 of mmCoreAndDevices). In essence, this PR addresses the java compilation errors thrown when creating DoubleVector objects in MMCoreJ; it does it by changing the call signature from "new DoubleVector(count)" to "new DoubleVector(count, value)", with value=0 (proposed as a workaround in the Swig Changelog of version 4.0.0).

Compilation goes well under Linux; haven't tested it under Windows or MacOS. This change might potentially need some further testing. I am happy to keep updating the PR if required.

Related issue in mmCoreAndDevices https://github.com/micro-manager/mmCoreAndDevices/issues/37


fabio-echegaray:
Apparently Swig 3.0 doesn't like the change. Closing by now.

Issue #1404, opened 2022-02-14
Possible improvements to nightly build


marktsuchida:
We currently perform nightly builds by kicking off independent Windows and macOS builds at 6:00 UTC, each building the tip of `main`. This has these downsides:

- A new build is performed even if no code changes were made in the past day
    - This consumes compute and storage resources (a minor problem)
    - It also results in more work when testing past nightly builds to determine when a bug was introduced
- There is no strictly correct mapping between the code revisions (Git SHAs) and the build; it can only be estimated
- There is no guarantee that the Windows and macOS builds from the same date use the same code revisions

After recent changes (#1365 and https://github.com/micro-manager/mmCoreAndDevices/pull/101; also see #1308), we have achieved a state where the Git SHA of the `micro-manager` repo fully specifies all code used to build Micro-Manager. (This does not mean the build is fully reproducible, but at least it is unambiguous what revision of our code is used, provided that the build is set up correctly.)

It would be nice if we could make it so that it is easy to tell the exact code used for nightly builds, and also avoid performing redundant builds when the code is exactly the same.

One way to do this might be as follows:
- Create a long-lasting branch called `nightly`, which tracks `main`, but is only updated, automatically, once a day at 6:00 UTC,
- Set the nightly builds to run when `nightly` has new commits (instead of blindly at 6:00 UTC), and
- Record the Git SHA into the build as part of the version number, so that Help > About Micro-Manager can display the exact commit used in the build. (Consider also adding an API method that returns the SHA.)

(A variant of this approach would be to use tags, such as `nightly-20220214`, instead of the `nightly` branch. But that would result in a lot of tags. It does have the advantage of providing the reverse mapping from code revisions to build dates.)

For anybody looking to download the nightly builds, the only change will be that there won't be a build for every date (which is already the case due to occasional failed builds). I doubt this would disrupt anybody's workflow.

An alternative way might be to do away with "nightly builds" entirely, instead providing "development builds", which could be produced every time code is pushed to `main`. However, I don't feel this is particularly useful to non-developer users -- we would have to add not only the date but also the time to keep the builds in order, and it just makes the whole situation more complex. (We can separately provide builds for every push to `main` for the purpose of developers testing their code; these need not be kept for the long term, e.g. expiring after 14 days.)

Thoughts and feedback appreciated!

henrypinkard:
> One way to do this might be as follows:
> 
[...]

PR #1408, opened 2022-02-17
update pycromanager java

henrypinkard:


marktsuchida:
@henrypinkard Would appreciate it if you could give it some time to propagate before merging these updates. I get alerted for failed builds every time due to the package not yet being available from Maven.

henrypinkard:
I just worry that I'll move onto something else and forget. Do you know if its possible to have the ivy.xml use `Latest` instead of the specific version? That would solve this and prevent me from having to make all these PRs in the first place

henrypinkard:
> now the dependency resolution will succeed but the build may still fail, nondeterministically (e.g. if MMStudio uses a new method added to pycromanager-java). And the only way to determine what happened would now be to inspect the details.

I think this is extremely unlikely, because by design, MMStudio only calls one function of pycro-managerJava, and I have no plans to change this:
https://github.com/micro-manager/micro-manager/blob/4aaf8646688868fca13c80484e22bf37281a3648/mmstudio/src/main/java/org/micromanager/internal/MMStudio.java#L602

And using latest would save me the trouble of having to constantly create PRs just to bump versions, and ensure that I never put in the wrong version, which would make the python half of pycromanager more robust. In fact, while typing this I realized I have the wrong version of all the sub libraries right now.

How about we change to latest and if it ever causes a problem for you we can revert back to the current way?

marktsuchida:
Ok, I see that build errors may not happen often with pycro-manager-java specifically, but they can happen with other dependencies. Even if it doesn't result in errors, it makes the build less reproducible.

Constantly creating PRs for bumping the version is a _good_ practice -- so good that people have automated it (see [Dependabot](https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/about-dependabot-version-updates) -- it doesn't support Ivy but we can use it once we switch to Maven).

Making the builds less self-describing and less reproducible is really not nice. It is in fact a form of technical debt. It can waste people's time when troubleshooting just about any issue. (It's also the opposite of what we were just discussing in #1404.)

henrypinkard:
> it doesn't support Ivy but we can use it once we switch to Maven

Great that'll be a big upgrade.

But I still wish we had a better mechanism that wasn't me setting a timer or relying on you and Nico to merge at the right time. The Forgetful Vulnerability in Henry 30.4.1 is buried deep in the Kernel and I have no idea how to fix it. Is it possible we could add a 15 min delay somewhere in the automated builds that keep causing errors?

marktsuchida:
:) That's an interesting idea. I don't think I can suppress failure notifications, but it looks like I can [automatically retrigger](https://plugins.jenkins.io/naginator/) a failed build. I'd say this is good enough for now: I'd rather see occasional failures than add unnecessary complexity, especially if the errors have a chance to fix themselves. So please keep on with the quick PRs!

henrypinkard:
Ok sounds good, thanks!

Issue #1396, opened 2022-02-10
Save function does not split large stacks intelligently


dickinson-lab:
We have a workflow that involves acquiring 10,000 frames of data, each with 3 fluorescence channels.  Because the stack is larger than 4 GB, it has to be split into multiple .ome.tif files.  However, this splitting seems to be done based on individual 2d images, rather than timepoints or channels.  For example, depending on the exact size of the image, the first file in the dataset might contain frames 1-187 of all three channels, plus channel 1 of frame 188.  The second file then begins with channels 2-3 of frame 188, etc.  

This is a problem because none of the resulting .tiff files contains a full complement of channels for each time point it contains. As a result, the individual files cannot be opened, de-interleaved and displayed correctly by any downstream software, including ImageJ/FIJI (the only exception is MicroManager itself, which opens and reconstructs the entire dataset). 

The same issue applies to datasets with multiple z planes, where (for example) one file might contain half the stack at a given time point, while the rest of the stack is found in the next file.

Possible solutions: 
- Implement a save function that splits the data into files in some principled way: For example, require that no z stack or channel series be split between two files (and reduce the file size as necessary to ensure that this happens).  
- Allow a user (or the callling function of `Store.save()`) to specify the maximum size of the file that can be saved, thereby allowing the user to calculate and set a limit below 4GB that will result in large files being split "evenly" into tiffs that are individually openable. 

I have a slight preference for the second option because it requires fewer up-front assumptions about the nature of the user's dataset, but thoughts and suggestions are welcome.  I'd be glad to help code a solution if the experts can point me to the corner of the source code that actually decides how to split images during the saving (I looked but couldn't find it).

marktsuchida:
Indeed this is a major problem. Related: #889.

Your second option is an interesting idea, because it seems much less daunting than trying to split logically by image coordinate (which may need extending the file format, and in any case requires a fundamental reworking of the saving code). There is a chance it could be unstable if a byte limit is used, though, because the size of frame metadata may vary. Perhaps it could be made to limit the number of planes per file.

Everything having to do with saving in the multi-page TIFF format is in the package [org.micromanager.data.internal.multipagetiff](https://github.com/micro-manager/micro-manager/tree/main/mmstudio/src/main/java/org/micromanager/data/internal/multipagetiff).

[...]

PR #1415, opened 2022-02-22
Modified MultipageTiffWriter to split large stacks more intelligently.

dickinson-lab:
Proposed fix for #1396

### DIFF
--- a/mmstudio/src/main/java/org/micromanager/data/internal/multipagetiff/MultipageTiffWriter.java
+++ b/mmstudio/src/main/java/org/micromanager/data/internal/multipagetiff/MultipageTiffWriter.java
@@ -495,14 +495,15 @@ public final class MultipageTiffWriter {
     * @return true if there is enough space, false otherwise
     */
    public boolean hasSpaceToWrite(Image img, int omeMDLength) {
+      boolean splitByFrame = true;
       PropertyMap mdPmap = ((DefaultMetadata) img.getMetadata()).toPropertyMap();
       int mdLength = NonPropertyMapJSONFormats.metadata().toJSON(mdPmap).length();
       int ifdsize = ENTRIES_PER_IFD * 12 + 4 + 16;
       int channelsLeft = numChannels_ - img.getCoords().getC();
       int slicesLeft = numSlices_ - img.getCoords().getZ();
-      int axis1Size = 1;
-      int axis1ImagesLeft = 1; 
-      int axis2ImagesLeft = 1;
+      int axis1Size;
+      int axis1ImagesLeft; 
+      int axis2ImagesLeft;
       if (orderedAxes_.get(0).equals("z") ) {
           axis1Size = numSlices_;
           axis1ImagesLeft = slicesLeft;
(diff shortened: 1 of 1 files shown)


marktsuchida:
I like this direction a lot -- if it works safely in all cases, it's a beautiful fix!

Does this work even if the 0th element of `orderedAxes_` is neither `Coords.Z_SLICE` (=`"z"`) nor `Coords.CHANNEL`? Or if `orderedAxes_` is empty? What happens if the intended size is set but the axis order is not set? (These situations may not happen with MDA, but could happen if the axis order is set by a script, I think.)

Also, does this work with Tools > Options > Save XY Positions in separate Image Stack Files turned either on or off?

I think the idea of limiting the chucking to slices and channels makes sense, because the other axes (time and position) have the danger of being huge (not fitting in 4 GiB). But what if somebody acquires 4 channels with 50 slices that are 32 MiB each (> 6 GiB of pixels alone)? We probably need a fallback to the current behavior so that there is no chance of falling into an infinite loop of creating new files.

@nicost and @henrypinkard I would also appreciate your thoughts on this.

dickinson-lab:
Thanks Mark, these are very good points.  For the first two, I am pretty sure it should default back to the old behavior, but I will test and make sure.  

The last point is a very good one, I will probably need to add some additional lines of code to deal with the situation where a single timepoint is >4 GB.  Should be pretty quick now that I know what to do.

dickinson-lab:
Ok, I am pretty sure I've addressed all the questions you asked: 

> Does this work even if the 0th element of orderedAxes_ is neither Coords.Z_SLICE (="z") nor Coords.CHANNEL? Or if orderedAxes_ is empty? 

Yes, if orderedAxes_ is empty or if the first axis is neither Z nor channel, it defaults to the original behavior. 

>What happens if the intended size is set but the axis order is not set? (These situations may not happen with MDA, but could happen if the axis order is set by a script, I think.)

There seems to be a default value of orderedAxes_ (which is channel - time - z - position) that is set somewhere else in the code.  This allows the save function to work correctly even if the calling function doesn't specify axis order. 

> Also, does this work with Tools > Options > Save XY Positions in separate Image Stack Files turned either on or off?

Yes.  If "Save XY Positions in separate Image Stack Files" is unchecked, files will not necessarily be split intelligently, in the sense that parts of a single position can end up split between two different files.  I considered adding additional code to keep each position in a single file if possible, but decided against it (if a user wants this behavior, they can just check the option to save each position in its own file." 

> What if somebody acquires 4 channels with 50 slices that are 32 MiB each (> 6 GiB of pixels alone)? We probably need a fallback to the current behavior so that there is no chance of falling into an infinite loop of creating new files.

I have added this in an additional commit, so the possibility of an infinite loop is eliminated now. 

Let me know if there are any outstanding concerns or additional tests you want me to run before merging these changes.

marktsuchida:
Sorry I haven't looked at this yet - will do soon.

marktsuchida:
It looks like the decision to give up due to a single time point not fitting in a single file is made based solely on `frameSize`, whose calculation takes into account the number of channels and slices, but not the number of stage positions, which would be significant in the case where the time series is the outer loop of a multi-position acquisition, and "Save XY Positions in separate Image Stack Files" is disabled.

Also, the `omeMDLength` needs to be accounted for when comparing `frameSize` to `MAX_FILE_SIZE`. (Technically there is also other per-file overhead such as summary metadata, but I expect that is small enough that `extraPadding` should take care of it.)

Sorry it took so long for me to review this -- among other things that happened, I wanted to take a careful look at this because it changes how data is saved (and we don't have good automated tests in this area, making all changes rather scary).

dickinson-lab:
> It looks like the decision to give up due to a single time point not fitting in a single file is made based solely on frameSize, whose calculation takes into account the number of channels and slices, but not the number of stage positions, which would be significant in the case where the time series is the outer loop of a multi-position acquisition, and "Save XY Positions in separate Image Stack Files" is disabled.

Yes, this was intentional.  I'm not using stage positions as a basis for splitting files, since that behavior already exists (the user can just check the "Save XY positions in separate files" option, if desired). 

The purpose of "giving up" when a single time point won't fit is to make sure that we don't fall into an infinite loop of making new files.  There's no danger of that happening when there are too many stage positions in a file, since we don't consider the number of stage positions when deciding whether to create a new file in the first place.   

I tested this using the demo configuration to acquire a dataset with 144 positions, 31 Z slices and 4 channels (8.7 GB) with "positions in separate files" unchecked.  What happens is that stage positions 1-65 end up in the first file, 66-130 in the second file, and the last 14 positions in a the third file.  Each file contains a full complement of slices and channels for the positions it contains, and can be opened individually.  In my opinion, this is the ideal behavior in this situation. 

The reason it works this way is that `filePosition_` is included in the calculation of `size` on lines 531-534 (and on lines 538-540, which are unchanged from the original code).  This ensures that the file can never overflow. 

> Also, the omeMDLength needs to be accounted for when comparing frameSize to MAX_FILE_SIZE.

Yes, good catch.  I'll push a new commit now that adds this (line 524). 

No worries about the delay, I appreciate you taking a careful look. Let me know if you don't agree with my approach to dealing with stage positions, or if you have other concerns.

marktsuchida:
You're right, I got confused about the XY positions.
I think this is now ready!

nicost:
Thanks so much @dickinson-lab  and @marktsuchida.  Great effort and awesome outcome!

dickinson-lab:
Thanks Mark!  I appreciate all your comments and suggestions.  We love Micro-Manager so it feels great to be able to contribute to the project, albeit in a small way.

Issue #1418, opened 2022-02-28
State Device that declares large number of positions results in unwieldy number of Labels


nicost:
Currently, when a state device creates a "State" property with 64 states, for instance using the following code:

```
pAct = new CPropertyAction(this, &DigitalOutputPort::OnState);
CreateIntegerProperty("State", 0, false, pAct);
SetPropertyLimits("State", 0, 63); 
```

the Hardware Configuration Wizard will create 64 labels (org.micromanager.interbal.hcwizrd.LabelsPage), that can be edited by the user.

Labels were meant for things like filter wheels, where each position has a filter with a known name.  However, with digital output ports, labels are rarely useful, and the list of 64 (unused) labels in configuration files with Arduino  have always bothered me.

The Hardware Configuration Wizard can be "tricked" in not creating labels by  returning zero in the State Device API function "GetNumberOfPositions", however doing so certainly is a hack, and any code relying on that function will produce erroneous results.

On the other hand, for a 32 pin digital output port with 2^32 states, the HCW will use up all available memory in the JVM, so letting it create these labels is not workable.  I could declare the device as a generic device, but then will miss out on very useful utilities like the StateDeviceShutter.  

Are there clean ways within the existing API to convince the HCW not to generate labels?  If not, is there a minimal extension to the API that would let us do this cleanly?

marktsuchida:
This was the reason why I artificially limited `DATTLStateDevice` to 8 DA devices, even though it could otherwise apply to more channels.

We can imagine adding a conditional to the HCW to avoid creating labels when there are greater than a certain number (say, 256, to avoid changing behavior for most existing devices) of allowed values (with no change to the API).

If that still creates `Label` lines when resaving the config, maybe we could avoid writing `Label` lines when they are only repeating the default.

There may be an additional need for the Device Property Browser to avoid dropdowns when there are too many options. And the config group editor when creating a single-property group on `Label`.

And any other (e.g. plugin) code that enumerates labels, or enumerates all choices of multiple-choice properties, has the potential to break.

So the reality is that we cannot safely have more than a reasonable number of states, if we care about not breaking things. At least not without going through a lot of trouble.

If there is any way to avoid having a greater-than-8-bit `State` in your device (I presume NIDAQ), I would strongly consider those options.

nicost:
[...]

PR #1419, opened 2022-03-01
Hardware Configuration Wizard: Only show State devices thave have a Label property on the label page.

nicost:
State devices with more than 10 or states become very unwieldy on the Label page of the HCW.  Moreover, many of these (often digital Output) devices really do not need labels.  I run in an issue with a 32 pin output port that needed 2^32 labels, and run out of JVM memroy.  This PR only shows state devices that have the "Label" property defined.  Therefore, not defining it ensures that the user does not need to define labels.  Also, only labels that differ from the default "State-#" will now be written to file (but note that existing "State-#" labels will not be removed from a config file).


marktsuchida:
Link: #1418.

nicost:
Thanks!  Forgot to do that;)

PR #1422, opened 2022-03-02
Studio: Implements DisplayWindow.setFullScreen(bool)

### LINKED ISSUE #1416: mm.live().getDisplay().setFullScreen(true) does not work
I believe that is because the course code in DisplayController.java has the following line of code:

   @Override
   public void setFullScreen(boolean enable) {
      throw new UnsupportedOperationException();
   }

I am not quite sure why this function is in the API but not implemented. The actual code setting fullscreen mode is in org.micromanager.display.internal.displayWindow.DisplayUIController:setFullScreenMode(boolean fullScreen).

nicost:
Closes #1416


nicost:
In addition, ensures that full screen display appears on the screen that has the display window, and ensures that focus is not stolen from a frame that has the focus.

marktsuchida (approved):

PR #1445, opened 2022-05-19
HCW now updates device pre-init properties after each change

### LINKED ISSUE #1442: Have Hardware Configuration Wizard read in pre-init settings after every change
It can be desirable to change the allowed values in a pre-init setting, depending on the choices made in one of them.  Therefore, it would be helpful to re-create the dialog in the HCW after every change made in pre-init settings.  It may be that such behavior is desirable elsewhere in the MM UI, but this is a safe place to start.

nicost:
It is sometimes desirable for a device to change its pre-initialization properties based on the setting of one or more pre-initialization properties.  This PR modifies the HCW so that it writes each pre-init property to the device after each change, and then reads back in all now available pre-init properties.  This may be a problem with devices that take a long time to read in such properties (I am not sure if this a problem in reality or not).  Since this code is needed by anyone configuring Micro-Manager it may need some scrutiny.  I used the demo state device in the branch HCWPropChangeDemoTest of mmCoreAndDevices to test this code.  That branch can be deleted after testing.

Closes #1442.

### DIFF
--- a/mmstudio/src/main/java/org/micromanager/internal/hcwizard/PropertyTableModel.java
+++ b/mmstudio/src/main/java/org/micromanager/internal/hcwizard/PropertyTableModel.java
@@ -81,15 +81,15 @@ class PropertyTableModel extends AbstractTableModel implements MMPropertyTableMo
          devices_[0] = dev;
       }
       
-      model_.dumpComPortsSetupProps(); // >>>>>>>
-      
+      model_.dumpComPortsSetupProps();
+
       ArrayList<PropertyItem> props = new ArrayList<>();
       ArrayList<String> dn = new ArrayList<>();
       for (Device device : devices_) {
          for (int j = 0; j < device.getNumberOfProperties(); j++) {
             PropertyItem p = device.getProperty(j);
             if (mode == PREINIT) {
-               if (!p.readOnly && p.preInit && !device.isSerialPort() && !p.readOnly) {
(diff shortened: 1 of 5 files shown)


mmstudio/src/main/java/org/micromanager/internal/hcwizard/PropertyTableModel.java
```diff
@@ -50,7 +51,9 @@ class PropertyTableModel extends AbstractTableModel implements MMPropertyTableMo
    PropertyItem[] props_;
    String[] devNames_;
    DeviceSetupDlg setupDlg_;
-   
+   CMMCore core_;
+
+   // OK to delete?
```
marktsuchida: If it's not used, it can be deleted (and probably should be).

It also looks like `mode` is never other than `PREINIT`, meaning that `updateValues` could be simplified, but I digress.

marktsuchida (approved):
I don't know this code well enough to catch any problems by just looking at changes, but the basic idea should be safe, especially given that most existing devices that have pre-init properties use the value-stored-in-property-object approach (i.e., no property handler) -- and therefore do not actually do much when a pre-init property is set.

One thing that would be good to check is what happens if setting a property (with a property handler) fails. From the code it looks like the error is ignored (logged) and the table would revert to the current hardware values, which seems good to me.

nicost:
Great point!  Simplified the code and merged.

PR #1451, opened 2022-06-07
Emu plugin: Reformatted to adhere to checkstyle Micro-Manager styling

nicost:
There should be no functional changes.


plugins/Emu/src/main/java/de/embl/rieslab/emu/utils/settings/Setting.java
```diff
@@ -1,141 +1,141 @@
...
+            this.value = value;
+        }
+
+        @Override
+        public String toString() {
+            return value;
+        }
+    }
+
+    ;
```
marktsuchida, line 140: Spurious `;`

marktsuchida (approved):
Meanwhile, it looks like this PR only contains formatting changes and reordering of methods and fields. If that's right, then please go ahead and merge.

nicost:
Nick set up the checkstyle file a while ago (buildscripts/codestyle).  He had this running on github CI (I asked him to switch that off until we are compliant).  I set up InteliJ to use that file and autoformat the EMU code.  I do not know that much about checkstyle, but I assume that it can be used stand-alone, in CI, and at least in some editors?

Before merging, I want to make sure with JDeschamps that we have a plan how to work with the EMU code in the future.

marktsuchida:
I'm aware of that history. The problem with CheckStyle (by itself) is that it cannot format code from the command line. I think the IntelliJ plugin simply imports the CheckStyle settings for IntelliJ's formatter to use. But we can't require everybody to manually format their code in IntelliJ before each commit (at least I don't want to do that). And even if other IDEs and editors support CheckStyle rules, they may not produce the same formatting results. So people constantly reformatting code would produce lots of spurious changes in Git diffs.

I think it will be ideal if we could use clang-format for the formatting part (because of its deterministic results), and CheckStyle for other checks.

nicost:
Does this "https://checkstyle.sourceforge.io/cmdline.html" work?

marktsuchida:
We can use that to run automated checks. But I don't see any option to reformat source code.

nicost:
It looks like this maven plugin can do it: https://docs.openrewrite.org/tutorials/automatically-fix-checkstyle-violations.  I will experiment.

PR #1454, opened 2022-06-10
Better error reporting in ReportingUtils and ProfileAdmin

nicost:
Previously, certain calls to ReportingUtil.showError would lead to null pointer exceptions and no display of the error to the user.  In addition, errors in ProfileAdmin were not shown to the user.  This PR should somewhat improve that situation.  I believe there was an issue related to this, but I can not locate it.


marktsuchida (approved):
Looks good to me!

PR #1462, opened 2022-06-14
Update pycromanager libraries

henrypinkard:


henrypinkard:
Not sure what this failing check is about but I don't think its related to this PR...

marktsuchida:
I think the new versions are not resolved yet, as usual. The errors are slightly confusing because we don't stop the build when there are missing dependencies (something I mean to change, but I need to make it optional so it's complicated).

marktsuchida:
As I expected, it fixed itself.

@henrypinkard I just noticed that all your recent PRs are not based on the tip of `main`, but instead track your own development history (probably what you call `master`). This doesn't result in incorrect code, but permanently leaves an extra parallel branch in the history of our official repo. It makes it more complicated to understand the history when viewing, e.g., with `gitk` (if 5 people did the same thing as you, it would become a complete mess).
It would be great if you could switch to starting topic branches off the official repo's `main`, so that your PRs end up branching off a recent commit, rather than some ancient commit.

marktsuchida:
@henrypinkard Thanks!

PR #1463, opened 2022-06-15
Emu plugin: Adds Window Position and bounds memoryo to most Emu Windows

nicost:


jdeschamps (approved):
Love it! so simple and elegant.

PR #1470, opened 2022-07-11
Update how-to-build.md

RileyMShea:
fixed typo in Ubuntu quickstart.  conda env's have to be activated with `conda`, can't activate with `mamba`.


RileyMShea:
Nevermind I think it works in newer releases of mamba

PR #1471, opened 2022-07-12
Emu: Add laserUI example plugin.

nicost:
This is another simple plugin, that only allows to set the power of an arbitrary number of lasers.  Output is in native units of the device (rather than % in the SimpleUI).  I plan to combine this with a "Linearizer" device in the Utilities Device Adapter collection so that the user can set exact laser output (in mW or other physical units).

I am unsure whether this is the best approach (i.e. adding this to the Emu code base).  I would like this plugin to be automatically added in the Micro-Manager installer, and other approaches seem much more cumbersome, but @jdeschamps, please let me know if this is OK with you.

Also, it is not clear to me how to instantiate multiple EMU panels simultaneously.  Is that possible, and if not, would it be useful to add that capability?


plugins/Emu/src/main/java/de/embl/rieslab/emu/plugin/examples/laserui/LaserPanel.java
```diff
@@ -0,0 +1,378 @@
...
+   }
+
+   @Override
+   protected void propertyhasChanged(String propertyName, String newValue) {
+      /*
+       * This method is called when an UIProperty has changed.
+       * Here we can modify the UI to reflect this change.
+       */
+
+      if (getUIPropertyLabel(POWER_MW).equals(getUIPropertyLabel(POWER_MW))) {
```
jdeschamps: `if (getUIPropertyLabel(POWER_MW).equals(propertyName)) {`

plugins/Emu/src/main/java/de/embl/rieslab/emu/plugin/examples/laserui/LaserPanel.java
```diff
@@ -0,0 +1,378 @@
...
+   }
+
+   @Override
+   protected void addComponentListeners() {
+      /*
+       * In this method we can add Swing actionListeners to the
+       * JComponents or call the methods from SwingUIListeners.
+       */
+
+      // The JSlider will update the UIProperty (percentage) and the JLabel (with a "%" suffix)
```
jdeschamps: `// The JSlider and the JTextField will update the UIProperty, as well as each other`

plugins/Emu/src/main/java/de/embl/rieslab/emu/plugin/examples/laserui/LaserPanel.java
```diff
@@ -0,0 +1,378 @@
...
+               takeAction(this, getUIPropertyLabel(POWER_MW), textField, slider, min, max);
+            }
+         }
+      }
+   }
+
+   @Override
+   protected void parameterhasChanged(String parameterName) {
+      /*
+       * This method is called when a UIProperty has changed.
```
jdeschamps: `* This method is called when a UIParameter has changed.`

I guess the comment error is also present in the other UI example

plugins/Emu/src/main/java/de/embl/rieslab/emu/plugin/examples/laserui/LaserPanel.java
```diff
@@ -0,0 +1,378 @@
...
+   public String getDescription() {
+      /*
+       * Here, we return the description of the ConfigurablePanel,
+       * this description is used to help the user understand how the
+       * panel works.
+       */
+      return
+            "Laser panels control each a single laser and allow for rapid on/off and power "
+                  + "percentage changes. If the laser does not have a power percentage property, "
+                  + "the UI property parameters SLOPE and OFFSET can be used to rescale the "
```
jdeschamps: The description should rather reflect the use of the _"Linearizer device in the Utilities Device Adapter collection"_, and remove mention of the SLOPE and OFFSET since these are not present anymore.

plugins/Emu/src/main/java/de/embl/rieslab/emu/plugin/examples/laserui/LaserPanel.java
nicost: Great catch!

nicost:
Thanks!  I think I resolved all your comments, but will wait merging until I have written the linearizer device, and thought more deeply how to use this plugin with or without the linearizer.

Issue #950, opened 2020-09-17
Crop to ROI behavior is broken with multi-camera utility


shahlab-ucla:
Drawing a rectangle ROI and clicking the crop to ROI button on an image snapped using multi-camera returns an error stating that no ROI was found.

This works as expected with individual cameras.

If a frame is snapped with the multi-camera and then transferred to ImageJ, drawing an ROI and then clicking the crop to ROI button works as it should.

In cases where the the on-the-fly processor pipeline is used, it would be nice if the crop to ROI took any configured transformations into consideration so that the correct region of each sensor is cropped on each camera. The crop seems to occur before images go into the processor pipeline (presumably because cropping is handled by the camera driver?), which results in mismatched ROI's if an off-center one is used when one camera is being rotated and/or mirrored to correspond to the other.

This can be reproduced by adding a second demo camera and the multi camera utility to the demo configuration (see attached).
[MMConfig_demo.cfg.txt](https://github.com/micro-manager/micro-manager/files/5241734/MMConfig_demo.cfg.txt)

marktsuchida:
I can reproduce this. The error message is
```
Error: There is no selection in the image window.
Use the ImageJ rectangle tool to draw the ROI.
```
despite there being a rectangular selection.

marktsuchida:
The "set ROI to center quad" and "clear ROI" buttons work with multi-camera, so the problem is probably with getting the selection from the GUI.

Until this gets fixed, a temporary workaround is to set the ROI from the Script Panel:
```
mmc.setROI(128, 128, 256, 256);    // x, y, width, height
```

shahlab-ucla:
Thanks Mark. Is it possible to crop to ROI from a script as well? I poked around the API but wasn't able to find this.

marktsuchida:
@shahlab-ucla Micro-Manager datasets do not generally allow editing. You would have to either create a new, cropped datastore, or first export to ImageJ.

nicost:
Fixed in PR #1476.  Sorry this took so long!

PR #1476, opened 2022-07-17
Studio: can get and set ROI correctly in Composite images.

nicost:
Closes issue #950. This fix pertains only to the internal MMROIManager.  Instead of relying on ImageJ to get the active ImagePlus, it uses the Studio DisplayManager and a deprecated function to get the ImagePlus.  As long as we use ImageJ behind the scenes, I do not see a way not to use this deprecated function.

### DIFF
--- a/mmstudio/src/main/java/org/micromanager/internal/MMROIManager.java
+++ b/mmstudio/src/main/java/org/micromanager/internal/MMROIManager.java
@@ -32,6 +32,7 @@ import java.util.List;
 import javax.swing.JOptionPane;
 import org.micromanager.data.Image;
 import org.micromanager.display.DataViewer;
+import org.micromanager.display.DisplayWindow;
 import org.micromanager.internal.utils.ReportingUtils;
 
 /**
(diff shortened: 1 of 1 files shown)


mmstudio/src/main/java/org/micromanager/internal/MMROIManager.java
```diff
@@ -49,7 +50,15 @@ public MMROIManager(MMStudio studio) {
     * Enquires with the UI what ROI(s) are set, and send these to the camera.
     */
    public void setROI() {
-      ImagePlus curImage = WindowManager.getCurrentImage();
+      DataViewer dv = studio_.displays().getActiveDataViewer();
+      DisplayWindow dw = null;
+      if (dv instanceof DisplayWindow) {
+         dw = (DisplayWindow) studio_.displays().getActiveDataViewer();
```
marktsuchida: One ends up pointing out the small things in small PRs :)
But this could be `dw = (DisplayWindow) dv;`

mmstudio/src/main/java/org/micromanager/internal/MMROIManager.java
```diff
@@ -145,7 +154,15 @@ private Rectangle updateROI(Roi roi) {
...
+      }
+      ImagePlus curImage = null;
+      if (dw != null) {
+         curImage = dw.getImagePlus();
+      }
```
marktsuchida: Would be nice to factor this out as a private method, since it is an exact duplicate of the other change.

marktsuchida (commented):
Just minor comments!

mmstudio/src/main/java/org/micromanager/internal/MMROIManager.java
nicost: Perfect!

mmstudio/src/main/java/org/micromanager/internal/MMROIManager.java
nicost: Fully agreed.  All done.

PR #1484, opened 2022-07-30
Magellan: Enable multi-select in copying positions, and checsktyle

nicost:
Please have a look at the second commit.  I can take it out of the PR if it interferes with your workflow.  That second commit makes the GUI class pass the Checkstyle checks that we started running on the main code (Nick started it;), eventually resulting in more uniform code, which seems a good thing.


henrypinkard (approved):
Yes, seems like a good idea. Regretfully, this was all made in the netbeans GUI builder. Not sure its worth the effort to update it to MigLayout, given how long it has been unchanged for.

PR #1469, opened 2022-07-08
Linting

marktsuchida:
For #1460.

So far, this brings the MMStudio code to a state where Checkstyle passes. Some Javadoc checks have been ~turned off temporarily~ turned into warnings.

A couple of fixes were done on the whole repo, not just MMStudio: eliminating CRLF and non-ASCII characters.

To be done before this is ready:
- [x] Add a pre-commit hook to check MMStudio
- [x] Document how to use the pre-commit hook (and point to IDE plugins)
- [x] Add a CI job to perform the same checks as the pre-commit hook
- [x] Turn back on Javadoc checks as warnings in a manner that can be suppressed in the pre-commit hook and CI (but show up in IDEs)

PR #1496, opened 2022-08-23
Ndtiff datastore

henrypinkard:
This PR adds a compatibility layer between NDTiff and the Datastore API. There were minimal changes to the studio code, aside from duplicating things already there for the other storage classes

I'm not sure why a couple of the files showed up that they were entirely changed, but not all files...strange.

In any case, I'm going to merge this now so it doesn't screw up synchronization with the pycro-manager changes also needed to make it possible


marktsuchida:
Please don't merge important changes like this without giving us a chance to review.

henrypinkard:
As I mentioned above, I needed to make an exception to that general policy so as to not introduce problems with the other projects related to this due to the fragmented nature of the way things are currently built. I took great care to structure this is such a way to avoid any possibility of breaking anything in the studio. You are welcome to review and suggest further changes.

marktsuchida:
As it turns out, I'm okay with the changes you made here, although there is unfinished work (I'll create a separate issue).

But synchronizing with other projects is not an excuse. If multiple projects need to be updated in lockstep, the problem is with the interface design or update policy of those projects. If you were not the only person requiring lockstep updates, how would we even deal with such a situation? I think you need to come up with a way to allow updates to happen at different timings. This is what APIs and modularity are for. Let me know how I can help.

> I'm not sure why a couple of the files showed up that they were entirely changed, but not all files...strange.

This is because you changed the line endings from LFs to CRLFs. There must be something wrong with your git config (`core.autocrlf` should be `true` on Windows), or (more likely) you branched off before I fixed all the newlines in this repo (https://github.com/micro-manager/micro-manager/pull/1469/commits/f4d24e2b1225228509b2291ebc6b1bdfca001249, part of #1469).

No big deal; please fix those files in a separate PR. I'll make a note to enforce line endings in the CI check so that we catch it before merging in the future (#1497).

henrypinkard:
> But synchronizing with other projects is not an excuse. If multiple projects need to be updated in lockstep, the problem is with the interface design or update policy of those projects. If you were not the only person requiring lockstep updates, how would we even deal with such a situation? I think you need to come up with a way to allow updates to happen at different timings. This is what APIs and modularity are for. Let me know how I can help.

Yes, agreed. I think I will need your help with this as it seems to me tightly integrated with the build system changes

henrypinkard:
Thanks!

Issue #1497, opened 2022-08-23
CI: lint job should flag files with CRLF


marktsuchida:
Would be good to preserve the clean state achieved after #1469.

Cannot use pre-commit for this, since it will break on Windows.

marktsuchida:
Consider using [`.gitattributes`](https://www.aleksandrhovhannisyan.com/blog/crlf-vs-lf-normalizing-line-endings-in-git/) and checking in CI that renormalizing does not change anything.

Also
- [ ] Fix the few files that got CRLF in #1496.
- [ ] Convert files in `bindist/` to also use LF (these were left unchanged in #1469 because I thought they may need CRLF to work correctly, but that is actually immaterial because they will be checked out with CRLF when building)

PR #1500, opened 2022-08-23
Set newline handling in .gitattributes

marktsuchida:
Part of #1497 (no CI check yet).

Make Git check in with LF newlines for all text files.
Apply to existing files that had CRLF; now all text files have LF in the repo.


marktsuchida:
Merging despite lint failure because failures are not from this PR.

PR #1511, opened 2022-09-09
fix NDTiff bugs for multiple axes

henrypinkard:


marktsuchida:
Would appreciate it if we could avoid leaving commented code. If it needs to be restored later, it should be restored from the Git history. If the fact that the code was there in the past is important to future readers of the code, a comment to that effect can be left.
Also would appreciate it if you could wait for the checks to pass before merging. Thanks.

PR #1512, opened 2022-09-10
Changes the default setting in MDA dialog for Z-Stack from absolute to relative

nicost:
This one trips me up every time.  I always use relative Z.  Absolute Z may be dangerous under certain conditions, so relative Z will always be safer.  This should not change anything for people who are already using absolute Z in the MDA, only the default for new users.


marktsuchida (approved):
Could not agree more!

PR #1514, opened 2022-09-10
Linting Studio, Magellan, Gaussian and Image Processing

nicost:
Only linted errors.  Still some to go, but this makes nice headway.


marktsuchida:
Thanks. If you (possibly in the future) bring a project (plugin, library) to a state of no errors, you can change the regular expression in `.pre-commit-config.yaml` to include those projects (currently only `mmstudio` is included). This will enable checks by the pre-commit hook, as well as the "lint" check on pull requests. If you feel like it.

Issue #1515, opened 2022-09-13
Ratio Imaging plugin


SaeedArshad194:
I am using two cameras to measure FRET, Nico said  the problem is that the plugin only lists the channels in the Channel group, and not the “composite” channel names you get with the MultiCamera plugin. Who can fix the bug?

nicost:
Should work after merge of #1519.  Please test and provide feedback.  I believe this already is in the latest nightly build. Now waiting for someone wanting to do this with the SplitView plugin.

nicost:
Closing due to lack of feedback.

PR #1519, opened 2022-09-19
Ratio plugin: Added support for ratio images from multiple cameras.

nicost:
Needed a way to specify the special channel names generated when using multiple cameras.  There is quite a bit of heuristics involved, but it seems to work.


nicost:
Closes issue #1515.

PR #1525, opened 2022-09-26
Linting plugins: MMClearVolume, Magellan, libraries: ImaegProcessing, Gaussian

nicost:
All but two files in the libraries-Gaussian are now error free in Checkstyle.  I am not quite sure how to add the linted directories to .pre-commit-config.yaml (and possibly elsewhere?).

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -10,4 +10,5 @@ repos:
           - --
           - -c
           - buildscripts/codestyle/checkstyle.xml
-        files: '^mmstudio/src/main/java/'
+        files: '^(mmstudio/src/main/java/|libraries|plugins/MMClearVolumePlugin|plugins/Magellan|plugins/SharpnessInspector)'
+        exclude: '^(libraries/Gaussian/src/main/java/edu/ucsf/valelab/tsf)'
--- a/libraries/Big/src/main/java/edu/umassmed/big/I2I.java
+++ b/libraries/Big/src/main/java/edu/umassmed/big/I2I.java
@@ -16,12 +16,19 @@ import java.io.RandomAccessFile;
 
 public class I2I {
 
-   private String history[] = new String[15];
-   private String i2i_fileName;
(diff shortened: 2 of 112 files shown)


marktsuchida:
In `.pre-commit-config.yaml`, the (currently) last line
```
        files: '^mmstudio/src/main/java/'
```
specifies that Checkstyle is only run on files matching the regular expression `^mmstudio/src/main/java/`.

So changing it to
```
        files: '^(mmstudio/src/main/java/|libraries/Gaussian/src/main/java/)'
```
should do the job. (The `src/main/java/` part may not be needed for libraries and plugins, unlike mmstudio. Non-Java files are excluded by a separate mechanism.)

If you want to exclude particular files, it is easier to add an `exclude:` field (at the same indentation as `files:`)

Further details are in the [pre-commit docs](https://pre-commit.com/#pre-commit-configyaml---hooks). You can test (after `git add .pre-commit-config.yaml`) with `pre-commit run --all-files`.

No need to add it anywhere else.

nicost:
Still about 1800 errors to go in plugins...  We'll get there;)

marktsuchida:
@nicost I think this broke the build; can you take a look?

```text
 [mm-javac] /Users/mark/devel/micro-manager/plugins/ChannelCorrector/src/main/java/org/micromanager/channelcorrector/ChannelCorrectorPanel.java:232: error: cannot find symbol
 [mm-javac]       Polygon maxima = FindLocalMaxima.FindMax(siPlus, distance_, threshold_,
 [mm-javac]                                       ^
 [mm-javac]   symbol:   method FindMax(ImagePlus,int,int,FilterType)
 [mm-javac]   location: class FindLocalMaxima
```

nicost:
Sorry about that!  Did not have that plugin in my InteliJ project.  Fixed in PR #1527.

marktsuchida:
Thanks! I found a few more of the same but fixed.

PR #1529, opened 2022-10-10
Avoid moving C++ build dir in Windows Ant build

marktsuchida:
Let C++ components, including SWIG-generated sources, build where they normally do when built from Visual Studio, instead of forcing to `build/` outside of `mmCoreAndDevices/`.

This means that `ant jar` in `mmCoreAndDevices/MMCoreJ_wrap` should work after building `MMCoreJ_wrap` in Visual Studio.

Added Ant target `clean-cpp`. Fixed all clean targets.

Closes micro-manager/mmCoreAndDevices#255.


henrypinkard:
Thanks @marktsuchida !

PR #1532, opened 2022-10-23
Emu: Prevent exception when using WindowBuilder

jdeschamps:
A recent commit (https://github.com/micro-manager/micro-manager/commit/92641596a6fb5f1022907251bdef5f23b70ef811) leads to an error when trying to use Emu outside of MM, for instance with Eclipse WindowBuilder ([example on image.sc](https://forum.image.sc/t/error-creating-configurablemainframe-emu-micromanager/72665/6)).

This fix catches the exception and solves the issue.


marktsuchida:
Thanks, @jdeschamps.

I think this is a good opportunity to fix the issue at its source, so that we don't need to add try-catch around every call to `setUpBoundsMemory()`. I created #1533.
Can you see if that addresses this issue?

jdeschamps:
Worked like a charm! Thanks!

PR #1537, opened 2022-10-28
Remove Boost from toplevel configure.ac

marktsuchida:
We don't need it any more, because MMCore no longer depends on Boost. The DeviceAdapters configure.ac still configures Boost for the device adapters that require it.


henrypinkard:
Nice!

Issue #1509, opened 2022-09-06
Fix or disable code formatting check


marktsuchida:
See discussion in #1501 regarding lack of way to format with IntelliJ on macOS. (@henrypinkard raised the issue again in #1507.)

What would solve all our woes is having an autoformatter _that can be invoked from the pre-commit hook_, so that the code is always formatted correctly whether the IDE is configured or not (I tried to bring this up when we first started looking at Checkstyle, but probably failed to emphasize how important it could be). Unfortunately, I looked into this earlier this year and did not find an immediately attractive solution:
- google-java-format is not suitable for our project because it only allow their specific formatting style.
- Eclipse's formatter is well regarded and some other IDEs (including IntelliJ) have plugins for it. But it cannot be run from the command line (note 1).
- clang-format supports Java, but didn't seem to be as "good" as the IDE formatters (such as IntelliJ). Also it is not as widely supported by IDEs.

note 1: There is a Maven plugin for Eclipse's formatter. Maybe we could hack something that generates a temporary pom.xml just for the formatting, but I don't know how easy that is. We need to limit it to the changed files (otherwise long waits on every commit), which doesn't seem easy.

In any case, we'd need to come up with a style definition for whatever formatter we choose that is compatible with our Checkstyle style -- or remove the formatting-related rules from the latter.

So with that option not looking easy, what shall we do for now?
- We could turn off the check, but then somebody has to fix things later somehow (by re-saving the files in a working IntelliJ or by waiting until we have an autoformatter).
- For IntelliJ on macOS, we could manually configure its coding style until it is close enough, so that if any manual fixes are needed they are few enough to be easy.
- Can IntelliJ export its own style rules? If so, we could export the style from Windows IntelliJ (after loading the Checkstyle style) and import it on macOS, hopefully without the kind of error seen in #1501. The exported style could be saved in `buildscripts/codestyle` and we make it a rule to always (manually) update it when the Checkstyle definitions change (which should be rare).

marktsuchida:
Just discovered a tool that can fix (at least some) Checkstyle violations automatically:
https://docs.openrewrite.org/tutorials/running-recipes/automatically-fix-checkstyle-violations
Unfortunately (as is all too common with Java tools) it cannot be run from the command line, and only (practically) works with Maven or Gradle. So something to set up once our Java projects are mavenized.

marktsuchida:
[...]

Issue #1501, opened 2022-08-23
Checkstyle failures


marktsuchida:
Checkstyle is finding style issues with the changes made in #1496.
@henrypinkard Would you mind fixing these?
See [doc/pre-commit.md](https://github.com/micro-manager/micro-manager/blob/main/doc/pre-commit.md) for details.

henrypinkard:
There seems to be a problem with the instructions for setting this up. When I go to do the final step

> Additionally, import the Checkstyle rules into code formatting settings: go to Settings > Editor > Code Style > Java; from the Gear (⚙️) menu, choose Import Scheme > Checkstyle Configuration.

I get this error:
![Screen Shot 2022-08-23 at 10 04 47 AM](https://user-images.githubusercontent.com/7969470/186221425-4a39a03b-d36a-4993-bb5c-e77a7d6bd23a.png)

Maybe this is related to problems in the codestyle file? ([stakcoverflow](https://stackoverflow.com/questions/57037571/maven-checkstyle-plugin-caused-by-com-puppycrawl-tools-checkstyle-api-checksty))

Latest IntelliJ on Mac

Thoughts?

marktsuchida:
Checkstyle 10.3.1 or 10.3.2 from the command line (or pre-commit hook) is working (including on macOS), so the problem is specific to the IDEA plugin and unrelated to the StackOverflow question you linked.

But not sure at the moment how to solve this....

marktsuchida:
I can reproduce it on macOS. Previously I had only tried on Windows.

nicost:
I ran into this previously (a few days ago) on my laptop whereas it had worked previously, and failed to fix it.  However, I just installed the latest CheckStyle plugin on my desktop, made sure in CheckStyle settings that version 10.3.2 is selected, and then run it without issues.  I'll check again on my laptop...

marktsuchida:
It turns out that a more detailed exception log can be [viewed](https://intellij-support.jetbrains.com/hc/en-us/articles/207241085-Locating-IDE-log-files).

<details><summary>full stack trace</summary>

```
2022-08-23 12:49:10,816 [1037797]   WARN - #o.i.i.c.i.CheckStyleCodeStyleImporter - Failed to import style
org.infernus.idea.checkstyle.exception.CheckstyleServiceException: Error executing command 'OpLoadConfiguration': internal error - Failed to call com.puppycrawl.tools.checkstyle.ConfigurationLoader.loadConfiguration()
[...]

PR #1542, opened 2022-11-05
upates for cleaned up AcqEngJ API: Magellan and bumped library versions

henrypinkard:


marktsuchida:
@henrypinkard I'm just going to keep repeating this.... Please don't merge pull requests that don't pass the tests.

If Checkstyle is more a hinderance than help, we can decide to disable it (e.g. for Magellan). But it's not okay to just merge things that don't pass CI, because then we just keep getting an ❌ for every subsequent commit, and the CI becomes pointless for everybody.

nicost:
Fixed in #1546.  Next time, just ping me Henry, and I'll fix them before merging.  All very simple, and helpful to make the code more readable, such as: maximum line length of 100, no empty lines in imports, alphabetic ordering of imports.

henrypinkard:
@marktsuchida Sorry, I (once again) made the mistake of pushing a new PM version that needed to be synced on this side, so I merged it to prevent cryptic errors for users

I do think it is worth considering disabling checkstyling for the time being, until we have have a super easy IDE-integrated solution or some kind of bot that takes care of it. I am personally neutral as to whether things are consistently formatted, though I certainly understand the advantages for others. However, I find it very frustrating that a required system that is meant to make things easier is currently creating _more_ work, whether it is manually doing it, investing a bunch of time diving into the surprisingly complex details of how to do it semi-automatically, or the de facto current strategy of @nicost cleaning up my repeated code messes (thank you Nico!).

How about if we disable it as one of the automatic checks until https://github.com/micro-manager/micro-manager/issues/1509 is resolved?

marktsuchida:
@henrypinkard I think another underlying problem is the need to merge a particular PR at a particular time. I think this suggests that the components are overly coupled. We need to come up with a way to manage version updates that does not require such synchronization, because it is unscalable. Perhaps we should discuss the next time we're on Zoom.

Regarding disabling Checkstyle checks: I agree to discuss this possibility further (and have been a little uneasy from the start about enabling checks without having automated fixes where possible), but I would like you to enable the pre-commit hooks in your development workflow before we even go there. Even if your IDE didn't help, you should have gotten errors when committing (before pushing, that is) if you had that set up. And you would have seen that the majority of errors were about line length, which can be configured in your IDE project even if auto-import of the Checkstyle rules doesn't work. The same also applies to import-related rules. After those, there would have been only 3 issues remaining, all of which are sensible and easy enough to manually fix ("All overloaded methods should be placed next to each other" and "Each variable declaration must be in its own statement"). Some of these are probably ones that IDEs can't auto-fix anyway.

Until you actually try following our current procedure, I'm not sure we can have a fair discussion about how much "more" work this is creating. (For example, we have the ability to disable certain rules if they are too cumbersome and not helpful, but without actually looking at what is happening, our choices are only to blanket enable or disable.)

henrypinkard:
> @henrypinkard I think another underlying problem is the need to merge a particular PR at a particular time. I think this suggests that the components are overly coupled. We need to come up with a way to manage version updates that does not require such synchronization, because it is unscalable. Perhaps we should discuss the next time we're on Zoom.

For sure. We touched on this briefly before, and I think there's room for improvement (more interpretable errors), but I think a more sustainable fix may need to wait for the overhaul of the Java build system. Though if you have other ideas next time we zoom I'm all ears!

> Regarding disabling Checkstyle checks: I agree to discuss this possibility further (and have been a little uneasy from the start about enabling checks without having automated fixes where possible), but I would like you to enable the pre-commit hooks in your development workflow before we even go there. Even if your IDE didn't help, you should have gotten errors when committing (before pushing, that is) if you had that set up. And you would have seen that the majority of errors were about line length, which can be configured in your IDE project even if auto-import of the Checkstyle rules doesn't work. The same also applies to import-related rules. After those, there would have been only 3 issues remaining, all of which are sensible and easy enough to manually fix ("All overloaded methods should be placed next to each other" and "Each variable declaration must be in its own statement"). Some of these are probably ones that IDEs can't auto-fix anyway.
> 
> Until you actually try following our current procedure, I'm not sure we can have a fair discussion about how much "more" work this is creating. (For example, we have the ability to disable certain rules if they are too cumbersome and not helpful, but without actually looking at what is happening, our choices are only to blanket enable or disable.)

@marktsuchida I tried this before (https://github.com/micro-manager/micro-manager/issues/1501), became very confused, and it did not end up working. Hence: https://github.com/micro-manager/micro-manager/issues/1509. Are you suggesting something different?

I want to emphasize how incredibly frustrating I find this system. Manually perfecting code format (though it may seem trivial to you) is a major pet peeve of mine. I really don't understand at all what you're talking about in https://github.com/micro-manager/micro-manager/issues/1509, and I'm very much not keen on allocating more time in to learning and setting up more stuff related to code formatting, when they are so many more pressing and important issues to tackle. If I'd realized what this system made and we'd had our voting system worked out at the time, I would have raised major objections to implementing such a system. Is it really so much to ask to just disable this system for now? All the code can be manually formatted and maybe auto formatted when later when https://github.com/micro-manager/micro-manager/issues/1509 is worked out (probably more efficient to batch this type of thing anyway).

nicost:
> If I'd realized what this system made and we'd had our voting system worked out at the time, I would have raised major objections to implementing such a system. Is it really so much to ask to just disable this system for now? 

Just a bit of history here:  Checkstyle was introduced by Nick Anthony (https://github.com/micro-manager/micro-manager/pull/1206), first as a CI that basically showed that no PR passed.  Nick clearly felt the need for a bit more organized code formatting than we had before.  I concurred with him that it is nice to have more easily readable and more uniformly formatted code, and worked with him to set up a style sheet that more or less adhered to the style we had.  I then put quite a bit of effort into cleaning up mmstudio and some major plugins (see: https://github.com/micro-manager/micro-manager/pulls?q=is%3Apr+is%3Aclosed+Checkstyle).  I am hoping this pays off and the code is more accessible to new comers.  Again, I am happy to clean up your PRs (does not take me more than a few minutes), but it would be quite sad to see all those hours of work I have put into this in the past go out of the window.

henrypinkard:
That is totally fair, and I really do appreciate that. Maybe that is the best way forward for now

marktsuchida:
> @marktsuchida I tried this before (#1501), became very confused, and it did not end up working. Hence: #1509. Are you suggesting something different?

Perhaps. My understanding is that you tried the IDE setup part. The other part (which is more fundamental) is setting up the git pre-commit hook. Did you try that?

marktsuchida:
See #1548 for a possible part of a solution (IDEA settings). The pre-commit hook is still needed to catch style issues other than formatting, though.

henrypinkard:
It seems to me that while a consistent code formatting certainly cannot hurt in making MM development more accessible to newcomers, adding the friction of having to configure more settings in order to make contributions that pass checks (or being deprived of the dopamine hit of all green check marks) might very well do the opposite. I've always tended to think that getting everything configured in the right way before the actual coding can begin is one of the most intimidating/difficult parts of getting involved in a project, and I think it would be advantageous to have that process be as smooth and simple as possible. @nicost if you're going to do this for me (which, again, thank you) and potentially others, would it not make more sense to simply have this be something that you do from time-to-time unrelated to the continuous integration? Seems like this would get the best of both worlds: we wouldn't sacrifice the benefits of having all the code cleaned up, and there's no duplication of effort from many people having to go through a confusing configuration process

marktsuchida:
It's a tradeoff. Either we ask everybody to do a little work to conform to standards (I do agree that it needs to be "a little"), or the few core maintainers need to do endless janitorial work (or watch the codebase degrade). Nothing wrong with janitorial work (and it will never be completely eliminated), but as you said, there are more important things to spend time on.

And it scales better when more (if not all) contributors are able to submit clean code from the beginning by enabling automatic checks -- this does not conflict with helping along new people whose PRs may not initially have everything perfect, through a combination of pointing them to the rules/setup and helping them fix their PR (or, in some cases, doing it for them).

Also note that uniform code formatting is most beneficial when applied to each and every commit, because that way diffs are always as clean as possible and free of spurious changes due to formatting. The best way I know of to build this into our workflow is pre-commit hooks, backed up with automated PR checks that will catch cases where the hooks were't configured. If there is periodic reformatting, it adds a lot of noise to the change history (some might view this as worse than not reformatting, and I find it hard to disagree). And if any part of the fixing (which, again, is not limited to pure formatting only) is done manually, there is the risk of introducing bugs, which is bad if done long after the code was merged. If the checks happen at the source, the version that gets tested is the checked version so such risk is minimized.

Why is trying to improve the configuration process (and its documentation) so that it is less confusing not an option here? Do we have to give up on pre-commit if the first attempt was not perfect for everybody? I'm not sure how to get to a better configuration experience without constructive feedback, but I will emphasize that the way things are currently is not the way they have to be forever unless disabled.

I think it would help to separate the issues specific to our current Checkstyle configuration (remember we have full control over all the rules; we can disable unhelpful ones) and issues of setting up the pre-commit hook itself. Personally, I don't think it is unreasonable to ask people to enable pre-commit (I see other projects using it quite successfully), but I do see room for improvement with the actual Checkstyle checks we currently run (in those other (typically Python) projects, the workflow is not nearly as complicated).

Issue #1465, opened 2022-06-17
Make AcqJ available to MMStudio


henrypinkard:
[AcqEngJ](https://github.com/micro-manager/AcqEngJ) is more performant, flexible, and easier to change/improve than the current Clojure acquisition engine used by MMStudio. It is already used my Pycro-manager and Micro-Magellan. This issue describes adding it a back end to the Multi-dimensional Acquisition system in the Micro-Manager GUI, for now as an option in parallel with the current Clojure engine, but with the aim of eventually replacing and retiring that engine.

Doing so will provide many many immediate benefits, owing to shortcomings in the existing  Clojure acquisition engine and saving formats. Many of these limitations pointed out by @marktsuchida in https://github.com/micro-manager/mmCoreAndDevices/issues/180:

- Get rid of special treatment of position/time/Z indices. These can still be used, but aren't baked into the acquisition engine in the same way, making it easier to do more general experiments that vary along different axes.
- No circular dependencies on MMStudio. In fact, no dependencies on MMStudio at all (AcqEngJ sits above MMCoreJ only)
- The ability to use "Acquisition hooks" at various stages of the acquisition cycle, which enable injection of customized code that runs synchronously with the acquisition engine. More info on those [here](https://pycro-manager.readthedocs.io/en/latest/acq_hooks.html)
- A full syntax for instructions (acquisition events) that can be spelled out in JSON. This means experiments can be saved/loaded from files, or generated from GUI as it is now
- More transparent hardware sequencing behavior. Events get combined and can be printed to JSON before/during dispatch

Since AcqEngJ is much more flexible than the current engine, much of this will be straightforward and just require translation layers between our current ways of storing acquisition settings and the way AcqEngJ describes experiments (more on this below). One thing that comes to mind that still needs to be sorted out is:

- **Autofocus** is not built in an any way to AcqEngJ right now. Currently, in pycromanager people just do it through acquisition hooks. There will need to be some translation of the MDA settings for autofocus to make this work

#### Additional features that could be easily added
- More fine grained control over sequencing--e.g. run sequencing over z-stack, but not over channels
- Possible performance improvements. @marktsuchida mentioned `lazy sequence generation is inefficient`. Wondering if this is specific to Clojure engine or something more general
-  Better integration of [Pycro-Manager](https://github.com/micro-manager/pycro-manager), such that arbitrary python code could be injected into the acquisition cycle to synchronize hardware, or used to modify/divert images. 

### Some details on how to do it
[...]

PR #1547, opened 2022-11-07
Incorporate AcqEngJ into MMStudio

henrypinkard:
https://github.com/micro-manager/micro-manager/issues/1465

I've got a preliminary version working and am able to run some basic types of MDAs through the GUI. There's a new option to turn on and off AcqEngJ backend. This shouldn't affect anything the clojure engine is doing, so I would think it can be merged in not too long and tested and improved from there. However, it would be helpful if you could have a look @nicost and @marktsuchida to ensure that I haven't accidentally introduced unintended consequences for those that haven't checked the box.

There are corresponding changes to AcqEngJ and some other deps that should be merged simultaneously with this.

### Some implementation questions:
- [ ]  AcqEngJ has multiple places to hook in code to run either on the thread where acquisition events come from or on the acquistion thread itself (before the stage/channels get set, after they get set but before the camera is activated, or after the camera is activated). Any idea which of these time points the `attachRunnable` interface should be mapped to?
- [ ]  What is supposed to be happening with autoshutter--I've always been confused about this
- [ ] The position list: is the acquisition engine supposed to move multiple XY stages if they're both defined in the position list, or just the main one? 
- [ ] Should Z positions in the position list have any effect?
- [ ]  Autofocus: I haven't attempted to put this in yet. What is the correct behavior? I'm not especially familiar with how AF in MM is supposed to work
- [ ]  Any knowledge of what metadata tags are required? The metadata requirements are a huge mess as there are many hidden tags required in different parts of the software and as far as I know the minimal set is not documented anywhere. I remember some of the requirements, and the rest I've been slowly figuring out empirically by copying the clojure engine's output (which is hard, because there don't seem to be any exceptions when they're missing). Any additional knowledge would help this process
- [ ]  any objections to me deleting the [Clojure stuff](https://micro-manager.org/Clojure) on the website now? And maybe cleaning up [this page ](https://micro-manager.org/Multi-Dimensional_Acquisition_Programming) on MDA programming which appears to be outdated?


marktsuchida:
I'll try to look at the code in the coming days, but here are quick answers to your questions:

> * [ ]   AcqEngJ has multiple places to hook in code to run either on the thread where acquisition events come from or on the acquistion thread itself (before the stage/channels get set, after they get set but before the camera is activated, or after the camera is activated). Any idea which of these time points the `attachRunnable` interface should be mapped to?

The Clojure engine calls the runnable before opening the shutter for a snap, after most other changes have finished. However, I vaguely remember that it can overlap with something. Only the code has the true answer.

> * [ ]   What is supposed to be happening with autoshutter--I've always been confused about this

Autoshutter means that the current `Core`-`Shutter` is opened before a snap or sequence, and is closed when the snap or sequence finishes. If autoshutter is off, `Core`-`Shutter` is ignored. (Not sure what your question is about?)

> * [ ]  The position list: is the acquisition engine supposed to move multiple XY stages if they're both defined in the position list, or just the main one?

I think so, and also for multiple Z stages. This should be easy to test.

> * [ ]  Should Z positions in the position list have any effect?

Yes. They specify the Z position of the acquisition for that position. Aside from manually focusing on different positions, this is also used by the 3-point focus plane generated by the HCS site generator.

If autofocus is enabled (at least under certain conditions), I believe they specify the initial starting point for focusing. When a second image is taken at the same XY position, the computed focus from the previous round is used as the starting point (this Z position memory is internal to the acqEngine).

> * [ ]   Autofocus: I haven't attempted to put this in yet. What is the correct behavior? I'm not especially familiar with how AF in MM is supposed to work

It's very complicated with lots of case logic, and I don't think there is much hope of correctly replicating it without studying the existing code. Yet, changing the behavior without notice would break a lot of people's workflow. I can try and help (e.g. happy to study the code together), but I can't answer this off the top of my head.

This is probably the bulk of the hard-coded complexity of acqEngine aside from hardware triggering, but I don't think it is an unsolvable problem. acqEngine actually hard-codes behavior that is not always the most desirable, so having the new engine be capable of a superset of these behaviors (which I guess is automatically true if autofocus is implemented using its hooks) might make sense.

> * [ ]   Any knowledge of what metadata tags are required? The metadata requirements are a huge mess as there are many hidden tags required in different parts of the software and as far as I know the minimal set is not documented anywhere. I remember some of the requirements, and the rest I've been slowly figuring out empirically by copying the clojure engine's output (which is hard, because there don't seem to be any exceptions when they're missing). Any additional knowledge would help this process

Yes.

Okay, I'm kidding. But the problem is that nothing is truly known other than that the current code (more or less) works. It would be great to come up with a specification, and we really should, but the only way to do this is to patiently study where each tag is generated and consumed (note that this differs from empirical determination, which fails to find all tags that are conditionally present).

I started doing this when I created the `PropertyKey` enum class. My vision at the time was to put all knowledge about each known tag into this class in a programmatically accessible way: for example, so that you could ask for the set of required and optional tags for summary metadata. This was to include dealing with variant spellings, too (see `INTERVAL_MS`, sigh). But unfortunately I did not finish that work. Happy to collaborate on restarting this work (or some alternative), but I do have to disappoint you if you were hoping for a ready-made list of required keys.

> * [ ]   any objections to me deleting the [Clojure stuff](https://micro-manager.org/Clojure) on the website now? And maybe cleaning up [this page ](https://micro-manager.org/Multi-Dimensional_Acquisition_Programming) on MDA programming which appears to be outdated?

None whatsoever!

henrypinkard:
Thanks! I think this gives me enough to move forward

> Yes.
> 
> Okay, I'm kidding. But the problem is that nothing is truly known other than that the current code (more or less) works. It would be great to come up with a specification, and we really should, but the only way to do this is to patiently study where each tag is generated and consumed 

😂 You had my hopes up there for a moment

marktsuchida:
> You had my hopes up there for a moment

Well the literal answer to each sentence in your question was "yes". Especially the second sentence....

nicost:
B.t.w., I am unable to compile or test your branch, presumably since some dependencies are updates too.  Is it possible to update the buildscripts so that I can use the correct dependencies, or do I have to set up a super project that includes the code to all?

nicost:
I'd first like to test if the existing Acquisition engine still works as expected.  Are the changed dependencies backwards compatible?  In that case, it would be fine to add those to the main branch.  I would not want to merge this branch until we have ensured it does not break things (and until we cleaned up the code).  

B.t.w., PycroManager 0.28 does not download for me.  Wait a bit?

henrypinkard:
That makes sense.

No, its not backwards compatible.

PM 0.28 didn't build because I pushed it too soon after an updated AcqEngJ (this system is quite finicky). But I just re ran it so maybe in a couple mins: https://github.com/micro-manager/pycro-manager/actions/runs/3430886056

henrypinkard:
Should be released now

marktsuchida:
I'm not understanding why this is ready to be merged if one must fear for objectives crashing -- unless prominently marked as experimental and dangerous.

I'd like to review this in detail so I understand the consequences. Can I have a couple days before we go forward?

henrypinkard:
fine by me

henrypinkard:
> unless prominently marked as experimental and dangerous.

It is now prominently marked as experimental and DANGEROUS :)

Ready for merge @marktsuchida ? This shouldn't have any effects on normal operation of acquisitions unless the checkbox is clicked. 

In part I would like to merge this sooner rather than later because having multiple branches where acqEngJ  is potentially changing makes me much more prone to mistakes (like https://github.com/micro-manager/pycro-manager/issues/449). I definitely can (and will) improve workflows and tests in the future to make this less likely, but in the meantime this would be very helpful

henrypinkard:
Wait scratch that https://github.com/micro-manager/pycro-manager/issues/449 is unrelated

marktsuchida:
Yes, I'm all for branches with short lifetime, and one way to achieve that is to have an in-development feature in the main branch but disabled by default. With the clear marking now added, I think there is no reason not to merge.

Ready to merge once linting errors are fixed.

nicost:
Happy to help with the linting @Henry, but could you try https://github.com/micro-manager/micro-manager/pull/1548 first?  B.t.w., anything against merging #1548?

henrypinkard:
Yes, working on it...

nicost:
I linted this code and pushed it to nicost/acq_eng_studio_ns.  How do I push this into Henry's branch?

marktsuchida:
@nicost Since the branch for this PR is on @henrypinkard's fork, you need to
- add a "remote" for @henrypinkard's fork (`git remote add henrypinkard https://github.com/henrypinkard/micro-manager.git`),
- fetch it (`git fetch henrypinkard`),
- then checkout his branch (`git checkout acq_eng_studio`).
Then you can commit to that branch and push.

(The last command would need to be `git checkout acq_eng_studio henrypinkard/acq_eng_studio` if `acq_eng_studio` were not unique.)

nicost:
So weird that I can push to Henry's branch.  I did and it seems to have gone in and the Lint task now passes.  Please make sure it compiles before merging, I did not manage to get all dependencies correct.

nicost:
Nice!  The CI showed a problem compiling, and I just fixed it;)

marktsuchida:
> > So weird that I can push to Henry's branch.
> 
> Yeah strange...

When you create a pull request from a fork, there is a checkbox "Allow edits and access to secrets by maintainers". It is checked by default.

PR #1548, opened 2022-11-07
Add Java code style for IntelliJ IDEA

marktsuchida:
I think this may provide a practical short-term fix for #1509 (pending feedback), for IntelliJ IDEA only.

The file added here (`buildscripts/codestyle/idea-java-code-style.xml`) is exported from IDEA's `Settings > Editor > Code Style > Java`, after having imported code style settings from `checkstyle.xml`. Adding this because import of `checkstyle.xml` into IDEA seems to fail on macOS (#1501) (this export was created on Windows).

It can be imported by going to `Settings > Editor > Code Style > Java`, clicking the Gear next to the Scheme, and selecting `Import Scheme > IntelliJ IDEA code style XML`. (`Settings` is `Preferences` on macOS.)

You probably want to set the Scheme to "Project" (as opposed to "Default") before importing and choose to import into the current scheme when asked. That way the settings will only apply to the current project.

If this works, we can update `doc/pre-commit.md` to refer to this.

The only catch is that it may need regenerating if we change the Checkstyle settings, but I suspect it will be rare that we change the types of settings contained in this file. (And in any case it's not hard to regenerate.)

Cc: @henrypinkard @nicost.

### DIFF
--- a/buildscripts/codestyle/idea-java-code-style.xml
+++ b/buildscripts/codestyle/idea-java-code-style.xml
@@ -1,0 +1,46 @@
+<code_scheme name="Project" version="173">
+  <JavaCodeStyleSettings>
+    <option name="CLASS_COUNT_TO_USE_IMPORT_ON_DEMAND" value="999" />
+    <option name="NAMES_COUNT_TO_USE_IMPORT_ON_DEMAND" value="999" />
+    <option name="PACKAGES_TO_USE_IMPORT_ON_DEMAND">
+      <value />
+    </option>
+    <option name="IMPORT_LAYOUT_TABLE">
+      <value>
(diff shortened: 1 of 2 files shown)


marktsuchida:
Note that this only covers the code style settings, where the import failed. You still need to install the Checkstyle-IDEA plugin and configure it, if you want to see the style errors in the IDE.

nicost:
Seems like a good idea to me.  @henrypinkard can you try?  B.t.w, will be nice to add this to https://micro-manager.org/Using_IntelliJ.

henrypinkard:
@nicost Tried it, still cannot get it to work

marktsuchida:
@henrypinkard What were the symptoms?

PR #1551, opened 2022-11-09
Adds out and pc to gitignore.  pc is a local script to start the pre-commit hook.

nicost:


marktsuchida:
The `/out/` makes sense.

Regarding the `pc` script -- I'm not going to say never, but is there any chance we can either provide a script that works for everybody, or keep personal tools out of the common .gitignore?

For example, I put the following in my `.bashrc`. It allows me to type `venv` anywhere to activate the virtual environment (creating if missing), and makes `venv/` self-ignoring by putting a `.gitignore` inside it.
```bash
alias venv-activate='source ./venv/Scripts/activate'
# Non-windows: alias venv-activate='source ./venv/bin/activate'

venv() {
	if [ ! -d ./venv ]; then
		echo 'Creating venv...'
		python -m venv venv
		echo '*' >venv/.gitignore
	fi
	venv-activate
}
```

Also it seems that pre-commit now does work in Git Bash when installed globally (this wasn't the case when I wrote the instructions to use a venv). Can you see if this works for you, too? If not, I'll try to come up with a better way.

marktsuchida:
> I do not understand `"it seems that pre-commit now does work in Git Bash when installed globally (this wasn't the case when I wrote the instructions to use a venv). Can you see if this works for you, too?" `. What should I do differently than the instructions?

There are (roughly) two ways to install a Python package: globally or into a virtual environment.

`pip` will install packages to the virtual environment if one is currently active; otherwise it will install to a global location.

(The "global" location is usually per Python installation, and on Windows is often per user, but this does not matter for the present issue.)

It is usually recommended that Python libraries be installed to a venv so that versions can be chosen per project without clashing (and so that you can determine exactly what the dependencies are). But we are not doing Python development (in this repo). For a command line tool like `pre-commit` that just _happens_ to be written in Python, it usually makes more sense to install it globally, so that `pre-commit` is a valid command in bash without having to activate a virtual environment.

The only reason I wrote instructions for using a venv is that, at the time, in Git Bash (only) it did not work correctly otherwise. This seems no longer the case for me and, looking at upstream issues, it appears that there were particular versions of Git for Windows and/or Python that could cause issues.

So I'm hoping that we can remove all mentions of venv from our [instructions](https://github.com/micro-manager/micro-manager/blob/main/doc/pre-commit.md#installing-pre-commit), replacing it with the much simpler:
```bash
python -m pip install pre-commit
cd path/to/micro-manager
pre-commit install
```
Once this is done successfully, there is no need to activate anything when you come back to the same git clone, even if in a new terminal.

We might add a note that Git and Python should be updated if it doesn't work, but I think this is much less cumbersome, especially for people who are not already familiar with venv.

nicost:
Just test, and the global install works perfectly!  Can you add that to the instructions?  That should make things very easy.

Merging to add the out folder (would be sad if that got accidentally pushed...).

marktsuchida:
@nicost Thanks, will do. The hook will only run in repos where `pre-commit install` has been run.

PR #1555, opened 2022-11-12
AcqEngine: Fix bug with camera channel names, causing acquisitions with camera 2 to fail after running acquisition with multi-camera.

nicost:
This is not an elegant fix, but all I can do with no knowledge of Clojure.  The acquisition engine knows the number of camera channels and should take those into account.  This fix ensures that the code tagging the metadata will use the correct camera-channel-name.


nicost:
Tested with the Democamera setup.  Works as expected.

marktsuchida (approved):

VOndrejPM:
I have tested the latest build and I can confirm that it is now working with MicroManager Demo cameras as well as with Prime BSI cameras. Thank you for the quick fix.

Issue #1557, opened 2022-11-16
Deadlock in viewer


nicost:
```text
2022-11-15T18:34:54.003795 tid3564 [IFO,App] EDTHangLogger: Missed heartbeat; waiting to see if we are stuck on a single event
2022-11-15T18:34:54.348870 tid5016 [IFO,Core] Default shutter set to LaserShutter
2022-11-15T18:34:55.006109 tid3564 [IFO,App] EDTHangLogger: Event handling has exceeded at least 1001 ms (currently 5502 ms since heartbeat timebase)
                                   [       ] Stack traces follow (note: thread states queried later than stack traces)
                                   [       ] Thread 2 [Reference Handler] WAITING
                                   [       ]   at java.lang.Object.wait(Native Method)
                                   [       ]   at java.lang.Object.wait(Object.java:502)
                                   [       ]   at java.lang.ref.Reference.tryHandlePending(Reference.java:191)
                                   [       ]   at java.lang.ref.Reference$ReferenceHandler.run(Reference.java:153)
                                   [       ] Thread 3 [Finalizer] WAITING
                                   [       ]   at java.lang.Object.wait(Native Method)
                                   [       ]   at java.lang.ref.ReferenceQueue.remove(ReferenceQueue.java:143)
                                   [       ]   at java.lang.ref.ReferenceQueue.remove(ReferenceQueue.java:164)
                                   [       ]   at java.lang.ref.Finalizer$FinalizerThread.run(Finalizer.java:209)
                                   [       ] Thread 5 [Signal Dispatcher] RUNNABLE
                                   [       ] Thread 6 [Attach Listener] RUNNABLE
                                   [       ] Thread 21 [RMI TCP Accept-0] RUNNABLE
                                   [       ]   at java.net.DualStackPlainSocketImpl.accept0(Native Method)
                                   [       ]   at java.net.DualStackPlainSocketImpl.socketAccept(DualStackPlainSocketImpl.java:131)
                                   [       ]   at java.net.AbstractPlainSocketImpl.accept(AbstractPlainSocketImpl.java:409)
                                   [       ]   at java.net.PlainSocketImpl.accept(PlainSocketImpl.java:199)
                                   [       ]   at java.net.ServerSocket.implAccept(ServerSocket.java:545)
                                   [       ]   at java.net.ServerSocket.accept(ServerSocket.java:513)
[...]

PR #1562, opened 2022-11-17
Studio: Prevent deadlock related to animation fps values.

nicost:
This change looks like it will prevent the deadlock in #1557.  I do not quite like pulling in a google class (AtomicDouble), but it seems the cleanest.  Otherwise, we should lock just the get and set fps (i.e. not synchronize them with everything else in this class.

### DIFF
--- a/mmstudio/src/main/java/org/micromanager/display/internal/animate/AnimationController.java
+++ b/mmstudio/src/main/java/org/micromanager/display/internal/animate/AnimationController.java
@@ -13,6 +13,7 @@
 
 package org.micromanager.display.internal.animate;
 
+import com.google.common.util.concurrent.AtomicDouble;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.concurrent.CancellationException;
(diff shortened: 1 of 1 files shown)


marktsuchida (approved):
I think this is fine if it fixes the issue. It should be safe because `animationRateFPS_` is only accessed through `getAnimationRateFPS()`, and anyone calling the latter already knows that the value may change at any moment.

The atomic double is pretty simple to implement ourselves if we ever need to (just use an atomic long and convert to/from double when loading and storing). But the ready-made `AtomicDouble` is cleaner, given it doesn't add a new dependency.

marktsuchida:
Should this be linked to #1557?

nicost:
Yes!  Closes #1557.  Will that do it?

marktsuchida:
@nicost I think the "Closes" keyword only works if it is in the first post of the PR (this was unclear to me at first, but it probably makes sense). You can also use the Development section of the right panel to link to issues.

PR #1568, opened 2022-11-20
bump versions

henrypinkard:


marktsuchida:
Build still failing nearly 2 hours later.

henrypinkard:
Missed a push. 

Is it possible to setup dependabot to automate these with the current build system or does that need to wait for mavenization?

marktsuchida:
We need to be using Maven. Also, the merging of pull requests (created by Dependabot) would be manual even then (I think you can make it automatic, but that is not always what we'd want for all dependencies and would probably cause more problems than it solves).

Simply not merging until the automated checks pass seems like the best solution, keeps the noise down, and works right now. We will have better checks on PRs in the future, but they can't help if you ignore them.

henrypinkard:
This was a problem of me not pushing (and building) a dependency though. So a dependabot wold not ever have created the PR in the first place. I may set up a dependabot for pycromanager et al. in the future. If its possible in the future to make a dependabot for MM in the future that only looks for changes in these dependencies and merges when checks are complete and passed, this would seem to be the ideal candidate for automation, because it would be doing exactly what I currently do manually.

marktsuchida:
Future automation will be nice, I agree. I don't see how that has anything to do with ignoring automated checks today. It will be important to pay attention to the checks even if all or part of this is automated. It's not that hard -- just wait a few minutes for the green check. If it fails, and it's due to dependencies not being available yet, you can wait a little and re-run the failed checks before merging. This should take less than a minute of (active) time. Seems like an acceptable cost for smooth collaborative development. I really don't understand the thinking leading to the conclusion that it is okay to merge code that is known not to build. It's always less work to keep things isolated -- in this case, it's wonderful that we can check PRs (if only partially for now) before merging so that we are reasonably confident that the build will not break when merged. Why throw that away?

henrypinkard:
I think it will be easier to understand in person, lets discuss at our next meeting

PR #1575, opened 2022-11-24
Studio: Prevent null pointer exception in SCOPE_DATA_KEYS metadata handling. 

nicost:
This code is very difficult to understand (still, after all these years), but I do believe it relied on a side effect.  When the Core no longer inserts the system state cache in tagged image metadata, this code throws a null pointer exception. Keep it in place just in case, but prevent null pointer exceptions.


marktsuchida (approved):
The code is trying to remove device property entries from the flat metadata if they already appear in the "scope data".

I think the previous code made the assumption that every key that appears in the "scope data" (assumed to be output of `getSystemStateCache()`) is also contained in the flat metadata. Turning off the inclusion of device properties in `TaggedImage` means that this (already inappropriate) assumption no longer holds. This fix is correct.

PR #1576, opened 2022-11-27
Studio: Snap/Live now ask GetTaggedImage not to include the System State Cache in the metadata.

nicost:
This avoids the superfluous "user" metadata replicating the device state already available under the "device" metadata.

### DIFF
--- a/mmstudio/src/main/java/org/micromanager/acquisition/internal/DefaultAcquisitionManager.java
+++ b/mmstudio/src/main/java/org/micromanager/acquisition/internal/DefaultAcquisitionManager.java
@@ -256,6 +256,8 @@ public final class DefaultAcquisitionManager implements AcquisitionManager {
          throw new RuntimeException("No camera configured.");
       }
       core.snapImage();
+      boolean includeSystemStateCache = core.getIncludeSystemStateCache();
+      core.setIncludeSystemStateCache(false);
       ArrayList<Image> result = new ArrayList<>();
       for (int c = 0; c < core.getNumberOfCameraChannels(); ++c) {
          TaggedImage tagged = core.getTaggedImage(c);
@@ -265,6 +267,7 @@ public final class DefaultAcquisitionManager implements AcquisitionManager {
          temp = temp.copyWith(newCoords, newMetadata);
          result.add(temp);
       }
+      core.setIncludeSystemStateCache(includeSystemStateCache);
       return result;
    }
 
--- a/mmstudio/src/main/java/org/micromanager/acquisition/internal/DefaultAcquisitionManager.java
+++ b/mmstudio/src/main/java/org/micromanager/acquisition/internal/DefaultAcquisitionManager.java
@@ -259,15 +259,20 @@ public final class DefaultAcquisitionManager implements AcquisitionManager {
       boolean includeSystemStateCache = core.getIncludeSystemStateCache();
       core.setIncludeSystemStateCache(false);
       ArrayList<Image> result = new ArrayList<>();
-      for (int c = 0; c < core.getNumberOfCameraChannels(); ++c) {
-         TaggedImage tagged = core.getTaggedImage(c);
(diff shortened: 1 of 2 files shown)


mmstudio/src/main/java/org/micromanager/acquisition/internal/DefaultAcquisitionManager.java
```diff
@@ -265,6 +267,7 @@ public List<Image> snap() throws Exception {
          temp = temp.copyWith(newCoords, newMetadata);
          result.add(temp);
       }
+      core.setIncludeSystemStateCache(includeSystemStateCache);
```
marktsuchida: Probably safer to put this in the `finally` clause of `try` ... `finally` so that it won't get skipped when there is an exception thrown.

mmstudio/src/main/java/org/micromanager/internal/SnapLiveManager.java
```diff
@@ -291,6 +294,8 @@ private void startLiveMode() {
       }
 
       synchronized (this) {
+         includeSystemStateCache_ = core_.getIncludeSystemStateCache();
```
marktsuchida, line 297: The call to `stopLiveMode()` above would have overwritten the Core's state unconditionally with the value of `includeSystemStateCache_`, so this line is not successfully saving the state before the call to this function.

I think there needs to be a mechanism to only "restore" the Core state when `stopLiveMode()` is actually stopping live mode, as opposed to being called redundantly (which is allowed).

marktsuchida (commented):
Hope my comments make sense, let me know if not. Basically I'm worried about the cases where the original value may not be restored, or where the "original" value gets spuriously "restored" when it shouldn't be.

Also it would be nice to record in the PR or commit message what (if any) user-visible effects this change has (in the resulting metadata).

Finally, it would be good to test this under conditions where Live is automatically paused and resumed (for example when setting a property in the Device Property Browser), if you haven't done so already.

mmstudio/src/main/java/org/micromanager/acquisition/internal/DefaultAcquisitionManager.java
nicost: Good point.  Added a try - catch with a finally clause.

mmstudio/src/main/java/org/micromanager/internal/SnapLiveManager.java
nicost: Great point! Moved the "restore" call to a place that makes much more sense.

nicost:
> Also it would be nice to record in the PR or commit message what (if any) user-visible effects this change has (in the resulting metadata).

Now added in PR.

> Finally, it would be good to test this under conditions where Live is automatically paused and resumed (for example when setting a property in the Device Property Browser), if you haven't done so already.

Done. No problems seen.

mmstudio/src/main/java/org/micromanager/acquisition/internal/DefaultAcquisitionManager.java
```diff
@@ -259,15 +259,20 @@ public List<Image> snap() throws Exception {
...
+         for (int c = 0; c < core.getNumberOfCameraChannels(); ++c) {
+            TaggedImage tagged = core.getTaggedImage(c);
+            Image temp = new DefaultImage(tagged);
+            Coords newCoords = temp.getCoords().copyBuilder().channel(c).build();
+            Metadata newMetadata = temp.getMetadata().copyBuilderWithNewUUID().build();
+            temp = temp.copyWith(newCoords, newMetadata);
+            result.add(temp);
+         }
+      } catch (Exception ex) {
+         studio_.logs().logError(ex);
```
marktsuchida: I don't think we want to swallow the exception here. The caller of `snap()` should get to handle the exception, as has been the case.
```suggestion
```

PR #1577, opened 2022-11-28
Studio: add way to tune extra wait for StageControl Snap after move.

nicost:
The code had a 100ms wait time to accomodate ASI stages going setting BUsy to false, but then moving again.  A similar problem occurs with Nikon Stages, but they need a longer wait.  We are paving over problems in the device adapters, but at least this gives an easy way for the user to make these nice cabilities usable, wheras it may be very difficult to fix this in the device adapters.  It is interesing, however, that these types of problems do not seem to show up in MDAs, so may be something else is going on.


nicost:
> (but only add methods to an internal class; did you mean to add this adjustment to the GUI as well?)

That was the goal for now.  This makes it easy to set in a startup script.  That may be all that is needed, but also gives me the space to investigate and hopefully get to a better solution. 

> Probably good to update the comment to also mention Nikon

Done.

PR #1582, opened 2022-12-05
Extend MMConfig_demo.cfg

ieivanov:
This PR extends the MM demo config to add an LED state device with sequenceable states and associated channel configs built around a multiband dichroic and emission filter.

Related to https://github.com/micro-manager/mmCoreAndDevices/pull/307 and https://github.com/micro-manager/pycro-manager/pull/458


marktsuchida:
This looks fine to me, but I'll wait to merge until micro-manager/mmCoreAndDevices#307 is resolved (probably sequence=on will need to be added to System-Startup).

ieivanov:
Perfect. 

I think it will be OK if the DStateDevice defaults to Sequence=Off and we don't turn it on in the Startup settings of this config. If folks would like to use that functionality they can modify the config after installation or change the settings with a call to the Core.

Here I'm renaming the previous DShutter from `Shutter` to `White Light Shutter`, do you think that's OK?

ieivanov:
@marktsuchida I think this is ready to merge as well, unless you want to rename the `White Light Shutter` back to `Shutter` for backward compatibility, which will be very easy.

marktsuchida:
Thanks. Renaming the device labels in the demo config shouldn't pose any compatibility issues (unlike renaming device _names_ in the device adapter).
I agree that it makes sense to leave sequencing off to keep the demo simple.

PR #1589, opened 2022-12-13
Image exporter now can include labels in the file names.

### LINKED ISSUE #1588: Save images as displayed: save using useful labels
When saving as pngs, saved images are currently numbered.  It would be useful to provide more useful names, for instance, when saving multiple positions, add the position name, when saving multiple channels, add the channel name, etc...

nicost:
ExportMovie now can write labels in the file names. These should make it much easier to work with exported images. Cleaned up the code a bit while I was at it.

Closes #1588


nicost:
Also fixes a possible deadlock in the display code (ensure that DisplayController.show() is actually run on the EDT).

PR #1602, opened 2023-01-11
bump versions

ieivanov:
@henrypinkard and this too, right?


henrypinkard:
Yep, correct. I think the test should pass once the new version of PycroManagerJava is available through maven

PR #1605, opened 2023-01-14
bump AcqEngJ and PycroManagerJava versions

ieivanov:


henrypinkard:
Closing because these version numbers are outdated since I had to bump the versions many times while testing the build system automation

Issue #1583, opened 2022-12-06
Java code formatting and Checkstyle


marktsuchida:
I had a Zoom chat with @henrypinkard about code formatting, style checking, pre-commit hooks, and workflow for merging PRs.

Here's a tentative proposal we came up with (with some details added by me). @nicost what do you think of this general direction? Worth trying? @henrypinkard Please feel free to elaborate, of course.

- Introduce automation so that _specific_ pull requests can be auto-merged once checks pass
    - This will be applied only to PRs that are manually marked by a core developer with a dedicated tag `automerge`
    - The intent is to only use this for trivial PRs (e.g. bumping the version of a dependency) where a core developer is sure that it should be merged as long as the tests pass but does not want to manually wait and merge
    - This should be possible with [this action](https://github.com/marketplace/actions/automatically-merge-pr) (there may be other ways)
    - This can be implemented independent of all of what follows

- Remove the pure formatting (indentation, brace position, etc.) rules from our Checkstyle configuration

- Instead, add a pre-commit hook to run clang-format (which supports Java)
    - This has the advantage that the pre-commit hook will automatically reformat files (after which you can re-stage those files and successfully commit)
    - We'll need to create a clang-format config file that is close to our current style (@marktsuchida will come up with this if we're okay with this general direction)
    - We will likely need to reformat all our Java code once (100% automatic)
    - There is an IDEA plugin for clang-format which we should test and validate (if it doesn't work, we can still provide an IDEA settings file like we do now)
    - This does not add any setup requirements for the pre-commit hook
    - We can defer our final decision on whether to make this switch until after we've actually seen how the formatting behaves

- For the other Checkstyle rules, turn them into warnings instead of errors
    - Warnings can be printed by the pre-commit hook but will not be (easily) visible on GitHub
    - Non-formatting rules include things like name capitalization, method order (overloads togeter), switch statement fall-throughs, etc.
    - These rules typically don't have automated/mechanical fixes (at least not safe ones)
    - For major changes and API changes, some of these style issues can be addressed during code review
    - (Added by @marktsuchida) We never really evaluated each of these rules and whether they help our project. It might be more productive to concentrate on the smallest subset that we truly care about (for example, do we really need to enforce comments on Unicode escapes?)

- Let's make it a rule that we don't (intentionally) merge PRs that do not pass checks (says @marktsuchida)
    - There will be no reason to merge before checks finish, once we can enable automerge for the PR
[...]

PR #1612, opened 2023-01-17
Update pycro-Manager and dependencies

henrypinkard:
Created by Github action

### DIFF
--- a/buildscripts/ivy.xml
+++ b/buildscripts/ivy.xml
@@ -43,7 +43,7 @@
       <!-- Magellan dependencies -->
       <!-- https://mvnrepository.com/artifact/org.zeromq/jeromq -->
       <dependency org="org.zeromq" name="jeromq" rev="0.5.1"/>
-      <dependency org="org.micro-manager.acqengj" name="AcqEngJ" rev="0.23.2">
+      <dependency org="org.micro-manager.acqengj" name="AcqEngJ" rev="0.23.3">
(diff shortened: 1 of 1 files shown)


marktsuchida:
Just trying to understand what is going on here -- is the merge from 'main' into 'dependency-update' done manually here?

It would be nice to keep the history just a little less confusing by avoiding merging 'main' into temporary branches, instead preferring rebasing and force-pushing (unless something I'm not aware made it necessary in this case). Perhaps it is best to use "Squash and Merge" for these auto-generated PRs; otherwise our commit history is swamped by noise.
image

henrypinkard:
I think that was a one-time (or a couple time) thing I did while setting this all up after and messing around with github actions. Sorry, i did not realize that would muck up the history. 

Though I am somewhat confused by when to squash, when to rebase and when to merge. Here is the action that is generating these:  

https://github.com/micro-manager/pycro-manager/blob/2d449c474d8d73766cc5687e78165d751e21f2e6/.github/workflows/maven_deploy_and_mm_update.yml#L74

Do you see a way to make it work better with the histories?

marktsuchida:
I think I figured out what was happening and commented at #1583 (but please correct me if I'm wrong).
I think the root of the problem is reusing the same branch for multiple pull requests, which is usually not a good idea (for both human- and bot-created PRs).
No need to apologize -- I just want to improve the process going forward. Let's continue this discussion (if necessary) in #1583.

PR #1613, opened 2023-01-18
AcqEngJ: Most features now working

nicost:
Still to do: "Keep shutter open", and autofocus.  Probably good to merge now, and continue work in a new branch.


mmstudio/src/main/java/org/micromanager/acquisition/internal/acqengjcompat/AcqEngJAdapter.java
```diff
@@ -334,6 +378,24 @@ public static void addMMImageMetadata(JSONObject imageMD) {
...
+         } else if (imageMD.has("YPosition_um_Intended")) {
+            imageMD.put(PropertyKey.Y_POSITION_UM.key(), imageMD.getDouble(
+                  "YPosition_um_Intended"));
+         }
+         if (imageMD.has("ZPosition_um")) {
+            imageMD.put(PropertyKey.Z_POSITION_UM.key(), AcqEngMetadata.getZPositionUm(imageMD));
+         } else if (imageMD.has("ZPosition_um_Intended")) {
+            imageMD.put(PropertyKey.Z_POSITION_UM.key(), imageMD.getDouble(
+                  "ZPosition_um_Intended"));
+         }
```
henrypinkard, line 398: It could be good to use variables from acqEngJ rather than writing out strings. i.e. `AcqEngMetadata.X_UM`, `AcqEngMetadata.X_UM_INTENDED`, etc

henrypinkard: Or actually, you were adding "has" methods for everything in `AcqEngMetadata` so maybe that would be more consistent. Not a big deal either way though so long as it works

mmstudio/src/main/java/org/micromanager/acquisition/internal/acqengjcompat/AcqEngJAdapter.java
```diff
@@ -482,14 +555,94 @@ private Iterator<AcquisitionEvent> createAcqEventIterator(SequenceSettings acqui
...
+               if (thisPosition != lastPositionIndex_) {
+                  relativePositionStartTime_ =  System.currentTimeMillis() - startTime_;
+                  System.out.println("Position " + thisPosition + " started "
+                        + relativePositionStartTime_ + "  ms after acquisition start");
+                  lastPositionIndex_ = (int) event.getAxisPosition("position");
+                  positionMoved_ = true;
+               }
+               if (positionMoved_) {
+                  long relativeStartTime = relativePositionStartTime_
+                        + event.getMinimumStartTimeAbsolute() - startTime_;
```
henrypinkard, line 626: Is `event.getMinimumStartTimeAbsolute()` actually giving the absolute start time? Or is it relative?

mmstudio/src/main/java/org/micromanager/acquisition/internal/acqengjcompat/AcqEngJAdapter.java
```diff
@@ -482,14 +555,94 @@ private Iterator<AcquisitionEvent> createAcqEventIterator(SequenceSettings acqui
...
+                  int frame = (int) event.getAxisPosition("time");
+                  System.out.println("Pos " + thisPosition + ", Frame " + frame
+                        + " start at " + relativeStartTime);
+                  event.setMinimumStartTime(relativeStartTime);
```
henrypinkard, line 628: And I'm guessing this setting the relative start time? It would be good to change the name of this function. Right now its only used here and by lightsheetmanager (but doesnt work there yet as far as I know). So you can go ahead and change it, or let me know what it should be changed to and I can

mmstudio/src/main/java/org/micromanager/acquisition/internal/acqengjcompat/AcqEngJAdapter.java
```diff
@@ -840,9 +993,15 @@ public void setParentGUI(Studio parent) {
...
+    * This is ignored by the Clojure engine, and also does not have any function
+    * in this engine.  Deprecate?  Do not rely on this to do anything.
+    *
+    * @param stageLabel Name of the focus drive to use.  Ignored.
+    */
    @Override
    public void setZStageDevice(String stageLabel) {
-      zstage_ = stageLabel;
+      zStage_ = stageLabel;
    }
```
henrypinkard, line 1003: I would say deprecate. AcqEngJ right now just gets the Core Z device every time it controls hardware:

https://github.com/henrypinkard/AcqEngJ/blob/b676315feaae22611594f4de9f497e7030440f75/src/main/java/org/micromanager/acqj/internal/Engine.java#L482

mmstudio/src/main/java/org/micromanager/acquisition/internal/acqengjcompat/AcqEngJAdapter.java
```diff
@@ -1160,10 +1319,23 @@ public String getComment() {
...
+            } catch (Exception e) {
+               studio_.logs().logError(e);
+            }
+         }
+         curStore_ = null;
+         curPipeline_ = null;
+         currentAcquisition_ = null;
+         studio_.events().unregisterForEvents(this);
+      }
    }
```
henrypinkard, line 1337: I may be missing something, I dont see anywhere where `AcquisitionEndedEvent` is posted right now, except in a try-catch within the `runAcquisition` function. The correct ending behavior should probably be implemented:

Once `acquistion.areEventsFinished()` returns true, no more events will be executed (i.e. no more hardware will be controlled) and it is safe to move hardware back to initial positions. `acquisition.checkForExceptions()` Can also be called after completion, or at any time during acquisition. It will throw an exception if an exception was thrown by any of the other threads that the acquisition engine manages (whether or not this function was called, an exception thrown by an internal acquisition engine thread will cause the acquisition to abort) 

I don't think that the `AcquisitonEndedEvent` currently posted at the end of `runAcquisition(SequenceSettings sequenceSettings)` is currently correct, because, exceptions haven't been checked for to trigger that catch block, nor has the acquisition been given time to shut down

mmstudio/src/main/java/org/micromanager/acquisition/internal/acqengjcompat/AcqEngJAdapter.java
henrypinkard, line 624: Is `event.getMinimumStartTimeAbsolute()` actually giving the absolute start time? Or is it relative?

mmstudio/src/main/java/org/micromanager/acquisition/internal/acqengjcompat/AcqEngJAdapter.java
henrypinkard, line 396: It could be good to use variables from acqEngJ rather than writing out strings. i.e. `AcqEngMetadata.X_UM`, `AcqEngMetadata.X_UM_INTENDED`, etc

Or actually, you were adding "has" methods for everything in `AcqEngMetadata` so maybe that would be more consistent. Not a big deal either way though so long as it works

henrypinkard:
Awesome that you've gotten so much working!

henrypinkard:
Fine by me if you want to merge now and continue on a new branch

mmstudio/src/main/java/org/micromanager/acquisition/internal/acqengjcompat/AcqEngJAdapter.java
nicost: Yes, AcqEngJ has ` setMinimumStartTime(long l) `, which is relative to the start of the acquisition, and `getMinimumStartTimeAbsolute()`, which is absolute.  At one point it made sense to me that you can not set the absolute start time or get the relative start time, but I don't remember the reasoning.  Agreed the name of this function could be changed to `setMinimumStartTimeRelative(long l)` to make code more readable.  Let's do that after merging this PR, as it needs simultaneous update of AcqEngJ and micro-manager.

mmstudio/src/main/java/org/micromanager/acquisition/internal/acqengjcompat/AcqEngJAdapter.java
nicost: Agreed.  Let's coordinate with Mark, as this will need change to the Micro-Manager version (I believe).  We may take that opportunity to evaluate more of this api.

mmstudio/src/main/java/org/micromanager/acquisition/internal/acqengjcompat/AcqEngJAdapter.java
henrypinkard: okay sounds good

mmstudio/src/main/java/org/micromanager/acquisition/internal/acqengjcompat/AcqEngJAdapter.java
nicost: The `AcquisitionEndedEvent `is posted in the finally clause of the `AcqEngJDataSink.run()` method.  I added some check before generting the Even at the end of `runAcquisition` (commit 875e24714).

mmstudio/src/main/java/org/micromanager/acquisition/internal/acqengjcompat/AcqEngJAdapter.java
nicost: I'll fix this in a second merge.  Still using an older AcqEngJ here.

nicost: Realized first needed to back merge main branch, so fixed these now.

PR #1634, opened 2023-02-08
move AcqEngJ speed test from lightsheetmanager to studio

henrypinkard:


henrypinkard:
@nicost @marktsuchida any objections to me disabling linting as a required check until we've had a chance to define the smallest subset of changes that we actually care about, as discussed in https://github.com/micro-manager/micro-manager/issues/1583. Manually making these style changes is slowing me down by a factor of 2...

henrypinkard:
Ok, thanks! Another PR is fine

PR #1646, opened 2023-02-28
Institute Micro-Manager Governance Model

henrypinkard:

PR #1651, opened 2023-03-08
Refactor Explore acquisitions out of Magellan

henrypinkard:


henrypinkard:
Looks like I accidentally included governance docs in here too. Maybe best to merge after https://github.com/micro-manager/micro-manager/pull/1646 for clarity

henrypinkard:
@nicost Thanks for cleaning that up.

If you want to test this out, launch an explore acquisition with:

https://github.com/micro-manager/pycro-manager/blob/9321edba04881a43293f2b64f9fe6854201ec148/java/src/main/java/org/micromanager/explore/ExploreAcqUIAndStorage.java#L100

It might be worth thinking about incorporating a way to launch this into the studio or the MDA window

PR #1653, opened 2023-03-09
Studio: Implement autostretch method in display-DisplayWindow

### LINKED ISSUE #1652: API's autostretch() method
Hi,
According to the MM 2.0 documentation, there is a method on the DisplayWindow interface called autostretch() that programmatically does the equivalent of clicking the 'Autostretch Once' button in the Inspect window when Live mode is activated –
 
https://valelab4.ucsf.edu/~MM/doc-2.0.0-beta/mmstudio/org/micromanager/display/DisplayWindow.html#autostretch()

When trying to use this function (via pycro-manager), however, it resulted in an UnsupportedOperationException error. 

The documentation states that the DisplayWindow interface is implemented by DefaultDisplayWindow, so I had a look in the recent source code, and although I couldn’t find that, I did find a class called DisplayController that implements that interface, however, the function reads as this: -

   @Override
   public void autostretch() {
      throw new UnsupportedOperationException();
   }

which seems to explain the error message :)

Is this a case that the documentation being aspirational and ahead of the code base, or was this function temporarily and permanently deprecated for some reason? 
Is there perhaps a reason why implementing this functionally might be tricky / have unexpected traps? The code to do that functionality must exist, and clicking the 'Autostretch Once' button in the UI works great.

Otherwise, perhaps there is there simply another approach to achieve the same effect, such as a way to programmatically check, then uncheck the Autostrech check box?

Any opinions or advice would be very welcome.

Thanks.

nicost:
Closes #1652.

### DIFF
--- a/mmstudio/src/main/java/org/micromanager/display/internal/displaywindow/DisplayController.java
+++ b/mmstudio/src/main/java/org/micromanager/display/internal/displaywindow/DisplayController.java
@@ -1042,16 +1042,23 @@ public final class DisplayController extends DisplayWindowAPIAdapter
 
    @Override
    public void autostretch() {
+      if (!SwingUtilities.isEventDispatchThread()) {
+         SwingUtilities.invokeLater(() -> autostretch());
+         return;
+      }
+
+      if (uiController_ == null) {
+         return;
+      }
+
       boolean autoStretchOn = getDisplaySettings().isAutostretchEnabled();
-      if (!autoStretchOn) {
+      if (!autoStretchOn) { // else: no need to do anything
          // since we want to immediately switch off autostretch after applying it, we can not
          // run this action coalesced, but need to do it directly.
          uiController_.applyDisplaySettings(getDisplaySettings().copyBuilder().autostretch(true)
                  .build());
+         setDisplaySettings(getDisplaySettings().copyBuilder().autostretch(false).build());
       }
-      // Strange, but the description specifies that autostretch will be switched off even when it
-      // was on before.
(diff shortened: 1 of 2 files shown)


mmstudio/src/main/java/org/micromanager/display/internal/displaywindow/DisplayController.java
```diff
@@ -1042,7 +1042,16 @@ public void adjustZoom(double factor) {
...
+      boolean autoStretchOn = getDisplaySettings().isAutostretchEnabled();
+      if (!autoStretchOn) {
+         // since we want to immediately switch off autostretch after applying it, we can not
+         // run this action coalesced, but need to do it directly.
+         uiController_.applyDisplaySettings(getDisplaySettings().copyBuilder().autostretch(true)
+                 .build());
+      }
+      // Strange, but the description specifies that autostretch will be switched off even when it
+      // was on before.
+      setDisplaySettings(getDisplaySettings().copyBuilder().autostretch(false).build());
```
marktsuchida: I agree that it is strange or at least unexpected (it also doesn't match what the GUI `Auto Once` button does).

Wouldn't it be better to change the description and make it so that this method does nothing when (continuous) autostretch is enabled? (It won't break anything because it never worked anyway :)

nicost: Great idea!  I changed the description and made the code behave accordingly.  Will merge unless you see an issue.

PR #1656, opened 2023-03-13
Viewer nd tiff saver

nicost:
Makes it possible to save in NDTIff format using the floppy disk icon in the viewer.  Uncovers some weirdness and-or inconsistencies in the various Datastore implementations. This should be revisited, but currently, I believe this should just work.

### DIFF
--- a/mmstudio/src/main/java/org/micromanager/data/internal/DefaultDataSaver.java
+++ b/mmstudio/src/main/java/org/micromanager/data/internal/DefaultDataSaver.java
@@ -81,7 +81,10 @@ public class DefaultDataSaver extends SwingWorker<Void, Void> {
       }
 
       final SummaryMetadata fSummary = summary;
-      duplicate_.setStorage(saver_);
+      if (!(saver_ instanceof NDTiffAdapter)) {
+         // NDTiffAdapter sets storage in the setSummaryMetadata implementation ????
+         duplicate_.setStorage(saver_);
+      }
       duplicate_.setSummaryMetadata(fSummary);
 
       // Copy images ordered by stage position index.
@@ -151,6 +154,28 @@ public class DefaultDataSaver extends SwingWorker<Void, Void> {
       });
       int counter = 0;
       double multiplier = 100.0 / tmp.size();
+      // Before we can put images into the new storage, we have to be sure that the SummaryMeta-
+      // data are there.  We set it before, but that function is asynchronous internally.
+      // I do not see ways other than polling.  Alternatively, the bus used to post
+      // SummaryMetadata Events, could be made synchronous, but I can not oversee the
+      // consequences.
+      long startTime = System.currentTimeMillis();
+      boolean timeOut = false;
+      while (!timeOut && duplicate_.getSummaryMetadata() == null) {
+         if (System.currentTimeMillis() - startTime > 10000) {
+            timeOut = true;
+         }
+         try {
+            Thread.sleep(100);
+         } catch (InterruptedException e) {
+            timeOut = true;
+            studio.logs().logError(e);
+         }
+      }
+      if (timeOut) {
+         studio.logs().showError("Failed to save data");
+         return null;
+      }
       for (Coords coords : tmp) {
          duplicate_.putImage(store_.getImage(coords));
          counter++;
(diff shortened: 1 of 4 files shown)


mmstudio/src/main/java/org/micromanager/data/internal/DefaultDataSaver.java
```diff
@@ -81,7 +81,10 @@ protected Void doInBackground() throws IOException {
       }
 
       final SummaryMetadata fSummary = summary;
-      duplicate_.setStorage(saver_);
+      if (!(saver_ instanceof NDTiffAdapter)) {
```
henrypinkard, line 84: Yes there's nothing for NDTiff to create until it has summary metadata. It expects it in the constructor and it is the first thing written to disk. I suppose this is somewhat different than how this API is designed, but if it works it works

henrypinkard: I dont really understand why there is a distinction between Datastore and Storage

marktsuchida: Datastore is common to all storages; Storage is different for RAM, single-page TIFF, multipage TIFF, etc. File (or memory) I/O is a separate concern from presenting a database-like container for images along axes. At least I'm guessing that was the idea; in reality DefaultDatastore knows way too much about the individual storages and the Storage interface is too close to Datastore so a lot of logic is duplicated anyway. And all of this makes too many assumptions about the storage format. Good to improve if anybody dares :)

nicost: Agreed, that matches what I got away with working through this interface a bit to make this work.  I may work on this a bit next time I hit this place in the code.  OK to merge?

marktsuchida (commented):
The only thing I would consider is whether the polling/timeout logic could be put into NDTiffAdapter (or perhaps a wrapper around it) so that the common code can be kept simple and format-agnostic. I haven't thought through how much work that may be, though.

mmstudio/src/main/java/org/micromanager/data/internal/DefaultDataSaver.java
henrypinkard: Ah got it, that makes sense

henrypinkard:
Fine by me if you want to put polling into NDTiff adapter, or to merge as is

nicost:
I don't think that the problem is inherent to NDTiff storage.  The real problem is that setting SummaryMetadata is asynchronous (i.e. the event is posted on an asynchronous bus, and left to the Storage implementations to handle).  It just so happens that the other storage implementations handle the event much faster than NDTiff, so that we did not notice it before.  Unless we want to post this on a second, synchronous bus, or possibly handle this directly rather than relying on this eventbus, polling in this place is the only way I see to make this work

nicost:
The following comment in org.micromanager.data.Storage sheds some light on the problem:

```
 * <p>Note that the Storage interface does not expose any "setter" methods
 * (e.g. putImage(), setSummaryMetadata(), etc.). It is expected that any
 * read/write Storage listen for the relevant events published by the
 * Datastore instead.</p>
```

Since the events are posted on an asynchronous eventBus, there is never any guarantee that any of the "set" functions have actually executed.  This is probably fine of "putImage()" and the like, but clearly is a problem for setSummaryMetadata().  One solution would be to have two eventBuses, one synchronous, the other asynchronous, but that is outside of the scope of this PR.  

I am not quite sure why this design decision was chosen, it seems a recipe for subtle bugs.

marktsuchida:
I don't think the event bus was asynchronous (change made in 2020) when that comment was written (2014). I don't remember whether or to what extent I expressed at the time my uneasiness with making it asynchronous (and I'm sure there were valid reasons for doing so on the other hand), but this might be an example where other parts of the design need to be redone for the whole system to work nicely under different assumptions. Not suggesting additional work, just clarifying the background.

Another place where the initial assumptions were later changed was that summary metadata now needs to be provided before any images are added. This makes 100% sense but was not part of the early design (at least not explicitly), so there may remain parts of the code that are not done the way we'd expect given the new requirement.

PR #1657, opened 2023-03-13
Duplicator plugin: select any channel, and saves output

nicost:
Previously, only ranges of axes could be selected.  That proved to be cumbersome when desiring to select a sub-set of channels.  

Also added the ability to save directly to any of the formats supported by MM.

### DIFF
--- a/mmstudio/src/main/java/org/micromanager/display/internal/displaywindow/imagej/ImageJBridge.java
+++ b/mmstudio/src/main/java/org/micromanager/display/internal/displaywindow/imagej/ImageJBridge.java
@@ -37,6 +37,7 @@ import java.awt.geom.Rectangle2D;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
+import javax.swing.SwingUtilities;
 import net.imglib2.display.ColorTable8;
 import org.micromanager.data.Coords;
 import org.micromanager.data.DataProvider;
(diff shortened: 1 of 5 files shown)


mmstudio/src/main/java/org/micromanager/display/internal/displaywindow/imagej/ImageJBridge.java
```diff
@@ -305,6 +306,14 @@ private void mm2ijSetDisplayAxisExtents(
...
    @MustCallOnEDT
    public void mm2ijEnsureDisplayAxisExtents() {
+      if (!SwingUtilities.isEventDispatchThread()) {
+         SwingUtilities.invokeLater(new Runnable() {
+            @Override
+            public void run() {
+               mm2ijEnsureDisplayAxisExtents();
+            }
+         });
+      }
```
marktsuchida: This method (and pretty much everything in `ImageJBridge`) is meant to always be called on the EDT only. The bug is in the caller if this is not the case. Hence the (unenforced, for-human) annotation [`@MustCallOnEDT`](https://github.com/micro-manager/micro-manager/blob/7f331552384dcf92a53cc5f75c22ee77ef338f6c/mmstudio/src/main/java/org/micromanager/internal/utils/MustCallOnEDT.java#L19-L28) (it _must_ be called on the EDT; it doesn't automatically schedule itself on the EDT).

The idea is that Swing GUI is fundamentally single-threaded, and the larger part of our code around it that we can carve out and keep single-threaded, the simpler the overall system.

Happy to suggest alternative fixes if you can provide details on how this gets called on a non-EDT thread. At the moment, the only call to this method I can find is also from a method marked with `@MustCallOnEDT`.

marktsuchida (changes requested):
Just one thing.

mmstudio/src/main/java/org/micromanager/display/internal/displaywindow/imagej/ImageJBridge.java
nicost: Good call.  I can not reproduce the issue after removing those lines.  May have been solved by other changes I made.  Removed for now.

marktsuchida (approved):

PR #1668, opened 2023-03-25
Expose abort in AcquisitionManager

jdeschamps:
Hi!

This PR simply exposes a `stopAcquisition` method in `AcquisitionManager`, allowing to programmatically stop an acquisition without prompting users with a dialog.

This is following a [problem I ran into](https://forum.image.sc/t/stopping-acquisition-programmatically/79056). Alternatively I thought of making the user dialog optional in the `haltAcquisition` method.

Does this break a particularlogic in the current development of the acquisition engine? Obviously this would require the same method in the AcqEngJ wrapper.

I have not yet tested it, running into some issues downloading the 3rdparty stuff before building MM (hopefully in the next days I'll get it running).

### DIFF
--- a/mmstudio/src/main/java/org/micromanager/acquisition/AcquisitionManager.java
+++ b/mmstudio/src/main/java/org/micromanager/acquisition/AcquisitionManager.java
@@ -106,9 +106,9 @@ public interface AcquisitionManager {
    void haltAcquisition();
 
    /**
-    * Stop any ongoing acquisition as soon as possible, without user dialog.
+    * Abort any ongoing acquisition as soon as possible, without user dialog.
     */
-   void stopAcqusition();
+   void abortAcquisition();
 
    /**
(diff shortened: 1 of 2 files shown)


jdeschamps:
I finally managed to compile MMJ_ and simply test whether this worked. As expected, this allows closing an acquisition programmatically without user validation.

Does this make sense for the API point of view?

@henrypinkard : can this be easily implemented for AcqEngJ?

henrypinkard:
I think the functionality makes sense, though it might be better to name it also `halt`, since stop will mean, when the new camera API is implemented, "finish what you're doing and then stop", vs something like halt/abort, which means "stop everything right now".

For adding it to AcqEngJ wrapper, I think this should just be a matter of calling `acq.abort`. AcqEngJ has no GUI components so the same method should work between the old and new API calls here

jdeschamps:
Noted! Changed it to `abortAcquisition()` now.

PR #1669, opened 2023-03-25
Added option to copy to the system clipboard to the export images as displayed gear menu.

nicost:
In addition, pressing Ctrl-c in a display window will now copy the currently displayed image to the system clipboard.  This code is in MMKeydispatcher, which is no super elegant.  Suggestions for a better place are welcome, however, copying the currently displayed image with a key combination is super useful.

### DIFF
--- a/mmstudio/src/main/java/org/micromanager/display/ImageExporter.java
+++ b/mmstudio/src/main/java/org/micromanager/display/ImageExporter.java
@@ -52,7 +52,11 @@ public interface ImageExporter {
        * Output as an ImageJ stack, which will open in a new window; this
        * "format" does not create any files on disk.
        */
-      OUTPUT_IMAGEJ
+      OUTPUT_IMAGEJ,
+      /**
+       * Output to the System Clipboard, will only work for a single image
+       */
+      OUTPUT_CLIPBOARD
    }
 
    /**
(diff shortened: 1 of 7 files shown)


marktsuchida (commented):
Nice function!

1. On macOS, copy should be Command-C, not CTRL-C. I don't know this area of Swing well, but I think it is worth investigating how the Copy command (in the Edit menu) is normally supposed to be implemented in Swing apps.
2. Is there no danger of interfering with selecting and copying text? (See the `checkSource` method that is called for hot keys.)
3. It really doesn't make sense for `MMKeyDispatcher` to know about viewer specifics (as you point out). However, making this right might require refactoring much of our key handling code. Perhaps for now a good solution would be to put the actual export code somewhere appropriate (viewer) and make `MMKeyDispatcher` simply call it.
4. Alternatively (to all of the above?), I'm pretty sure there is a way to attach keyboard actions to a single window. The `DisplayUIController` could do this, and the app-wide `MMKeyDispatcher` wouldn't even need to know about it (and it would automatically only work when the window is in focus). Perhaps something like [this](https://stackoverflow.com/questions/24724883/jframe-keyboard-shortcut-using-actionmap).

nicost:
Nice!  I think this does it.

nicost:
OK to merge?

henrypinkard:
Whoops Sorry just accidentlly merged and umerged (https://github.com/micro-manager/micro-manager/pull/1674). I'll let @marktsuchida sign off on this one

PR #1670, opened 2023-03-28
Micro-Manager Improvement Proposal 1: Roadmap for performance, accessibility, and flexibility

henrypinkard:
@marktsuchida @nicost , It still needs some work, but I thought it would be good to get this out there to get more feedback and refine it


marktsuchida:
I'll take a better look asap (only skimmed so far), but I would say each of the major sections of this merits a separate MMIP (or multiple), if we want this to be a useful medium for agreeing upon designs. Not that it's not a good idea to have an overarching MMIP like this to define general priorities, and explain the interrelationships between them. (Perhaps this is what you intended.)

As a roadmap, it would be helpful to propose an order in which the goals may be tackled such that a continuous evolution is possible. Especially given that (in my estimation) there is potentially several person-years' worth of work proposed here.

henrypinkard:
@marktsuchida as a reminder, our policy on these is:

> Champion writes up an initial proposal, which is quickly merged into the micro-manager repository in the MMIP folder (regardless of whether the proposal will be changed and/or ultimately accepted). Additional PRs may be made by the author to update or expand the MMIP, or by maintainers to set its status, discussion URL, etc. Regardless of whether the MMIP is implemented, it will remain in this folder.

So unless there is something glaring here, lets follow through with that and open issues for discussions of areas for improvement when you have a chance to go through in more detail

marktsuchida:
??? I was only giving feedback because I thought you asked for it? Did I break a rule here?
I'm happy to wait until its merged if that is the preferred way.

nicost:
This is super well written and is a great basis for future work, thanks Henry!  Couple of remarks:
- In this day and age, creating a new data storage format is frowned upon.  There should be a strong rational, i.e. reasons why existing formats are not sufficient (which is currently not explicit in the document), and the plan to include easily usable APIs at least for reading (which appears to be included)
-  It should at least be noted that there already is an acquisition engine written in Python (https://pymmcore-plus.github.io/pymmcore-plus/).  Why not use that one?  Why not figure out together what the best route may be?
- 5 - Ability to add arbitrary data, can link to https://github.com/micro-manager/mmCoreAndDevices/pull/313

Awesome work Henry!

henrypinkard:
Yes! Lets hear them

nicost:
If your input is more related to the User Interface, it may be better to open a new issue, rather than adding to this document (which is more concerned with the software architecture of MM). Input is always welcome, and issues here on Github, or discussions on https://image.sc (tag with "micro-manager") are the preferred way.

tlambert03:
> It should at least be noted that there already is an acquisition engine written in Python (https://pymmcore-plus.github.io/pymmcore-plus/). Why not use that one? Why not figure out together what the best route may be?

Thank you @nicost  :)

> Yes, this can be mentioned. There are also many other [this one, for example](https://github.com/czbiohub/coPylot) that use python, some of which use MM, some of which don't.

I would say that pymmcore-plus is not just any like one of "many other" acquisition engines written in python.  It has been designed specifically for MM, with near feature parity to AcqEngJ (with the current exception of data storage, because I agree strongly with @nicost's comments about new storage formats [here](https://github.com/micro-manager/micro-manager/pull/1670#issuecomment-1489250713)... and we've been agnostic on this so far). 

> Regarding why not to coordinate with pymmcore-plus in particular on this. 

I'm saddened to hear you say that you don't want to coordinate.  I started putting more work into pymmcore-plus largely because you expressed [very little interest](https://github.com/pymmcore-plus/napari-micromanager/issues/97#issuecomment-1041891685) in a pure-python engine at the time and seemed mostly content with the python->java->c++ solution.

> A primary goal creating a python acquisition engine would be to enable swapping seamlessly between Java and Python backends in pycro-manager.

i don't think it would be that hard at all to add a thin adapter layer between backends.  It would be made particularly easy if you swapped your `multi_d_acquisition_events` for a list of `useq.MDAEvents`, which was essentially designed to be compatible with the pycromanager pattern (and is ready to go for easy serialization, etc...).  Most of the pycro-manager API [has equivalents](https://pymmcore-plus.github.io/pymmcore-plus/guides/pycromanager_migration/) in pymmcore-plus.

I think that there is perhaps a bit too much emphasis on the pycro-manager API specifically.  The "core" here has always been MMCore, and pycro-manager added the new/recent APIs.  The 

> The related [schema for multi-dimensional acquisitions](https://github.com/pymmcore-plus/useq-schema) is trying to solve a different challenge: a generic language that works across all acquisition systems, not an MM-specific acquisition that enables full use of MM's current and future APIs.

I don't think that's an entirely fair characterization.  Yes, it's _trying_ to be an implementation agnostic schema (it doesn't include any engine in the library itself), but it is obviously designed first around the needs of MM.  It has pretty much all of the fields you would need to take advantage of MM's API (including [device property tuples](https://pymmcore-plus.github.io/useq-schema/schema/event/#useq.PropertyTuple)), and it's not like we're not open to suggestions if you felt like there was something you needed.

I even added [to-pycromanager](https://github.com/pymmcore-plus/useq-schema/blob/main/src/useq/pycromanager.py) utilities to make this cross-compatibility clearer.  Perhaps it's time for you to have another look?

The acquisition engine in pymmcore-plus is beginning to reach a nice state of maturity, with good feedback from early adopters.  It would be great to feel like it's not so easily dismissed here?

nicost:
Is the biggest question/conundrum how to "match" a pymmcore-plus event (https://pymmcore-plus.github.io/useq-schema/schema/event/#useq.MDAEvent) to a an AcqEngJ acquisition event (https://github.com/micro-manager/AcqEngJ/blob/main/src/main/java/org/micromanager/acqj/main/AcquisitionEvent.java)?  

I have to confess that - even though I have written code making extensive use of the - the AcqEngJ AcquisitionEvents are still a bit of a mystery to me.  I would be very much helped with having a well defined interface (i.e., with each function documented), separated from the implementation, and a clear definition of the keys that are used in the toJSON and fromJSON conversions.

Maybe these two are not very far apart, and uniting them can strengthen both?

tlambert03:
> AcquisitionEvents are still a bit of a mystery to me. I would be very much helped with having a well defined interface

exactly!  That was the goal of useq schema.

> Is the biggest question/conundrum how to "match" a pymmcore-plus event 

I don't think it's that big of a lift... the code above has a basic example.

> Maybe these two are not very far apart, and uniting them can strengthen both?

Indeed, the whole point of useq was to welcome feedback on a spec for an event (so that we don't all have to wonder what the event "can" be, and go back to the source code to figure out how it's supposed to be used).  By all means, if you see any limitations in the MDAevent that prevent you from using it as a spec, please suggest changes!

henrypinkard:
> I think that there is perhaps a bit too much emphasis on the pycro-manager API specifically.  The "core" here has always been MMCore, and pycro-manager added the new/recent APIs.  Suffice it to say that I don't think it should be an immediate foregone conclusion that pycro-manager *is* "the API" to control micro-manager from python.  (pymmcore could make that claim, a long time ago)

A little history here. Pycro-manager and pymmcore were adopted as the official low-level and high level APIs for micro-manager, respectively around the same time in early 2020. Pycro-manager was under development for several years prior under a different name. pymmcore is the successor of MMCorePy, which was formerly included in Micro-Manager and supported only Python 2.x. Neither of these has a high level API. Pycro-manager controls the core also, but this isn’t really an API since its just exposing MMCoreJ. All of this is not to say there can’t or shouldn’t be additional official high-level APIs in the future, just to clarify the current state of affairs.

> I'm saddened to hear you say that you don't want to coordinate.  I started putting more work into pymmcore-plus largely because you expressed [[very little interest](https://github.com/pymmcore-plus/napari-micromanager/issues/97#issuecomment-1041891685)](https://github.com/pymmcore-plus/napari-micromanager/issues/97#issuecomment-1041891685) in a pure-python engine at the time and seemed mostly content with the python->java->c++ solution.  So, it's surprising to say the least to hear that you are now interested in doing so, but without consideration of collaboration

@tlambert03, Let me be clear about this: I've asked you multiple times to discuss coordination/collaboration on Zoom, and you've never taken me up on the offer. I've made suggestions to enhance interoperability with your many cool tools and pycro-manager, adding features like headless mode, core callbacks, and napari viewer to enable this. Some of these were specifically at your request. However, as far as I know there has been little reciprocal effort/interest in pymmcore-plus to move forward on these suggestions. I know we have many differences of opinion, and that is okay. But I think willingness to collaborate is a 2-way street. So I reiterate my offer: if you want to find ways to coordinate, email me, and we can discuss anything and everything on Zoom. I sincerely would like to find areas of common interest where we could work together.

Probably should have done this a while ago, but in keeping with our governance model (see below), I’m going to go ahead and merge. Future PRs may modify parts of this MMIP, or it could be split into multiple MMIPs with additional detail as Mark suggests.

> Champion writes up an initial proposal, which is quickly merged into the micro-manager repository in the MMIP folder (regardless of whether the proposal will be changed and/or ultimately accepted). Additional PRs may be made by the author to update or expand the MMIP, or by maintainers to set its status, discussion URL, etc. Regardless of whether the MMIP is implemented, it will remain in this folder.

tlambert03:
> I've asked you multiple times to discuss coordination/collaboration on Zoom, and you've never taken me up on the offer.

? I know you offered once three weeks ago in https://github.com/micro-manager/pycro-manager/issues/552#issuecomment-1655871995 

suffice it to say I don't see it the same.  I feel like pymmcore-plus and useq-schema have made more attempts to accomodate pycro-manager than vice versa.  But hey, shall we start again?  I know @nicost mentioned a zoom call too.  I think it's safe to say we're probably both misinterpreting the other.  so let's have a chat!

PR #1693, opened 2023-05-02
Pfs plus software

nicost:
Adds an autofocus plugin that sets the offset of the PFS based on software autofocus.  One would think this can be done by setting the PFS offset as the ZDrive device, but that turns out to give bad results, likely because the Z position lags behind the PFS offset drive.  See the code for further comments.  Warrants a page on the website to document.


marktsuchida:
LGTM.

PR #1695, opened 2023-05-04
Magellan: respond to pixel size changes in the hardware.

henrypinkard:


henrypinkard:
@nicost

nicost:
This code does not build, presumably because it needs a newer version of PycroManagerJava than 38.10.   Can you push this out and update ivy.xml?

PR #1704, opened 2023-05-25
AcqEngJ compat layer: enable channels skip frames and use Z stack features

nicost:
I ended up having the iterators return null events to signal that they do not want to execute this specific branch of the event tree.  Accompanying small changes (https://github.com/micro-manager/AcqEngJ/pull/88) in AcqEngJ are needed.  Tested quite a bit (with demo equipment only) and seems to work quite well.


mmstudio/src/main/java/org/micromanager/acquisition/internal/acqengjcompat/AcqEngJAdapter.java
```diff
@@ -179,33 +177,13 @@ private Datastore runAcquisition(SequenceSettings sequenceSettings) {
          sb.customIntervalsMs(null);
       }
 
-      // Several "translations" have to be made to accommodate the Clojure engine:
```
henrypinkard, line 182: I think the OMETiff saving may rely on the metadata for timeFirst/slicesFirst etc in order to embed in the OME metadata. Do these changes affect the tags ending up in summary metadata? If this changes something, you may see that OME Tiffs cannot be opened by bio formats importer

mmstudio/src/main/java/org/micromanager/acquisition/internal/acqengjcompat/AcqEngJAdapter.java
```diff
@@ -448,14 +417,19 @@ private Iterator<AcquisitionEvent> createAcqEventIterator(SequenceSettings acqui
...
       if (acquisitionSettings.useChannels()) {
-         channels = MDAAcqEventModules.channels(acquisitionSettings.channels());
-         //TODO: keep shutter open
-         //TODO: skip frames
-         //TODO: z stack off for channel
+         if (chSpecs.size() > 0) {
+            Integer middleSliceIndex = (acquisitionSettings.slices().size() - 1) / 2;
+            channels = MDAAcqEventModules.channels(chSpecs, middleSliceIndex);
+         }
+         // TODO: keep shutter open, I believe this needs to be done in AcqEngJ; there should be
```
henrypinkard: I think the way to implement the shutter behavior is with an acquisition hook, because we're trying to move towards not hard coding particular microscope assumptions in the acqEng and this shutter behvaior is one of them

mmstudio/src/main/java/org/micromanager/acquisition/internal/acqengjcompat/AcqEngJAdapter.java
```diff
@@ -448,14 +417,19 @@ private Iterator<AcquisitionEvent> createAcqEventIterator(SequenceSettings acqui
...
+         //    a flag in AcquisitionEvent that the Engine should look at, much like how it checks
+         //    if sequencing is possible.
+         // TODO: Check if z stack off for channel image is taken at correct Z
```
henrypinkard: AcqEngJ provides an API for this:
https://github.com/micro-manager/AcqEngJ/blob/3798f07bd0d1a59a81c7992d665d5ee2e9708ae6/src/main/java/org/micromanager/acqj/util/AcqEventModules.java#L140

henrypinkard: Oh nevermind I see that you have it below

mmstudio/src/main/java/org/micromanager/acquisition/internal/acqengjcompat/AcqEngJAdapter.java
nicost: Bit of testing:

AcqEngJ with these lines present:
Order: time slice channel.   Fiji: Incorrect,   BioFormats:  Incorrect
Order slice channel. Fiji: Correct, BioFormats: Correct (but no colors).
Order channel slice.  Fiji: Correct, Bioformats Correct

AcqEngJ without these lines:
Order: time slice channel.   Fiji: Incorrect,   BioFormats:  Incorrect
Order slice channel. Fiji: Correct, BioFormats: Correct (but no colors).
Order channel slice.  Fiji: Correct, Bioformats Correct

I also tested a bit with the old acquisition engine (results were the same).  

I'll put this code back in, but leave a note why.

mmstudio/src/main/java/org/micromanager/acquisition/internal/acqengjcompat/AcqEngJAdapter.java
nicost: Seems to me that shutter behavior is super similar to sequencing.  Also, controlling light exposure to the sample seems like a very common (universal?) part of microscope acquisitions.  

I guess this is doable as a Hook, I can give a try after getting the Autofocus working.

mmstudio/src/main/java/org/micromanager/acquisition/internal/acqengjcompat/AcqEngJAdapter.java
nicost: Oops, are we now adding an offset twice?

mmstudio/src/main/java/org/micromanager/acquisition/internal/acqengjcompat/AcqEngJAdapter.java
henrypinkard: For most cases yes, though its much more of a concern for fluorescence than other modalities. But I'm not sure I would call it universal. I agree that it is universal for using MMStudio, and thus my feeling is its best to have it in that layer, to leave open the possibility of people using AcqEngJ in unexpected ways

nicost:
I added continuous focus and autofocus hooks.  Except for the shutter open stuff, I believe all functionality is there.  Ready for hardware testing.

mmstudio/src/main/java/org/micromanager/acquisition/internal/acqengjcompat/AcqEngJAdapter.java
nicost: Implemented in an acquisition Hook.

PR #1715, opened 2023-06-09
Improvements to AcqEngJ bridge that are tested on a microscope

nicost:
More testing is needed, but this is on the way to actually working.  Code also contains additions to make it easy to write the code to use different channel/slice settings at different positions (a feature we have wanted for more than a decade).

Work in progress, do not merge yet (I thought there was a way to set that in the PR, but I can not find it now).


nicost:
Ah, under the reviewers tab.  Go figure.

nicost:
Tested several times.  Removed danger label.  More testing obviously needed, but I believe this to be ready for wider distribution.

henrypinkard:
Great!

PR #1724, opened 2023-06-19
fix magellan dataset opening

### LINKED ISSUE #1721: Micro-Magellan: problems opening Explore datasets.
After acquiring an Explore dataset with Micro-Magellan, closing it, then opening in Magellan, shows an error message "Error, couldn't create display successfully".  The display appears to be OK, however, with data acquired at the microscope, it was not and failed to open.

henrypinkard:
fixes #1721


nicost:
Thanks!  Will test soon with microscope data.

PR #1725, opened 2023-06-20
Adds option to exclude pixels with value zero from histogram autostretch and stats.

### LINKED ISSUE #1711: Feature request: ignore zeros in LUT autostretch function
I'd like to request a feature be added to the micro-manager image inspector, specifically in the "Histograms and Intensity Scaling" submenu. Basically, I'd like to have an option to "Ignore Zeros" when the image autostretches the LUT. 

I'd like this because I often end up with images that have many pixels with 0 intensity (due to image transformations like deskewing OPM stacks). My argument for why this should at least be a feature is because most scientific cameras have pixel intensity offsets far above 0 to maintain sensor linearity. Therefore, zeros are more likely to arise from a processing artifact than from something physical from the sensor. In that vein, it'd be nice to be able to autostretch the image to only the nonzero values. 

Tagging @nicost here because we've discussed in person.

nicost:
Closes #1711

### DIFF
--- a/mmstudio/src/main/java/org/micromanager/display/DisplaySettings.java
+++ b/mmstudio/src/main/java/org/micromanager/display/DisplaySettings.java
@@ -85,6 +85,8 @@ public interface DisplaySettings {
 
       Builder autoscaleIgnoredPercentile(double percentile);
 
+      Builder autoscaleIgnoringZeros(boolean ignoreZeros);
+
       /**
        * Increases the number of ChannelDisplaySettings in this Builder to
        * the given number.  DefaultSettings for each channel will be added.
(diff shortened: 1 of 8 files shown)


mmstudio/src/main/java/org/micromanager/display/DisplaySettings.java
```diff
@@ -197,6 +199,15 @@ interface Builder {
...
    double getAutoscaleIgnoredPercentile();
 
+
+   /**
+    * When autoscaling, whether or not to take zero values into account.
+    * This can be useful for images that contain artificial zero values.
+    *
+    * @return Whether or not to ignore zeros in the
+    */
+   boolean ignoreZerosWhenAutoScaling();
```
marktsuchida: Since this is an API method, would be good to follow the convention of starting boolean-returning getters with `is`: maybe `isAutoscaleIgnoringZeors()`.
(No need to change the builder method name.)

marktsuchida (approved):
Looks great!
Just one nitpick below.
Apologies for the late review.

mmstudio/src/main/java/org/micromanager/display/DisplaySettings.java
nicost: Good suggestions, done.

PR #1734, opened 2023-07-03
Allow explore acquisition to zoom before data

henrypinkard:
Needs new versions from: https://github.com/micro-manager/pycro-manager/pull/627

### DIFF
--- a/plugins/Magellan/src/main/java/org/micromanager/magellan/internal/magellanacq/MagellanAcqUIAndStorage.java
+++ b/plugins/Magellan/src/main/java/org/micromanager/magellan/internal/magellanacq/MagellanAcqUIAndStorage.java
@@ -333,6 +333,11 @@ public class MagellanAcqUIAndStorage
    }
 
    @Override
+   public void increaseMaxResolutionLevel(int newMaxResolutionLevel) {
+      storage_.increaseMaxResolutionLevel(newMaxResolutionLevel);
+   }
+
+   @Override
    public String getDiskLocation() {
       return storage_.getDiskLocation();
    }


marktsuchida:
Build failed.

```text
 [mm-javac] /Users/jenkins/workdir/workspace/mm2-git-main-macos/micro-manager/plugins/Magellan/src/main/java/org/micromanager/magellan/internal/magellanacq/MagellanAcqUIAndStorage.java:335: error: method does not override or implement a method from a supertype
 [mm-javac]    @Override
 [mm-javac]    ^
 [mm-javac] /Users/jenkins/workdir/workspace/mm2-git-main-macos/micro-manager/plugins/Magellan/src/main/java/org/micromanager/magellan/internal/magellanacq/MagellanAcqUIAndStorage.java:337: error: cannot find symbol
 [mm-javac]       storage_.increaseMaxResolutionLevel(newMaxResolutionLevel);
 [mm-javac]               ^
 [mm-javac]   symbol:   method increaseMaxResolutionLevel(int)
 [mm-javac]   location: variable storage_ of type MultiresNDTiffAPI
```

Could it be that a newer NDTiffStorage is needed? (Build used `org.micro-manager.ndtiffstorage#NDTiffStorage;2.12.1 in central`.)

nicost:
Is there a dependency on pycromanager?  It seems that this code causes an error when starting MicroMagellan.  What dependencies need to be updated?  Can this be done in buildscripts/ivy.xml?

nicost:
I updated NDTiffStorage to the latest (2.13.0), but that does not fix it.  There seems to be an issue with NDViewer:
```java.lang.ClassNotFoundException: org.micromanager.ndviewer.api.ViewerAcquisitionInterface```
but there is no new version of NDViewer to be found.  I'll revert for now.

henrypinkard:
This was not yet meant to be merged, because I think I have some upstream issues to work on on the build automation. I should have marked this as a draft.

But seems like this is the type of thing that should be caught by tests/CI. Maybe something as simple as trying to download the declared dependencies? What do you all think?

nicost:
My bad, I was eager to test it out, and too lazy to pull down the branch to my own machine and build.  

This particular one was caught by the automated build and easily reverted.  Downloading dependencies would not have helped here since there was no update to ivy.xml.  It may be useful to catch problems in that file, but I am not sure if the pay off is worth the trouble.

henrypinkard:
Oh yeah thats a good point. I guess I should have added an update to ivy.

Probably the full build of MM on github is a ways away (if even possible), but seems like it wouldn't be too bad to build the java components as a test? Thoughts @marktsuchida ?

marktsuchida:
Yeah, we could do that. Currently only MMStudio (and its requirements) are built, for the purpose of building the Javadocs. The reason this PR didn't get caught was because it was in a plugin. We could do it now in a hacky way, or we could do it cleanly once MMCoreJ can be downloaded as a binary (first few steps of the build system plans). (I don't think we will ever have a single "full" build on GitHub, but the eventual goal is to have a full installer build(s) that fetches binaries that were also built on GitHub.)

On the other hand, these instances are so rare that I don't mind at all being the human notifier of broken builds. Even if we add the plugins, there will always be other stuff not tested, so the benefits are only incremental and imho not worth it if it results in an even messier [workflow](https://github.com/micro-manager/micro-manager/blob/3acb4555dfe75b0f5b5a4ae586126a63f1d8e759/.github/workflows/apidoc.yml). The problem is that we can't just build the Java components because everything depends on MMStudio and MMStudio depends on MMCoreJ (C++). I'd rather spend energy on getting back to the [plan](https://github.com/micro-manager/micro-manager/issues/1392) for fixing the root problems unless this becomes a frequently recurring problem.

I would want to emphasize that although we aim to keep `main` always building, under the current setup nobody should feel ashamed if a PR they merged (or authored) fails the build. This is expected to happen from time to time, is a consequence of the limitations of the current CI setup, and just needs to be dealt with once it happens (and it seems to only happen a few times a year).

And finally, it is not necessary to remind me of the benefits of catching errors in CI before merging [every time](https://github.com/micro-manager/mmCoreAndDevices/pull/331#issuecomment-1454273354) this happens. Things are not done the current way because I think the current way is good, and being able to run CI for PRs (without waiting for a full build and with better transparency) was from the beginning one of the primary motivations for the entire [plan](https://github.com/micro-manager/micro-manager/issues/1392). It's hard to retrofit things. But consider me reminded of the importance of making progress on that plan!

henrypinkard:
> could do it cleanly once MMCoreJ can be downloaded as a binary (first few steps of the build system plans)

For what it's worth, we [already have a downloadable MMCoreJ binary](https://s01.oss.sonatype.org/#nexus-search;quick~micro-manager), it just has to be manually uploaded when the core api changes. But Pycromanager and its dependencies rely on this currently. So its possible the java part of this could even be done first, and then the lower level stuff (though I leave it up to you whether it makes sense/offers any advantage)

> And finally, it is not necessary to remind me of the benefits of catching errors in CI before merging [every time](https://github.com/micro-manager/mmCoreAndDevices/pull/331#issuecomment-1454273354) this happens. Things are not done the current way because I think the current way is good, and being able to run CI for PRs (without waiting for a full build and with better transparency) was from the beginning one of the primary motivations for the entire [plan](https://github.com/micro-manager/micro-manager/issues/1392). It's hard to retrofit things. But consider me reminded of the importance of making progress on that plan!

Noted! I have little to no recollection of the previous time I raised this, but if I happen to say the same thing again in the future feel free to save the time of typing out a response and just link me back here :)

PR #1735, opened 2023-07-04
Revert "Allow explore acquisition to zoom before data"

nicost:
Reverts micro-manager/micro-manager#1734 since that code did not compile and MicroMagellan no longer worked.


marktsuchida:
Thanks, build is working now.

PR #1737, opened 2023-07-05
Multi-MDA: Enable different MDA settings at different positions.

nicost:
This PR addresses a very old request: acquire data with different settings at different locations.  An example use case is imaging cells - transfected with different constructs - growing in different wells of a multi-well plate.  This code uses AcqEngJ.  The UI lets the user input timelapse and autofocus settings, then pulls the channel and Z stack settings from Sequence Setting files stored on disk.  A summary for each is given, and the user can specify different position lists for each sequence setting file.  Still complicated, but at least this makes it possible to execute such experiments.  Needs extensive testing, which would benefit from simultaneous testing with the new Acquisition Engine.


marktsuchida (approved):
A frequently requested feature indeed!

It looks great, and I cannot think of any obvious problems. Let's go ahead and merge.

PR #1740, opened 2023-07-08
Adds Acquisition Settings to the Summary Metadata

### LINKED ISSUE #1719: Add Acquisition Settings to SummaryMetadata
We have long wanted to have a "ReUse" button, so that one can apply the acquisition settings of a dataset to the MDA window.  Also, it would be really nice to record the acquisition settings.  It seems that just adding JSON serialized acquisition settings to the Summary Metadata would make both of these things quite easy.  Is there anything against doing this (other than that it may break when the version of the SequenceSettings changes)?

nicost:
Reworked handling of SummaryMetadata in AcqEngJ adapter.  Added a version number to Sequence Settings, and refuse to de-serialize SequenceSettings from JSON if version is higher than the current version.  This should make it possible to "Re-Use" SequenceSettings in the near future.  Closes #1719.


mmstudio/src/main/java/org/micromanager/acquisition/SequenceSettings.java
```diff
@@ -832,13 +834,22 @@ public int acqOrderMode() {
...
 
    public static SequenceSettings fromJSONStream(String stream) {
       Gson gson = new Gson();
-      return gson.fromJson(stream, SequenceSettings.class);
+      SequenceSettings result = gson.fromJson(stream, SequenceSettings.class);
+      double version = result.getVersion();
+      if (version <= Version) {
+         return result;
+      }
+      return null;
```
marktsuchida: I think if we make an incompatible change in the future, `gson.fromJson` could fail before getting to the version check. It might be good to also handle exceptions from `fromJson`.

marktsuchida (approved):
Looks good; apologies for the very late review.

The slight concern would be that this causes the `data` package to (further) depend on other parts of MMStudio (in this case `SequenceSettings`). This could be a problem if we still have the hope of making the data module an independent library some time in the future. But I don't think there is much hope of doing so without making incompatible changes anyway, so I would not worry about it here.

mmstudio/src/main/java/org/micromanager/acquisition/SequenceSettings.java
nicost: Agreed with both.  I'd love to pull out the data package, but that will be a big job in any case, so we can take this one on while doing that project as well.  Catching the exception thrown by fromJson now (not quite sure why the compiler did not insist on me catching that in any case).

PR #1751, opened 2023-08-01
Improvements in remembering display settings

nicost:
This PR was prompted by the "Ignore zero pixel" setting not being remembered (i.e. setting it, closing a window, then opening a new one would not automatically set this checkbox). It turned out that there are various (legacy?) systems to store DisplaySettings, and I surely have not completely untangled these.  However, for Snap and MDA windows, settings (including Channel Display Settings) are now stored in and retrieved from PropertyMaps, which I believe was the goal altogether.  Further clean up (including adding the ability to create a Display while specifying the desired DisplaySettings) is needed, but I believe this is a good step in the right direction until I'll have more time.


marktsuchida (approved):
LGTM.

PR #1754, opened 2023-08-02
Magellan: remove data loading code that initializes viewer into viewer

henrypinkard:


henrypinkard:
Do not merge until new ivy versions from https://github.com/micro-manager/pycro-manager/pull/657 have merged

PR #1758, opened 2023-08-03
Studio: Adds image width and height to SummaryMetadata.

### LINKED ISSUE #1744: Image width and height missing from Metadata
Nenad and I realized that SummaryMetadata do not contain information about image width and height.  The Clojure acquisition engine supplies that information, but it is not converted to SummaryMetadata (or it sometimes can be found as both 0 is the SummaryMetadata,txt file).  This causes several problems, one of them that to determine image size, images actually need to be opened, rather than reading a text file containing the Summary Metadata, the other that it would be really useful for ImageProcessors (which start of with SummaryMetadata) to set things up knowing what image sizes to expect.  

I realize that it is nice for future compatibility not to rely on image size information in SummaryMetadata.  If that is indeed the case, we should at least make it an obligatory element of Image Metadata (where I also often can not find it).  Even though this information is internally present though the Image object, it would be very useful for outside readers to be able to extract this information from the metadata files.

Let me know if it is OK to add this to the SummaryMetadata, and enforce that all code paths actually add it.

nicost:
Checks that sizes of images added to storage are the same as what is listed in the SummaryMetadata (provided it has this data). Bumps the Studio version to 2.0.2 (as there are additions to the SummaryMetadata API).

Closing #1744


marktsuchida (approved):
Thanks, @nicost! LGTM.
Assuming you tested that previously saved files (without the width/height in summary MD) open without issue, I think this is good to merge.

PR #1760, opened 2023-08-14
A plugin for fast and high quality deskewing of light sheet data.

nicost:
This code is implemented as an "On-the-Fly-Processor", to allow for live processing during acquisitions, however, it does not change the data it received in any way, rather produces its own Datastore (and viewer) outputs.  Therefore, it has an option whether or not to keep the original data.  Four different output formats are enables: single plane TIFF, multi-plane TIFF,  RAMM, and Rewriteable RAMM (named "live").   The latter mode can be used for live display of projections.  To facilitate the UI, this plugin brings its own "Process" button for processing of existing data.


marktsuchida (approved):
I see no API changes and the changes to mmstudio appear to be fixes of miscellaneous bugs. The changes to build scripts look good (I'm assuming CLIJ "just works" by installing the JAR).
I have not dived into the actual plugin code much, but let me know if there is anything else in particular you'd like me to take a look at. Otherwise, LGTM.

PR #1761, opened 2023-08-16
Adds option to only show display once images arrive.

nicost:
This PR is prompted by the recent addition of the Deskew plugin.  The Deskew plugin is a Processor in the on-the-fly-processor pipeline that uses images to create its own Datastores, and optionally does not propagate images to original Datasink.  This works well, but does create display windows that never receive images, which is a bit of a bad user experience.

This PR is more of a discussion piece.  It adds an option to the API to only show the display once images actually arrive.  It works well for the use case described.  Another option would be to leave the API unchanged, and change the behavior of the current method "show()", so that displays are always only shown once images actually arrive.

I am not quite sure what happens to these "orphaned" displays and associated Datastores.  Are they garbage collected?  I currently do not have a Java Profiler working, so it is a bit hard to investigate. 

Looking for suggestions for the best path forward.

### DIFF
--- a/mmstudio/src/main/java/org/micromanager/display/DisplayWindow.java
+++ b/mmstudio/src/main/java/org/micromanager/display/DisplayWindow.java
@@ -324,4 +324,11 @@ public interface DisplayWindow extends DataViewer, Closeable {
     * construction, and after attaching listeners as needed, to show the Window.
     */
    void show();
+
+   /**
+    * DisplayWindows are not shown by default.  Call this function after
+    * construction, and after attaching listeners as needed, to show the Window.
+    * This version will only show the display window once there are images available.
+    */
+   void showOnceThereAreImages();
 }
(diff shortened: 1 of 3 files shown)


marktsuchida:
Hmm. I guess a fundamental fix would involve making the "destination" for acquisitions configurable, such that the datastore (with or without display) is just the default. Then there would be no overhead of creating a datastore and display that is never used (which is probably a good thing, especially when the datastore would create files or a folder).

It would definitely be nice to come up with a way to address the fundamental problem rather than painting over it (so to speak). Then again, from a pure design perspective, it would be best if processors didn't do things other than processing (like creating datastores).

Maybe the Deskew plugin could be better modeled as something that replaces the datastore downstream of the processing pipeline, rather than as a processor? For example, in terms of UI, we could imagine having an "Other..." button next to the file format selection in the MDA dialog that would allow swapping the data destination for an acquisition.

Current: `[acquisition engine (selectable)] -> [processing pipeline (configurable)] -> [datastore and display (hard-coded)]`
Potential design: `[acquisition engine (selectable)] -> [processing pipeline (configurable)] -> [datastore and display (pluggable)]`

As an aside, the display window used to (in 1.4 and I think early 2.0 beta) not show the window until there was an image. I remember deliberately changing that to open the window early, because otherwise it looks like nothing is happening when the MDA is such that it takes a significant amount of time before the first image is available. Not that this is directly relevant here, since you are not proposing to change the default behavior.

PR #1765, opened 2023-08-19
Formalize dependency updates caused by Clij2

nicost:
Also exclude unused imports caused by Clij2.  Rotate GPU deskew over its x axis, to match CPU orientation.


marktsuchida:
Many thanks for this, @nicost. Saved >12 MiB off the installer.

Issue #916, opened 2020-08-26
Allow plugin classes (and resources) to be accessed from outside


marktsuchida:
Currently, each plugin JAR (_not_ each plugin class) is given a separate class loader (whose parent is `MMStudio`'s class loader, which is also the ImageJ1 class loader).

This was done for two reasons:
1. To speed up the plugin loading process by preventing SciJava's `DefaultPluginFinder` from scanning every single JAR on the classpath (this was the initial motivation).
2. To forbid other code from depending on plugin-internal code.

While reason 2. is still valid, it would be nice to (a) be able to quickly test plugin code by invoking it from scripts and (b) be able to have single JARs that contain both library code (APIs intended for external use) and plugins.

Furthermore, the use of individual class loaders complicates reflective code (such as the ZeroMQ server).

I think (hope) that we can ignore any performance concerns, because Fiji launches fast enough these days, at least the second time, and in any case our future should be to use the ImageJ2 launcher.

So **we want to simplify the situation by putting everything on the classpath**. This is in fact closer to how the SciJava plugin mechanism is intended to be used, so it will also help when we move to shipping ImageJ2 instead of ImageJ1.

There is just one problem: currently, when MM is launched by double-clicking its icon, the classpath consists exactly of `ij.jar` (only), because that's how ImageJ1 works. (The classpath is whatever you set it to be if you launch from an IDE, of course.)

So it would appear that we need to
- Migrate to the ImageJ2 launcher (which puts all available JARs on the classpath),
- Move all plugin JARs (from `mmplugins` and `mmautofocus`) into the classpath (`jars` directory), and
- Update our plugin loading code (mostly just delete the JAR enumeration code we have and replace with straightforward calls to SciJava).

This, of course, is a rather large undertaking (although certainly desirable in the long term). Among many other things, it requires that we get all of our dependencies to be at compatible versions with ImageJ2.

A shorter-term, easier fix might be to create our own "main" class loader (as a child of the ImageJ1 one), and load everything (including MMStudio itself) using it. I'm not 100% sure what the consequences of this would be -- I would worry in particular about ImageJ macros accessing MM code.

Unfortunately for us, it is not possible to change either the classpath or the URLs searched by an existing `URLClassLoader` at runtime.

There might be some other way to hack this, so I'll keep thinking, but I'm out of ideas for now. Suggestions would be highly appreciated.

(Mildly related: #331.)

nicost:
>     To forbid other code from depending on plugin-internal code.
[...]

PR #1773, opened 2023-08-25
Studio: Use a single classloader for all plugins.

nicost:
This let''s plugins see each other.

### DIFF
--- a/mmstudio/src/main/java/org/micromanager/internal/pluginmanagement/PluginFinder.java
+++ b/mmstudio/src/main/java/org/micromanager/internal/pluginmanagement/PluginFinder.java
@@ -72,33 +72,31 @@ public final class PluginFinder {
     */
    public static List<Class> findPlugins(String root) {
       ArrayList<Class> result = new ArrayList<>();
+      List<URL> jarURLs = new ArrayList<>();
       for (String jarPath : findPaths(root, ".jar")) {
-         URL jarURL;
          try {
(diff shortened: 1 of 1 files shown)


marktsuchida (approved):
Ha! This is cleaver. I didn't think of the option of letting plugins call each other without letting anything else call them (by putting them on the classpath).

Is this sufficient to close #916? Does it solve the problems when running from an IDE?

I'm okay with merging this either way.

nicost:
Not quite sure of it either, this just occurred to me as a simple implementation to get all the plugins on the same classloader.  I believe this will still not make it possible to call plugins from scripts (which is a shame), but should let plugins call each other (not tested).  I'll dig a little more before merging.

nicost:
For scripts, we can use the Beanshell `addClassPath()` function:

```
bsh % import org.micromanager.hcs.SBSPlate;
bsh % plate = new SBSPlate();
// Error: EvalError: Unknown class: SBSPlate : at Line: 5 : in file: <unknown file> : new SBSPlate ( ) 

bsh % addClassPath("C:\\Program Files\\Micro-Manager-2.0\\mmplugins\\HCS.jar");
bsh % import org.micromanager.hcs.SBSPlate;
bsh % plate = new SBSPlate();
<96-Well>
```

A bit weird, since this will be a different thing than the SBSPlate loaded through Micro-Manager's GUI, but should be useful.

PR #1774, opened 2023-08-26
Replace AcqListeners with AcquisitionSettingsChangeEvent in API.

### LINKED ISSUE #1743: Access to AcquisitionEngine
Dear Micro-Manager team,

One of my plugins wants to know if changes are made in the MDA window. I have been using the AcquisitionEngine to add a SettingsListener to get events for changes in the settings.
`studio.getAcquisitionEngine().addSettingsListener(this);`

The getAcquisitionEngine funcionality seems to have been removed from the API completely. It would be great to have another way to receive changes in the MDA window as they happen.

See image.sc post below.
Thanks,
Willi

Image.sc discussion:
https://forum.image.sc/t/stopping-acquisition-programmatically/79056/3

nicost:
This simplifies the code, and enables external code to listen for changes in the MDA window.
Also avoid updating the GUI twice when a setting changes.
Bumps version to 2.0.3 since there is an API addition.
Closes #1743


mmstudio/src/main/java/org/micromanager/acquisition/internal/DefaultAcquisitionSettingsChangedEvent.java
```diff
@@ -0,0 +1,41 @@
...
+public class DefaultAcquisitionSettingsChangedEvent implements AcquisitionSettingsChangedEvent {
+   private final SequenceSettings newSettings_;
+
+   public DefaultAcquisitionSettingsChangedEvent(SequenceSettings newSettings) {
+      newSettings_ = newSettings;
+   }
+
+   @Override
+   public SequenceSettings getNewSettings() {
+      return newSettings_;
```
marktsuchida, line 39: (Comment only, no action needed.) I was a little worried about this because `SequenceSettings` is technically mutable (mutable public fields are deprecated). But it looks like most (all?) of our own use of it is updated to avoid mutating it in place. So it should be okay as long as new code avoids accessing the deprecated public fields.

marktsuchida (approved):
Looks good!

nicost:
On a related note, when can we remove deprecated code from the API?  The 2.0 release is now 2 years old, so that should have given everyone enough time to stop using the deprecated parts.  What should the version be after doing so?

marktsuchida:
@nicost The version should be 3.0 after removing deprecated APIs. I agree that it has been long enough.

PR #1777, opened 2023-08-29
Adds ability to read Sequence Settings from data in Viewer and apply to MDA window.

nicost:
Fixes some issue adding SequenceSettings to the SummaryMetadata (previously only in User data?).  The MDA window now checks each DataViewer that became active and enables/disables the ReUse button based on the ability to apply the SequenceSettings.  It checks Sequence settings by seeing if the Channel Group and Channel names match the currently loaded configuration.  More check could (should?) be added.

### DIFF
--- a/mmstudio/src/main/java/org/micromanager/internal/dialogs/AcqControlDlg.java
+++ b/mmstudio/src/main/java/org/micromanager/internal/dialogs/AcqControlDlg.java
@@ -95,6 +95,7 @@ import org.micromanager.data.internal.DefaultDatastore;
 import org.micromanager.display.ChannelDisplaySettings;
 import org.micromanager.display.DataViewer;
 import org.micromanager.display.internal.RememberedDisplaySettings;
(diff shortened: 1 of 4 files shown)


mmstudio/src/main/java/org/micromanager/internal/dialogs/AcqControlDlg.java
```diff
@@ -846,9 +852,49 @@ private JPanel createSaveButtons() {
       saveAsButton.setMargin(new Insets(-5, -5, -5, -5));
       saveAsButton.addActionListener((ActionEvent e) -> saveAcqSettingsToFile());
       result.add(saveAsButton, BUTTON_SIZE);
+
+      reUseButton_ = new JButton("ReUse Settings");
```
marktsuchida: Nitpick: "Reuse" is usually one word in English.

marktsuchida (approved):
LGTM.

PR #1789, opened 2023-09-08
Staying current with AcqEngJ API changes.

nicost:
I wish there was a stable API, and that every other AcqEngJ change would not break the MMStudio build....

### DIFF
--- a/mmstudio/src/main/java/org/micromanager/acquisition/internal/acqengjcompat/AcqEngJMDADataSink.java
+++ b/mmstudio/src/main/java/org/micromanager/acquisition/internal/acqengjcompat/AcqEngJMDADataSink.java
@@ -90,10 +90,10 @@ public final class AcqEngJMDADataSink implements AcqEngJDataSink {
    }
 
    @Override
-   public Object putImage(TaggedImage tagged) {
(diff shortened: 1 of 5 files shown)


nicost:
Count me confused.  What is going on @henrypinkard?  What is the signature of putImage() in AcqEngJDataSink?  Is changing this really worth the pain?

nicost:
Ajj.  Guess I figured this out.  My Intellij includes the sources to AcqEngJ, and updating the sources brought things back together.  Stable APIs would be so nice....

henrypinkard:
> I wish there was a stable API, and that every other AcqEngJ change would not break the MMStudio build....

Have you seen this happening? If so, something is amiss because [this test](https://github.com/micro-manager/micro-manager/actions/runs/6126150137/job/16629560934?pr=1789) that runs on PRs is supposed to catch it, as it did in this case. 

There was no way to make the change I did (which enabled a really useful new feature--asynchronous notifications from the acquisition engine) without updating the code that uses AcqEngJ. Changes like this do not happen often.

henrypinkard:
I will make sure to tag you if something like this merges in the future to avoid confusion

PR #1817, opened 2023-10-13
Studio: putative fix for occassional PropertyBrowser crash on exit.

nicost:


nicost:
Difficult to test since the null pointer was irregular, likely due to the random order things were happening on shutdown.

PR #1838, opened 2023-11-16
No longer produce many metadata.txt files with identical content.

nicost:
For multi-position acquisitions with separate files per position and saving metadata.txt files, only produce one metadata.txt rathern producing identical copies with the name of each tiff file.

### DIFF
--- a/mmstudio/src/main/java/org/micromanager/data/internal/multipagetiff/StorageMultipageTiff.java
+++ b/mmstudio/src/main/java/org/micromanager/data/internal/multipagetiff/StorageMultipageTiff.java
@@ -438,7 +438,7 @@ public final class StorageMultipageTiff implements Storage {
       if (!positionToFileSet_.containsKey(fileSetIndex)) {
          positionToFileSet_.put(fileSetIndex,
                new FileSet(image, this, omeMetadata_,
-                     splitByXYPosition_, separateMetadataFile_));
+                     splitByXYPosition_, fileSetIndex == 0 && separateMetadataFile_));
       }
       FileSet set = positionToFileSet_.get(fileSetIndex);
 


henrypinkard:
LGTM

marktsuchida (commented):
Wouldn't the correct fix be to have each metadata file only contain metadata for that position? That way, reader code doesn't need to know anything special about a dataset that happens to be a position in a multi-position acquisition.

But even that change is probably going to [break](https://xkcd.com/1172/) some people's workflow, if they [depend](https://www.hyrumslaw.com/) on being able to read a single metadata.txt to get data for all positions.

Is this motivated by avoiding redundancy or by saving space? If the former, I wouldn't make this change because it doesn't add any functional advantage but will probably waste people's time to adapt. If the latter, well, people can turn off the writing of metadata.txt and fetch it from inside the TIFF files -- although that's probably not very satisfactory.

Maybe if saving space is important it makes sense to go forward with this change, but document it on the [formats](https://micro-manager.org/Micro-Manager_File_Formats) page.

nicost:
Space and CPU cycles.  It is really convenient to have a text file with metadata to parse.  It is kind of crazy to have copies of the exact same file for each position (and there easily can be thousands of positions per acquisition).  I think that the original behavior was a bug (or un-intended side effect).  I agree that backward compatibility is an issue.  Micro-manager reads the data easily (I don't think it even looks at the Metadata.txt file), but who knows what other readers do.  I do not think that it is a good idea to keep this kind of crazy situation.  It makes a lot of sense to me to have only one metadata.txt file per dataset.

henrypinkard:
> I think that the original behavior was a bug (or un-intended side effect).

I would guess an unintended side effect

> (I don't think it even looks at the Metadata.txt file),

I believe thats right, and there aren't any things within MM that read this file, because it is also contained in the Tiff files themselves

marktsuchida:
Yes, the metadata.txt is entirely for users, never read by MM (at least for MultipageTIFF). So the whole point of it is to support use cases other than the MM reader. Otherwise one can disable its creation.

I can see that it would be convenient to have a single file when the multi-position dataset is a single entity to be treated as a whole during later analysis. But that is not what the user interface says: "Save XY positions in separate Image Stack Files" suggests that the result is a collection of individually valid (stand-alone) multipage-tiff datasets. And "Create metadata.txt file with Image Stack Files" suggests that the metadata.txt is per dataset.

The problem is that there are people who use multi-position acquisition not to tile a large specimen (resulting in a single dataset, conceptually), but to manually select a few fields of view and time-laps, cycling between them. It is then important that each of the several datasets produced that way behave independently as a stand-alone Micro-Manager dataset, because they may be processed later as separate datapoints.

So I agree that the behavior is a bug that should be fixed, but the bug is not that each dataset gets a metadata.txt; the bug is that each metadata.txt duplicates the metadata that only applies to the other datasets. (And yes, I do realize this fix would be much more work.)

But... I guess I'm not an absolutist about this. The main point I wanted to make is that this is almost certainly going to break some people's workflow (and I have no way of knowing how common that will be). If you think this is the best fix despite that, then I'm okay with it, so long as
- The change is well documented and
- The produced single metadata.txt file's name does not contain the name of the first position

E.g., currently I get `test_MMStack_Pos0_metadata.txt` and `test_MMStack_Pos1_metadata.txt` in a two-position acquisition. The new, single file should be named `test_MMStack_metadata.txt`, not `test_MMStack_Pos0_metadata.txt`.

nicost:
I now realize the copied of metadata.txt are actually not identical.  I was fooled but the Framekey.  Turns out that the Framekey does not have an entry for position, hence the image metadata for each position are different, and are contained in the corresponding metadata.txt file.  Probably best to not change the current situation.  Stuffing all these in one and the same metadata.txt file will be complicated and error prone and probably not be worth the effort at this point in time.

PR #1847, opened 2023-11-30
Snap on move improvements

### LINKED ISSUE #1823: Snap-on-Move Preview doesn't remember movement devices
After restarting micro-manager, the Snap-on-Move Preview doesn't remember which devices to poll. It'd be nice if I didn't have to re-add the devices each time I restart the software. 

Additionally, it'd be nice if I didn't have to remember to disable the Snap-on-Move Preview when I start an MDA. This may be too much to ask, so feel free to ignore this feature request if it's too difficult to implement.

nicost:
Closes #1823.  Now remembers movement criteria between MM invocations.  Also automatically stops operation while an acquisition is running.

@marktsuchida, could you check if the way I handle pausing the monitoring thread is correct?  It works OK with the demo config.

I'd be happy to move this out of the beta menu item, probably into "Device Control".

### DIFF
--- a/plugins/SnapOnMove/src/main/java/org/micromanager/plugins/snaponmove/MainController.java
+++ b/plugins/SnapOnMove/src/main/java/org/micromanager/plugins/snaponmove/MainController.java
@@ -44,7 +44,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.TimeUnit;
-import java.util.concurrent.atomic.AtomicBoolean;
 import mmcorej.CMMCore;
 import org.micromanager.PropertyMap;
 import org.micromanager.PropertyMaps;
@@ -93,12 +92,13 @@ class MainController {
    private final LinkedBlockingQueue<Map.Entry<MonitoredItem, MonitoredValue>>
          eventQueue_ = new LinkedBlockingQueue<>();
 
-   // The monitoring thread; null when not running (disabled)
+   // The monitoring thread; null when stopped or paused
    private Thread monitorThread_;
 
-   private final AtomicBoolean pausedForAcquisition_ = new AtomicBoolean(false);
+   private boolean pausedForAcquisition_ = false;
+
+   // Whether monitoring is "enabled" (whether running or paused)
    private boolean monitorEnabled_ = false;
-   private final Object pauseLock_ = new Object();
 
    private UpdatableAlert statusAlert_;
 
@@ -181,12 +181,12 @@ class MainController {
 
    synchronized void setEnabled(boolean f) {
       monitorEnabled_ = f;
-      if (!pausedForAcquisition_.get()) {
+      if (!pausedForAcquisition_) {
          handleMonitorThread(f);
       }
    }
 
-   synchronized void handleMonitorThread(boolean f) {
+   private synchronized void handleMonitorThread(boolean f) {
       if (f == (monitorThread_ != null)) {
          return;
       }
@@ -239,10 +239,15 @@ class MainController {
             synchronized (latestValues_) {
                lastSnapValues_.putAll(latestValues_);
             }
-            if (!pausedForAcquisition_.get()) {
+
+            boolean skipSnap = false;
+            synchronized (this) {
+               // Skip the first snap after resuming from pause.
+               skipSnap = pausedForAcquisition_;
+               pausedForAcquisition_ = false;
+            }
+            if (!skipSnap) {
                doSnap();
-            } else {
-               pausedForAcquisition_.set(false);
             }
             waitForChange();
          }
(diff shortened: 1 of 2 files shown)


plugins/SnapOnMove/src/main/java/org/micromanager/plugins/snaponmove/MainController.java
```diff
@@ -387,6 +392,9 @@ private void initializeMonitoredValues() {
 
    @Subscribe
    public void onZMoved(StagePositionChangedEvent e) {
+      if (pausedForAcquisition_) {
```
marktsuchida: Best to consistently synchronize on `pauseLock_` when accessing `pausedForAcquisition_`.

plugins/SnapOnMove/src/main/java/org/micromanager/plugins/snaponmove/MainController.java
```diff
@@ -403,14 +411,17 @@ public void onZMoved(StagePositionChangedEvent e) {
...
-         eventQueue_.put(new AbstractMap.SimpleEntry<MonitoredItem, MonitoredValue>(item, value));
+         eventQueue_.put(new AbstractMap.SimpleEntry<>(item, value));
       } catch (InterruptedException unexpected) {
          Thread.currentThread().interrupt();
       }
    }
 
    @Subscribe
    public void onXYMoved(XYStagePositionChangedEvent e) {
+      if (pausedForAcquisition_) {
```
marktsuchida: Ditto.

plugins/SnapOnMove/src/main/java/org/micromanager/plugins/snaponmove/MainController.java
```diff
@@ -440,6 +451,19 @@ public void onShutdownCommencing(ShutdownCommencingEvent e) {
       }
    }
 
+   @Subscribe
+   public void onAcquisitionStarted(AcquisitionStartedEvent e) {
+      pausedForAcquisition_ = true;
```
marktsuchida, line 462: And here, too.

plugins/SnapOnMove/src/main/java/org/micromanager/plugins/snaponmove/MainController.java
```diff
@@ -440,6 +451,19 @@ public void onShutdownCommencing(ShutdownCommencingEvent e) {
...
+   }
+
+   @Subscribe
+   public void onAcquisitionEnded(AcquisitionEndedEvent e) {
+      synchronized (pauseLock_) {
+         pauseLock_.notify();
+      }
+      pausedForAcquisition_ = false;
```
marktsuchida: The setting to `false` should be inside the `synchronized` block and _before_ the call to `notify()`.

plugins/SnapOnMove/src/main/java/org/micromanager/plugins/snaponmove/MainController.java
```diff
@@ -235,6 +234,12 @@ private void monitorLoop() {
             }
             doSnap();
             waitForChange();
+            synchronized (pauseLock_) {
+               while (pausedForAcquisition_) {
+                  pauseLock_.wait();
+                  waitForChange();
```
marktsuchida: It is better to avoid calling `waitForChange()` (which blocks for arbitrary duration) while holding `pauseLock_`, because that prevents another thread from clearing `pausedForAcquisition_`.
Best to set a boolean flag `shouldWaitAgain` inside the `synchronized` and call `waitForChange()` after exiting the `synchronized` when the flag is set.
Alternatively, it's probably okay to just leave out the `waitForChange()` here and let the next snap happen after an acquisition finished (though it would be good to test that this doesn't cause issues).

marktsuchida (changes requested):
Thanks so much for taking care of this! (Sorry I hadn't gotten around to it.)
The general structure looks good to me; added comments on threading issues.

nicost:
Whatever I tried using this approach (see https://github.com/nicost/micro-manager/tree/snapOnMove2), I can not avoid the Snap at the acquisition.  I really do not like that one, since it takes the focus away from the just acquired data, and will be quite annoying/confusing to users.  

Instead, I tried switching `pausedForAcquisition_ `to an AtomicBoolean.  That works, and does allow another thread to clear `pausedForAcquisition_`.   This makes it possible to move the second `waitForChange()` outside of the synchronized block, forcing another `waitForChange()` after the acquisition ends without blocking other threads to clear `pausedForAcquisition_`. 

Pushing current, working, code now.

marktsuchida:
I'm 60% sure this has no bugs, but it's confusing that an independent atomic variable is used when a regular boolean protected by the same lock should work (and be definitely correct). Because the two are not synchronized together, you need to make assumptions about how `onAcquisitionStarted` and `onAcquisitionEnded` are called to reason about the behavior.

Your snapOnMove2 branch is a bit different from my suggestion.... I meant something like:

```java
         for (; ; ) {
            lastSnapValues_.clear();
            synchronized (latestValues_) {
               lastSnapValues_.putAll(latestValues_);
            }
            doSnap();
            waitForChange();

            boolean waitedForAcqFinish = false;
            synchronized (pauseLock_) {
               while (pausedForAcquisition_) {
                  pauseLock_.wait();
                  waitedForAcqFinish = true;
               }
            }
            if (waitedForAcqFinish) {
               waitForChange();
            }
         }
```

This is exactly like your first version, except for the second call to `waitForChange()` being moved out of the `synchronized`.

marktsuchida:
Sorry I didn't get back to this today, but will take another look (with actual testing on my side). I must be misunderstanding something -- but I also suspect it will be better to stop the polling thread entirely when pausing for acquisition.

nicost:
Yes, that may be the better solution.  Needs a bit or reworking of the code to separate starting/stopping the monitorthread from registering for events from Studio.

nicost:
Latest push has those changes.  Much simpler, and works well, even when starting SnapOnMove while an acquisition is running.

marktsuchida:
Ready to merge.

nicost:
Perfect!  Thanks!

PR #1862, opened 2024-01-09
Optimizes Storage:getImagesIgnoringAxes implementations.

### LINKED ISSUE #1858: MDA slows significantly when displaying last image
Tagging @nicost here because he also witnessed the issue. 

Essentially, when I am acquiring a timelapse with a large number of slices (>100 or so), I notice a gradual increase in the inter-timepoint delay _only when the display window is set to display the last image._ The inter-timepoint delay starts out normally (< 1 s), but it can increase to as much as 10 s when the display window is trying to display every image.

This significantly slows down the empty rate of the sequence buffer. The problem is ameliorated when I disable the display update. 

Is there any way to avoid this slowdown?

nicost:
Closes #1858. The implementations of the function Storage:getImagesIgnoringAxes would iterate over all known coords to check if they match or not.  This is an expensive operation, especially for large datasets.  Our code only uses this function with the axis Channels as the ignored axes, so I added code to store those in a HashMap for fast retrieval.  This change avoids the slow down in display speed for large datasets.

The function GetImagesMatching appears to be almost the same in intent.  I changed its implementations to call getImageIgnoringAxes wherever appropriate.   

There is quite a bit of redundancy in this code.  I played with creating a special class for handling Coords for all Storage implementations, but this was not very pleasant either.

### DIFF
--- a/mmstudio/src/main/java/org/micromanager/data/internal/StorageRAM.java
+++ b/mmstudio/src/main/java/org/micromanager/data/internal/StorageRAM.java
@@ -25,6 +25,7 @@ import java.io.IOException;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
 import java.util.TreeSet;
 import org.micromanager.data.Coords;
@@ -47,6 +48,7 @@ import org.micromanager.data.SummaryMetadata;
  */
 public final class StorageRAM implements RewritableStorage {
    private HashMap<Coords, Image> coordsToImage_;
+   private Map<Coords, List<Coords>> coordsIndexedMissingC_;
    private Coords maxIndex_;
    private SummaryMetadata summaryMetadata_;
    private final Set<String> axesInUse_;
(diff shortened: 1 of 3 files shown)


mmstudio/src/main/java/org/micromanager/data/internal/StorageRAM.java
```diff
@@ -121,15 +131,12 @@ public synchronized List<Image> getImagesMatching(Coords coords) {
...
-         if (image.getCoords().equals(coords)) {
-            results.add(image);
-         }
+      try {
+         return getImagesIgnoringAxes(coords, coords.getAxes().toArray(new String[0]));
+      } catch (IOException ex) {
+         ReportingUtils.logError(ex, "Failed to read image at " + coords);
+         return null;
       }
-      return results;
```
marktsuchida: The reason 'subspace' (a term I made up, vaguely inspired by linear subspaces) was used is because the method was supposed to get all images that match `coords` on all the axes contained in `coords`, but the images may have additional coordinates. In other words, `getImagesMatching()` was supposed to return the images matching `coords`, ignoring axes that do _not_ appear in `coords`.

The previous code here (without the 'subspace') seems incorrect (fails to ignore axes not in `coords`), assuming `Coords.equals()` is defined as originally intended (exact match not considering axis order). Maybe this is why it is no longer used much?

I'm not sure I understand the new code -- wouldn't ignoring all the axes in `coords` mean always matching every image? Or am I missing something? (I would have thought that the axes to ignore should be those that appear in the storage but not in `coords`.)

Given that this method doesn't seem to have done what it was supposed to do for a long time, it may indeed make more  sense to deprecate it rather than totally change its behavior. Especially given that the original intended use cases of `getImagesMatching()` are now well covered by `getImagesIgnoringAxes()`.

(All this also applies to the other 2 storage implementations, of course.)

marktsuchida (commented):
I think the idea of special-casing channels makes plenty of sense, given how the viewer works. Nice optimization.

marktsuchida:
> There is quite a bit of redundancy in this code. I played with creating a special class for handling Coords for all Storage implementations, but this was not very pleasant either.

Yeah, the problem is in the abstraction of `Storage`, which doesn't really separate concerns very well. I think it makes sense to have some duplication here for now.

PR #1883, opened 2024-02-01
CRISP Plugin: fix for Update Rate [ms] ui control with new MS2000 firmware versioning convention 9.50+

bls337:
better tooltips
use doubles for CRISP object


nicost:
There is some documentation on how to set this up here: https://github.com/micro-manager/micro-manager/blob/main/doc/pre-commit.md.  It is a bit annoying to set up, but in the long run has helped keep things easy to work with...

bls337:
@nicost Thanks! I was going to ask about where to find the CheckStyle file but you beat me to it! I like to use stuff like this for other programming languages, ruff is a good option for Python.

PR #1887, opened 2024-02-05
Adds TestAcquisition to the MDA and makes this an option in the SnapOnMove plugin.

nicost:
This work was started with the goal to provide a way to provide a semi-live Deskewed view for the Snoutscope LIghtSheet microscope.  The Deskew plugin is great to give several views of LIghtsheet data that are more useful to humans than the raw, skewed data, but such views were not available while scanning through the slide.

This PR contains two parts to enable such views.  First, it adds a "Test Acquisition".  This is an acquisition with only the Z and Channel (i.e. it takes place at the current location and ignores time lapse settings).  Since this seemed handy also in other situations, I added the code to the acquisition/internal package, and made a button in the AcqControl window to use it.  

The other part is in the SnapOnMove plugin.  This now offers the option to use either Snap or Test Acquisition once the movement criteria are full-filled.  Together, these additions enable the desired behavior.

There are some usability issues.  One I am aware off is that changes in the pipeline (for instance, changes in the output of the Deskew plugin) are not immediately reflected in the output.  The user would need to close the TestAcq window for these changes to be taken into account.  I could provide a better way if there were a mechanism to find out if a pipeline changed.

The PR also contains a bug fix for the AcqEngJ code.

### DIFF
--- a/mmstudio/src/main/java/org/micromanager/acquisition/internal/acqengjcompat/AcqEngJAdapter.java
+++ b/mmstudio/src/main/java/org/micromanager/acquisition/internal/acqengjcompat/AcqEngJAdapter.java
@@ -786,7 +786,7 @@ public class AcqEngJAdapter implements AcquisitionEngine, MMAcquistionControlCal
             }
             try {
                if (when == AcquisitionAPI.BEFORE_HARDWARE_HOOK) {
-                  if (event.isZSequenced() || event.getZIndex() == 0) {
+                  if (event.getZIndex() == 0) {
                      if (!event.isZSequenced() && sequenceSettings.useChannels()
                              && (sequenceSettings.acqOrderMode()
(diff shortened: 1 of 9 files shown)


nicost:
>  It would be best if the ConfigFrame did not have access to MMStudio or the settings and instead called the controller.

Yes!  I'll get that change in soon.

One idea would be to move TestAcquisition to the Snap/Live are of the code.  It has some similarities to Snap/Live.  Not sure though, since there are many dependencies on acquisition as well.

marktsuchida:
Thanks! Yes, I think okay to merge.

I don't think we want Snap/Live to know about any concrete acquisition components. It might be possible to make the "snap" action pluggable in Snap/Live, so that the snap button performs a test-acq. Not sure how easy, though.

PR #1899, opened 2024-02-20
Datastore close on save

jsokoloff23:
RAMDatastore has option to save asynchronously but currently doesn't have an option to close datastore when it is finished. This adds a new save() method to DefaultDatastore class with parameter closeOnSave so that datastore can be deleted without close() calling from the main thread.

Notes: 

I'm not sure if this is an issue only for me because of how I've written our acquisition code with Pycro-Manager. To me, this seems like this would be the default behavior, but I suppose it's not because the displays from MDA are RAMDatastores, and it would be weird to close immediately after saving. 

Anyway, I've tested this with the nightly build and it works as intended--RAMDatastore is closed after saving is completed.

### DIFF
--- a/mmstudio/src/main/java/org/micromanager/data/Datastore.java
+++ b/mmstudio/src/main/java/org/micromanager/data/Datastore.java
@@ -169,6 +169,18 @@ public interface Datastore extends DataProvider {
    void save(SaveMode mode, String path, boolean blocking) throws IOException;
 
    /**
+    * Saves the datastore to the given path using the given format (SaveMode).
+    *
+    * @param mode     File format to save to
+    * @param path     File path used to save the data
+    * @param blocking when true, will block while saving data, otherwise will return
+    *                 immediately
+    * @param closeOnSave when true, will close datastore after saving.
+    * @throws java.io.IOException if an IO error occurs
+    */
+   void save(SaveMode mode, String path, boolean blocking, boolean closeOnSave) throws IOException;
+   
+   /**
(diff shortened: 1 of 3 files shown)


marktsuchida:
I'm not sure I fully understand the context in which this is needed, but it would be better to name the method something like `saveAndClose` rather than add an overload with an extra boolean flag (the calling code would be more readable).

The datastore for MDA is RAM only when saving is disabled -- but perhaps that's not what you meant?

jsokoloff23:
We use Pycro-Manager to run MM scripts to take z-stacks of whole larval zebrafish guts at 90 second intervals for four or eight hours at a time. Ideally, this interval would be lower but our current bottleneck is writing images. Currently, we either use MultipageTIFFDatastore and wait for the images to save at each imaging region (the buffer has to empty before another sequence acquisition can start), or we use a RAMDatastore and save at the very end of each timepoint, which improves our temporal resolution between imaging regions but takes longer overall because we aren't writing to the disk while images are being acquired.

The most obvious and simplest solution would be to just get some large SSDs to increase the write speed, but we often produce 20TB or more data per week and don't have enough time in between acquisitions to transfer data to our servers, making this not the greatest solution for us. And it would be a bit spendy.

This change is just to enable us to save images during the entire acquisition so we can maintain the temporal resolution between regions while also saving during the acquisition.

I hope all that makes sense!

I'd be glad to create a new method instead of overloading. I wasn't sure which implementation would be more appropriate.

PR #1914, opened 2024-04-05
Magellan: factor out studio from Magellan api.  

nicost:
This should make it possible to use the Magellan plugin without the Studio environment, but whether or not that really works should be tested.


henrypinkard:
Thanks!

PR #1922, opened 2024-04-11
Many improvements in AcqEngJ compat layer

nicost:
Should only be merged after AcqEngJ 0.37.0 is released.  
- Repositions stages at the end of acquisition
- Actually shows error, rather than silently doing nothing.
- When using position list, moves all drives rather than only XY drives
- When using position list, updates positions after autofocus, so those are used at the next time point.


henrypinkard:
Looks good. 

I think you could manually change the version number needed of AcqEngJ in ivy, then it will only pass the tests when the new version is posted. 1 less manual step to do that way

PR #1927, opened 2024-04-16
Studio: MultipageTiff reader: re-use ByteBuffer instead of allocating a new one for each image. 

nicost:
Also uses allocateDirect rather than allocate.
2250 x 255
Testing AllocateDirect versus Allocate on a dataset of 192 images 2250 x 2250 x 2 bytes (i.e. 192 * 10.125MB = 192 MB) with the script: "

```
start = System.currentTimeMillis();
mm.data().loadData("C:\\Users\\Nico Stuurman\\IMages\\imageData", false);
end = System.currentTimeMillis();
mm.scripter().message("It took: " + (end - start) + " ms");
```

10 times for each gave:

Old code: 6.2 +- 0.3 seconds
Allocate: 6.2 +- 0.9 seconds
AllocateDirect: 6.0 +- 0.83 seconds

Interestingly when run under the debugger in IntelliJ, Allocate was much slower than Allocate Direct (9.1 +- 2.0 seconds).

Not a huge difference, but the more images are in the dataset, the higher the performance improvement should be.  Also, this should create less work for the garbage collector.  Lastly, this seems an obvious thing to do, with the possible downside that it does not allow for multiple simultaneous calls to readImage (which seems a dangerous, untested thing to do in any case, so making this function synchronized may even be helpful).

### DIFF
--- a/mmstudio/src/main/java/org/micromanager/data/internal/multipagetiff/ImageByteBuffer.java
+++ b/mmstudio/src/main/java/org/micromanager/data/internal/multipagetiff/ImageByteBuffer.java
@@ -1,0 +1,52 @@
+package org.micromanager.data.internal.multipagetiff;
+
+import java.nio.ByteBuffer;
+import java.nio.ByteOrder;
+
+/**
+ * A simple class to hold a ByteBuffer and its size.
+ */
+public final class ImageByteBuffer {
+   private final ByteBuffer buffer_;
+   private final int size_;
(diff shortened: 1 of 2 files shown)


henrypinkard:
Seems reasonable. I put a lot of effort into optimizing this type of thing in NDTiff, you could look at what's there for further insight

marktsuchida (approved):
LGTM!

Issue #1937, opened 2024-05-01
Feature request: enable focus maintenance during MDA z-stack


fadero:
I'd like to request a feature where a focus maintenance device (e.g. ASI's CRISP, Nikon's PFS) is allowed to remain enabled during an MDA with a z-stack. Currently, the MDA disables the focus maintenance device.

I'm asking for this because the core focus I'm using (galvo mirror) is not the same device that is maintaining focus (ASI FTP with CRISP). So I'd like to keep the CRISP continuously active while my galvo mirror moves.

@nicost

nicost:
Implemented in #1938, use: `mm.settings().setUnlockAutofocusDuringZStack(false)`.

PR #1938, opened 2024-05-02
Studio: Make it optional to unlock autofocus during Z stack.

nicost:
Needed in some unexpected situations.  Use the Options mechanism, but do not expose in Options dialog.


nicost:
mm.settings().setUnlockAutofocusInZStack(false);

nicost:
Closes #1937

PR #1945, opened 2024-05-13
Upgrade the boost and GNU libtool for Mac build

marktsuchida:
Because Boost 1.77.0 does not appear to work with the latest Apple clang++.

GNU Libtool libraries need to match the command-line libtool, so use the version that matches the current one from Homebrew. (Jenkins build machine will receive a matching upgrade.)


marktsuchida:
(GNU libtool version needing to match command-line tool might apply only when building x86-64 on arm64).

PR #1950, opened 2024-05-20
Adds ability to select the most in focus Z stack in multiple places.

### LINKED ISSUE #1905: Request for new autofocus plugin
While discussing code needed for lightsheet microscopes, we realized there is a need for a new aufocus plugin in Micro-Manager that has some of the capabilities of the diSPIM plugin in 1.4.  

Desired features are:
- It should be a superset of the Oughtafocus plugin, in addition it should:
- allow setting the z-stage to use for focusing (which can be different from the core-focus drive)
- Optionally run a Z-stack and interpret this after the fact rather than run a Brant optimizer.  In this mode, it should run a sequence if the hardware allows it (this could most easily be done using the acquisition engine), and display a focus quality curve if requested

Optional: the code could be structured such that it could accept a DataStore and provide a focus score per image in the store.  

Also see the Image Sharpness pane in the Inspector window.

Name?

nicost:
- OughtaFocus plugin: adds option to run a complete Z stack and display the focus curve. Also adds the option to select a ZStage other than the default ZStage.
- FrameCombiner:  adds option to pick the sharpest frame (rather than combine features from multiple frames).
- ZProjector: adds option to select the sharpest frame.

Fixed bugs/confusion in Duplicator plugin and ImageProcessing library.

Closes #1905


marktsuchida:
@nicost Unfortunately this doesn't compile with the official build system:
```text
 [mm-javac] /Users/mark/devel/micro-manager/plugins/FrameCombiner/src/main/java/org/micromanager/plugins/framecombiner/FrameCombinerConfigurator.java:21: error: package org.micromanager.multichannelshading does not exist
 [mm-javac] import org.micromanager.multichannelshading.MultiChannelShading;
 [mm-javac]                                            ^
```

Is this perhaps a typo? I'm not sure why FrameCombiner would want to use `MultiChannelShading.MENUNAME`.

marktsuchida:
No problem, will do.

PR #1952, opened 2024-05-21
Studio: Make CoreEventCallback aware of the current acquisition engine.

### LINKED ISSUE #1948: Hardware communication differences with new acquisition engine?
Hi all,

We have a microscope that uses a ASI LX-4000 to control an XY stage, Z stage, and FW-1000 filter wheel. When we setup an MDA with multiple z positions and multiple channels, we encounter different behavior between the original acquisition engine and new experimental engine. It doesn't matter what order we acquire in, e.g. `[slice, channel]` or `[channel, slice]`

With the original acquisition, the acquisition completes fine. We can repeat it N times, do XY positions, etc.. and it all works well.

With the new experimental engine, we immediately get a serial error that the status of the one components connected to the LX-4000 cannot be read. For example, the position of the z stage cannot be read. Interestingly, the acquisition continues and then completes. The error also occurs if we run the same acquisition in pycromanager, which I believe uses the new experimental engine.

What can we do to help troubleshoot this? One guess I have is that the new engine is either sending too many commands to the LX-4000 or not waiting appropriately for a response before sending another command. 

Thanks!

nicost:
Should fix issue #1948.

CoreEventCallBack was never told about the current acquisition engine.  My only worry is that the changed order of initialization in MMStudio may cause unforeseen problems.   Hope that Mark can review.

Closes #1948.

### DIFF
--- a/mmstudio/src/main/java/org/micromanager/events/internal/CoreEventCallback.java
+++ b/mmstudio/src/main/java/org/micromanager/events/internal/CoreEventCallback.java
@@ -26,7 +26,7 @@ import javax.swing.SwingUtilities;
 import mmcorej.CMMCore;
 import mmcorej.MMEventCallback;
 import org.micromanager.Studio;
-import org.micromanager.acquisition.internal.AcquisitionWrapperEngine;
+import org.micromanager.acquisition.AcquisitionManager;
 
 /**
  * Callback to update Java layer when a change happens in the MMCore. This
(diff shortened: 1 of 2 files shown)


marktsuchida:
I guess the question is whether it is safe to not have the callback set yet when the hardware config is loaded.

There are calls to `coreCallback_.setIgnoring()` within `loadSystemConfiguration()`, which needs checking.

Also, does the "config changed" event generated during this initial config loading need to be handled?

nicost:
To be completely sure, I changed the order of callback creation and loading config file back to what it was before.  This works with the demo config, and I can not imaging there to be any change in behavior with the previous versions.

PR #1967, opened 2024-06-07
Include non MDA acquisitions in checks for active Acqs

henrypinkard:


henrypinkard:
Will need to pull corresponding suto version update from AcqEngJ

PR #1975, opened 2024-06-16
Studio: now actually saves MDA Settings in file.

nicost:
I wonder about the representation.  It would be nicer to store them as a propertymap rather than a string, will need to look into that before merging. Since these data have not been stored yet (will need to check format of MDA Settings stored from MDA window), that may still be possible without backward compatibility issues.

### DIFF
--- a/mmstudio/src/main/java/org/micromanager/data/internal/PropertyKey.java
+++ b/mmstudio/src/main/java/org/micromanager/data/internal/PropertyKey.java
@@ -709,7 +709,7 @@ public enum PropertyKey {
       }
    },
 
-   MDA_SETTINGS("MdaSettings", SequenceSettings.class){
+   MDA_SETTINGS("MdaSettings", SequenceSettings.class) {
       @Override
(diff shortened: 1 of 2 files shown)


marktsuchida:
@nicost You might be thinking this already, but I would say that it would be good to have one of two things:

1. A best-effort copy of the MDA settings as JSON-in-JSON (hard for humans to read but easy for scripts to extract), as is, with no formal promise about the content. (I.e., what you currently have.)

2. MDA settings stored under a well-defined (if not formal) schema, with documented (or documentable) and stable field keys.

I'm not strongly for or against either, but probably would rather not have something in between. Option 1 keeps a wall between the rigid metadata fields and the less-formally-defined ones. Option 2 eliminates less-formally-defined fields. Either way, we avoid losing the well defined schema of the metadata.

Option 2, despite its benefits, is a lot of work and also requires more future work when the schema needs to be extended. It might be that Option 1 gives 90% of the benefit with 5% of the work.....

nicost:
Alright, let's stick to option #1 (for now).  Thanks!

PR #1993, opened 2024-07-11
AcqEngJ adapter: ZStage now returns to start position at end of acquisitions.

nicost:


henrypinkard:
I could see this being a confusing change for people using pycro-manager that all of the sudden the Z drive is making an extra movement. You might consider making this an opt-in behavior used by the studio code only

nicost:
This is replicating behavior of the Clojure engine.  It was always the goal
to replicate the existing behavior as much as possible.  If you would like
different behavior for people accessing Studio through Pycromanager, then
forking the complete projects may be the way to go ( I would not even know
how to start determining who is calling the code, plus that will add
significant maintenance burden)..

By the way, it is very weird to find the z stage at a different position at
the end of an acquisition than at the beginning, I can't really imagine
anyone liking that behavrio.

henrypinkard:
Wait actually ignore my previous comment -- I was confused and thought this was a change to AcqEngJ, not its adapter. This is the correct way to do this.

PR #2013, opened 2024-10-12
Reader recycled buffers

nicost:
Copied the mechanism used in MultipageTIffWriter to recycle the Direct Buffers for reading.  I am not 100% sure this is correct, especially, when the Direct Buffer is handed off to "Image" as Buffer.toArray(), does that cause a copy into the JVM managed memory?  If not, it seems possible that the memory is re-used for things it should not be re-used for.  Testing looks OK though.


mmstudio/src/main/java/org/micromanager/data/internal/multipagetiff/MultipageTiffReader.java
```diff
@@ -92,6 +93,83 @@ public final class MultipageTiffReader {
...
+
+   private static int pooledBufferCapacity_ = 0;
+   private static ByteOrder pooledBufferByteOrder_ = null;
+
+   private static ByteBuffer getLargeBuffer(int capacity, ByteOrder byteOrder) {
+      if (BUFFER_POOL_SIZE == 0) {
+         return allocateByteBuffer(capacity, byteOrder);
+      }
+
+      synchronized (MultipageTiffWriter.class) {
```
marktsuchida: Wrong class, should be Reader

mmstudio/src/main/java/org/micromanager/data/internal/multipagetiff/MultipageTiffReader.java
```diff
@@ -92,6 +93,83 @@ public final class MultipageTiffReader {
...
+      return allocateByteBuffer(capacity, byteOrder);
+   }
+
+
+   private static void tryRecycleLargeBuffer(ByteBuffer b) {
+      // Keep up to BUFFER_POOL_SIZE direct buffers of the current size
+      if (BUFFER_POOL_SIZE == 0 || !b.isDirect()) {
+         return;
+      }
+      synchronized (MultipageTiffWriter.class) {
```
marktsuchida: Also should be Reader

mmstudio/src/main/java/org/micromanager/data/internal/multipagetiff/MultipageTiffReader.java
```diff
@@ -504,16 +582,22 @@ private Image readImage(IFDData data) throws IOException {
...
          // mapped).
          switch (pixelType) {
-            case GRAY8:
-               return new DefaultImage(pixelBuffer.array(), formatPmap,
+            case GRAY8: {
+               Image img = new DefaultImage(pixelBuffer.array(), formatPmap,
                      coords, metadata);
-            case GRAY16:
+               tryRecycleLargeBuffer(pixelBuffer);
+               return img;
```
marktsuchida: These can be factored out after the `switch` statement to avoid repetition (for each `case`).

mmstudio/src/main/java/org/micromanager/data/internal/multipagetiff/MultipageTiffReader.java
```diff
@@ -504,16 +582,22 @@ private Image readImage(IFDData data) throws IOException {
          // a nio buffer directly as the Image storage (even better if memory
          // mapped).
          switch (pixelType) {
-            case GRAY8:
-               return new DefaultImage(pixelBuffer.array(), formatPmap,
+            case GRAY8: {
+               Image img = new DefaultImage(pixelBuffer.array(), formatPmap,
                      coords, metadata);
```
marktsuchida: `ByteBuffer.array()` only works if the byte buffer is backed by a Java array (i.e., it's not a direct byte buffer) and returns the backing array.

And indeed the `DefaultImage` constructor (re-)wraps the array as a buffer, which would mean that immediately recycling the buffer is incorrect.

marktsuchida: I'm a little confused here -- if the previous implementation was reusing a single buffer (per reader), was that also overwriting the contents of `DefaultImage`s that are already passed out???

nicost: Similarly confused.  This situation only applied to 8 bit images, which probably were not used much.  To be safe, I now clone the ByteBuffer (which - as far as I understand - should make a copy of the values in the array).

nicost: Turns out that PixelBuffer is not backed by an array, so that code failed.  I added code to check if the array backing exists and make a copy if it does not.  That makes the 8 bit situation more similar to the 16 bit case.  Seems to work now.

nicost: Also remove functions/class added in #1934, that are indeed no longer needed.

PR #2020, opened 2024-10-29
No reload when saving the configuration file

nicost:
So far, the Micro-Manager UI has always unloaded the existing configuration and reloaded it from file when saving a configuration.  Depending on the hardware, this can be a time-consuming action.  Thinking about speeding up some of the UI components, I realized there is no need for this except possibly to check quickly whether the new configuration file is faulty (please correct me if I am wrong in this assumption).  Over the many years using this, I have never found a problem.  

This PR keeps the existing configuration in memory after saving a new config file.  In limited testing with hardware, I have not found a problem, and it does speed things up quite a bit.

### DIFF
--- a/mmstudio/src/main/java/org/micromanager/internal/ConfigGroupPad.java
+++ b/mmstudio/src/main/java/org/micromanager/internal/ConfigGroupPad.java
@@ -111,6 +111,9 @@ public final class ConfigGroupPad extends JScrollPane {
       if (data_ != null) {
          data_.rebuildModel(fromCache);
          data_.fireTableStructureChanged();
+         if (table_.getCellEditor() != null) {
+            table_.getCellEditor().stopCellEditing();
+         }
          table_.repaint();
       }
    }
--- a/mmstudio/src/main/java/org/micromanager/internal/DefaultApplication.java
+++ b/mmstudio/src/main/java/org/micromanager/internal/DefaultApplication.java
@@ -183,7 +183,9 @@ public class DefaultApplication implements Application {
          model.createSetupConfigsFromHardware(studio_.core());
          model.createResolutionsFromHardware(studio_.core());
          model.saveToFile(path);
-         ((MMStudio) studio_).setSysConfigFile(path);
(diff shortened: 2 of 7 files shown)


marktsuchida (commented):
> Over the many years using this, I have never found a problem.

I have. There have been bugs in device adapters that caused errors or crashes at this stage, due to subtle issues that allowed devices to initialize one by one but not in succession.

I can't remember which devices (maybe one of the NI ones, and one of the prominent cameras a long time ago?). I also vaguely remember that one of the camera company people asked me about a crash or error happening during the reload -- and we were able to find the bug (in the device adapter) causing it.

There was also an issue with certain serial ports, though that only manifested upon rapidly unloading and reloading so would not be a problem here.

Another issue is that devices can end up in a different state between the interactive configuration and the final config file. For example, you can add some devices, then remove one or more devices that are not the last ones in the list. It is (theoretically) possible that those removed devices had some effect on the loading of the later devices.

On the other hand, all of these issues are about buggy device adapters, and in that sense this change is a perfectly correct one. I guess I'm just not very confident that we don't have (or won't have in the future) more such bugs -- and removing this check could make such bugs harder to find early and diagnose.

How do you feel about having a check box (my preference would be to have it checked by default) like "Reload configuration to check"? I feel that this check is somewhat valuable at least in some scenarios, given that configuring the device is one of the primary ways that device adapter writers interactively test their code.

nicost:
Great idea!  Only problem I see is this will need another dialog (I don't think that I can add this to the file-chooser when saving the configuration file), or go into Options (which will be hard to find).

Just to be clear, the Hardware Configuration Wizard is still reloading the config file at the end (which is where I believe the problems you describe will happen).  This PR only touches the code invoked when saving the configuration file after making changes to the pixel size configuration or the configuration groups.  Have you ever seen bugs at that point?

marktsuchida:
My bad, somehow misread this as bing about the HCW. For the regular saving after changing config groups, pixel size, etc., it should be totally fine, and I don't think we need the option to check.

PR #2021, opened 2024-10-30
Studio: adds optional background frame

nicost:
Adds optional background frame that makes the Micro-Manager application look a lot better on crowded desktops.  Behavior is still a bit crummy, interfacing with other applications is a bit weird, but overall a better experience for users.  Needs testing on Mac and linux.


marktsuchida:
I'll be happy to test this on macOS and Linux when I get a chance.

nicost:
image

nicost:
Current code only cover one monitor in dual monitor setup.  Only way I see to make this work in multiple monitor configs is to make multiple background frames (one for each monitor) and synchronize their behavior.  Should be doable.  However, I believe the need for this is biggest for single monitor configurations, so we possibly can merge this first and deal with multiple monitors later.

nicost:
Merging as it is an optional feature, off by default.  Happy to try to address issues showing up in its use.

PR #2033, opened 2024-12-10
Light sheet control, plugin to set up the Multi Stage adpater nicely

nicost:
Original by Kurt Thorn, reworked by Nico Stuurman.  Let's the user set positions where the two stages driven by the 
Multi-stage adapter are in focus.  Calculates a linear fit between these reference point to set offset and slope for the second stage.  The plugin now displays the positions of the two stages and the linear fit in a plot for more immediate feedback.  Also adds saving the reference points to file (and restoring from file).


jondaniels:
How about renaming this plugin for clarity?  I think it should really be something like "Multi-Stage Control" (reflecting the name of the device that it expects) or "Paired Stage Control" (perhaps better reflecting the use case).  From what I can tell this functionality is used in some light sheet microscopes but really the plugin and logic is completely separate from what makes a light sheet a light sheet.

Separately I tried out the plugin to get a flavor of it and I just get an empty box when I launch it.  My test config file is a DemoCam device with a DStage and DXYStage plus a SingleAxisStage to get a 2nd 1D stage.

nicost:
Great suggestion!  I'll commit that change soon.  This is an old plugin that Kurt Thorn wrote a long time ago and only used at one system at UCSF.  It makes my life a lot easier to have it in the micro-manager repo, and who knows, it may be useful to others.  

When I open it with just the demo config, I get a dialog saying:` "Cannot find multi stage device. This plugin does not work without one. Use the Hardware Configuration wizard and select Utilities > MultiStage"`.  Did you not get that?  I do now indeed see the empty frame afterwards, and changed the code to avoid showing that.  

Thanks for the feedback Jon!

jondaniels:
I did create a separate config with demo multi-stage device but I must have been doing something wrong previously.  Seems to work now.  Thank you!

PR #2045, opened 2024-12-21
MDA window Z slices section feature additions

nicost:
Adds:
- selector for the current Z drive, including indicator of current position (will only work with callbacks)
- GoTo button for top and bottom, only active in absolute mode
- Proposed Z-step value that can be applied with a single button click. The proposed value is currently 5 x the pixel size, but after adding this feature to the pixel size calibration it can be read from that source.

image


marktsuchida (approved):
LGTM!

PR #2059, opened 2025-02-05
Pixel config: UI for new core pixel Size config functions

nicost:
Adds dxdz, dydz, and OptimalZUm to the pixel Size configuration dialog, and saves these settings in the config file.  Should only be merged after the corresponding PR for mmCoreAndDevices has been merged.

I hope the UI is still workable for people:

image


nicost:
Great suggestions and good catch!

PR #2060, opened 2025-02-06
Studio: Fixes the right alignment of the entries on the splash screen…

nicost:
….  Previous attempt did not work very well, this looks better.


marktsuchida:
@nicost Thanks! I was about to bring this up.

Issue #2075, opened 2025-03-06
pgFocus plugin fails to load in recent nightly builds


kmdouglass:
I am resurrecting an old custom-built Microscope setup. In the process, I have upgraded Micro-Manager to a recent nightly version and everything works except for the pgFocus plugin.

I realize that I might be the only person who uses this, so I am investigating myself. I am opening this issue to track the progress of my investigation.

## Steps to reproduce

1. Download and install a recent nightly version. I am using 20250228.
2. Create a new hardware configuration and add the pgFocus hub adapter and Stabilization component.
3. Navigate to `Plugins` > `Device Control` > `pgFocus`
4. Observe the following error:

> Error: pgFocus::getStringValue() Error reading values from pgFocus: Calibration curve

![Image](https://github.com/user-attachments/assets/5e47813c-0484-4b3d-a68c-dced319674b7)

## Additional details

- The plugin works fine on the old version of Micro-Manager that the setup last used: 20180108
- The commit history for [PGFocusFrame.java](https://github.com/micro-manager/micro-manager/commits/main/plugins/pgFocus/src/main/java/edu/umassmed/pgfocus/PGFocusFrame.java) shows only one change between 20180108 and now: https://github.com/micro-manager/micro-manager/commit/b9ef0613c6bd85212541bc8bed9104c5baed1ca7

@kbellve are you still active on this project?

kmdouglass:
Looking at the code, it seems like the error is triggered here in a method called `updateValues`: https://github.com/micro-manager/micro-manager/blob/main/plugins/pgFocus/src/main/java/edu/umassmed/pgfocus/PGFocusFrame.java#L418

The line at the above link calls a helper function called `getStringProperty`:

```java
   String getStringProperty(String property) {
      try {

         String val;

         val = core_.getProperty(pgFocus_, property);

         return (val);

      } catch (Exception ex) {
         ReportingUtils.showError(
               "pgFocus::getStringValue() Error reading values from pgFocus: " + property);
      }

      return (null);
   }
```

Likely the `getProperty` of the core method is failing. Unfortunately the exception information from the core is dropped in the error handling.

kmdouglass:
I recompiled the plugin to log the exception that's being hidden:

> 2025-03-06T17:07:07.544392 tid1916 [IFO,App] java.lang.Exception: Cannot get value of property "Calibration curve" [ Error in device "pgFocus-Stabilization": Invalid property name encountered: Calibration curve (2) ]

[...]

PR #2076, opened 2025-03-07
Fix calibration property name in the pgFocus plugin

kmdouglass:
This fixes the error described in #2075 in the pgFocus plugin. The plugin uses a property name for the `Calibration Curve` property that differs from the one defined in the device adapter in the case of only a single letter.

https://github.com/micro-manager/mmCoreAndDevices/blob/a9b83246ca9154c4e20b125df72a40027de2d8c8/DeviceAdapters/pgFocus/pgFocus.h#L117

I am guessing that this previously worked because getting properties from the core by name used to be case insensitive; this likely changed sometime after 2018.


kmdouglass:
No worries at all, @kbellve . It didn't take much time at all to debug and fix it :)

PR #2078, opened 2025-03-11
[mm1.4] ASIdiSPIM fix for hardware timepoints with interleaved channels

jondaniels:
make sure we are dropping the right images


jondaniels:
also added API changes, accidentally on same local branch (sorry)

PR #2094, opened 2025-04-23
Fluid control

Lars-Kool:
I've made a small GUI that automatically detects all pressure controllers and volumetric controllers (screenshot is for a syringe pump). It allows the user to either enter a value in the textbox, or use the slider (values are in uL/s for volumetric pumps and kPa for pressure pumps).

Screenshot 2025-04-23 101626

### DIFF
--- a/plugins/FluidControl/.gitignore
+++ b/plugins/FluidControl/.gitignore
@@ -1,0 +1,29 @@
+### IntelliJ IDEA ###
+out/
+!**/src/main/**/out/
+!**/src/test/**/out/
+
+### Eclipse ###
+.apt_generated
+.classpath
+.factorypath
+.project
+.settings
+.springBeans
+.sts4-cache
+bin/
+!**/src/main/**/bin/
+!**/src/test/**/bin/
+
+### NetBeans ###
+/nbproject/private/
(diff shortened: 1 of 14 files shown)


marktsuchida:
@Lars-Kool Thanks for the updates. (And sorry for the strict formatting checks without automatic fixing -- the plan is to improve this (#1583) but I haven't gotten around to it yet.)

Lars-Kool:
@marktsuchida No problem. Is there any way to run these checks locally (preventing unnecessary checks by Github Actions). Also, I'll create a PR soon to update the wiki to include instructions on how to add the correct codestyle (at least for IntelliJ, I have no experience with other IDE's).

nicost:
Sorry this took so long.  I went through the CheckStyle stuff, but then also wanted to test the plugin.  I ended up writing demo pump devices;).  That showed there are some specific properties needed.  I therefore added messages to the plugin when those properties are missing.  I also added event listeners that will redraw the UI when the configuration changes, and that changes the position of the slider when the underlying property value changes. 
Can you test this with your devices?  I may have misunderstood some things, so please correct if it does not work as expected.

PR #2099, opened 2025-05-06
Image Flipper: Fix bug when rotating non-square images by 90 or 270 degrees.

nicost:


marktsuchida:
I force-pushed to remove redundant commits. Will review code tomorrow.

PR #2101, opened 2025-05-07
Upgrade JRE shipped with Windows installer

marktsuchida:
Java 8 152 -> ~452+9~ 11.0.27+6.

This is a prerequisite for https://github.com/micro-manager/mmCoreAndDevices/pull/626.

@nicost Do you remember any particular things that have broken in the past for particular JRE versions that I should test before merging?


marktsuchida:
Seems to cause the ImageJ toolbar to get cropped.
image

marktsuchida:
@nicost Yeah, Temurin JRE 11.0.27+6 fixes the toolbar problem, too. If you aren't aware of any issues with 11, I think that may be the way to go.

nicost:
Worth testing a bit, but likely a good way to go. Update ImageJ at the same time?

nicost:
I ran 11.0.27+6 both direct and under IntelliJ. Under IntelliJ the display looks a lot better.  The application feels snappier.  Slight glitch was that the ClearVolume window was a bit strange, but still usabel for 3D display.  Magellan worked.  I will try to test at an actual microscope as well.

marktsuchida:
Thanks, @nicost. I changed this PR to switch to 11.0.27+6, but let's give it a few days before merging.

I think we can upgrade ImageJ on a different day (before or after), just in the interest of not changing many things at the same time (same nightly build) -- unless we need a newer ImageJ in order to work with JRE 11.

nicost:
Two problems with Java 11:
- I am unable to edit the exposure time box in the main window
- I can not edit the numeric value in sliders in the config groups in the main window
Image display seemed a bit less fluent with Java 11 than Java 8, but I am not absolutely sure of this difference.

marktsuchida:
Yep, seeing the same here. I also cannot edit non-slider (numeric) values in the main window config group table. I wonder if it is something about the main window rather than about the individual widgets.

Also noticed that the exposure cannot be changed in the Device Property Browser, either; it reverts back to what's stuck in the main window.

The Live frame rate does seem a bit slower to me, too, though I'm not yet sure if it's enough to be a concern.

nicost:
Found the fix.  The problem is in a function that was a workaround for misbehavior on the Mac.  Changing MainFrane.java  at line 195 to:

```
 public void windowActivated(WindowEvent event) {
            if (JavaUtils.isMac()) {
               // This is a workaround for a bug in Java on OSX that causes
               // the menu bar to be disabled.
               setJMenuBar(null);
               setJMenuBar(getJMenuBar());
            }
}
```

You may want to test on Mac OS if this function can be deleted altogether.

marktsuchida:
Wonderful, will test on macOS!

marktsuchida:
Prerequisite: #2103

PR #2105, opened 2025-05-12
Update StartMMStudio.m

KonstantinNeuhaus:
While the general change works, I made a small error in the first occurrence of checking the Matlab Version.


marktsuchida:
Thanks!

PR #2111, opened 2025-05-16
Remove DPI awareness settings for Windows

marktsuchida:
These should not be needed with JDK >= 9, and we are now shipping 11.


marktsuchida:
Just to record: with JRE 8, the system property (`sun.java2d.dpiaware`) did not seem to do anything but the registry entry did work, so testing a staged build without running the installer would cause MM to show up tiny on hi-dpi monitors. With JRE 11, everything looks fine even with a staged build, regardless of whether `sun.java2d.dpiaware` is set.

Issue #2114, opened 2025-05-19
macOS code signing


marktsuchida:
I'm creating a new issue for the macOS part of #396, since this is going to have at least a couple of sub-issues.
See also #2001, although this is independent of full arm64 support.

Removing the macOS Gatekeeper issues requires:
- Signing all binaries (machine code files: in our case, dylibs and Mach-O bundles)
- Signing the `ImageJ.app` launcher that we use (for now)
- Signing the disk image (DMG) -- while not required, Apple says this helps with avoiding app translocation (path randomization)
- Notarizing the DMG
- Stapling the notarization ticket to the DMG -- while not required, this allow the app to run without contacting Apple

Each of the above steps is just running a command. In addtion, we need to set up the necessary credentials in our CI.

PR #2115, opened 2025-05-19
macOS code signing and notarization

marktsuchida:
For #2114.

So far, this adds scripts to perform the signing, notarization, and stapling as part of the official build. They assume that credentials are set up externally by the CI environment (which should be fine).

Notarization is not working yet because of dylibs inside JARs that are not signed (#2116).


marktsuchida:
This is now fully working on my local computer if #2119 is also applied. Can be merged once #2119 is merged, because signing and notarization will not yet happen on CI until enabled with flags to the build script.

PR #2133, opened 2025-05-30
Build the PluginUtilities library

marktsuchida:
Add the `PluginUtilities` library (#2132) to the build.


nicost:
Thank you Mark!

PR #2138, opened 2025-06-10
 Add Color Channel Support (BGRA) to AutofocusTB for Color Cameras (e.g. Basler)

lokeshvelayudham:
###  Summary

This PR enhances the `AutofocusTB` plugin to support color image inputs (specifically BGRA format) from color cameras such as Basler. It allows users to select a specific channel (Red, Green, or Blue) to perform autofocus, while preserving full compatibility with grayscale images.

---

###  Problem

The original implementation assumes grayscale-only images (either 8-bit or 16-bit), and fails when color images are returned from the camera (e.g., 8-bit BGRA from Basler). This causes:
- Blank or incorrectly rendered images
- Autofocus scoring to malfunction
- Incompatibility with modern color cameras

---

####  1. Added a new plugin setting: `Color Channel`
- Options: `"Grayscale"`, `"Red (BGRA)"`, `"Green (BGRA)"`, `"Blue (BGRA)"`
- Added to the plugin properties and integrated with `applySettings()` and `getProperties()`

####  2. Enhanced `snapSingleImage()` logic:
- Detects number of image components
- If `>= 3` and input is a `byte[]`, it extracts the selected channel from BGRA data
  - Red: `bytes[4 * i + 2]`
  - Green: `bytes[4 * i + 1]`
  - Blue: `bytes[4 * i + 0]`
- Creates a `ByteProcessor` for the selected channel
- Displays the image and stores it in `ipCurrent_` for scoring

####  3. Preserved existing grayscale handling
- Original logic for 8-bit and 16-bit grayscale remains unchanged
- Only color extraction is added as an enhancement

#### 🪵 4. Improved logging
- Image dimensions, number of components, and selected channel are logged for easier debugging

---

###  Before vs After

| Image Format      | Original Behavior     | New Behavior                     |
|-------------------|------------------------|----------------------------------|
| 8-bit Grayscale   | ✅ Works                | ✅ Still works                    |
| 16-bit Grayscale  | ✅ Works                | ✅ Still works                    |
| 8-bit BGRA (Basler)| ❌ Fails or shows blank | ✅ Works with selectable channel  |
| Unsupported       | ❌ May crash            | ✅ Logs error and skips gracefully|

---

###  Motivation

- Enables autofocus for color cameras in Micro-Manager
- Useful in cases where a specific channel (e.g. Red fluorescence) provides better contrast
- Keeps UI simple and minimally invasive

---

###  Tested

- ✅ Verified autofocus works on grayscale cameras
- ✅ Tested with a Basler USB3 color camera (BGRA format)
- ✅ Confirmed correct single-channel extraction
- ✅ Ensured `ImagePlus.show()` renders as expected
- ✅ Maintains existing behavior when using grayscale input

---

###  Contributor Note

This was discovered while integrating a Basler color camera with Micro-Manager at BioInVision Inc http://www.bioinvision.com/.  Since color input was not supported in the original plugin, I’ve added this logic to help improve compatibility for other users with similar setups. This feature is backward-compatible and opt-in via the new "Color Channel" setting.

**Contributor:**
 *Madhusudhana Gargesha , BioInVision - https://www.linkedin.com/in/madhusudhana-gargesha-b8823317*
 *Lokesh Poluru Velayudham, BioInVision - https://www.linkedin.com/in/lokeshvelayudham

Let me know if you'd prefer this as a separate plugin or if additional configuration is preferred.

### DIFF
--- a/autofocus/src/main/java/org/micromanager/autofocus/AutofocusTB.java
+++ b/autofocus/src/main/java/org/micromanager/autofocus/AutofocusTB.java
@@ -6,6 +6,7 @@ import ij.gui.ImageWindow;
 import ij.process.ByteProcessor;
 import ij.process.ImageProcessor;
 import ij.process.ShortProcessor;
+import ij.process.ColorProcessor;
 import java.awt.Color;
 import java.awt.Rectangle;
 import java.util.Date;
@@ -45,7 +46,7 @@ public class AutofocusTB extends AutofocusBase implements AutofocusPlugin, SciJa
    private ImageProcessor ipCurrent_ = null;
 
    /**
-    *  Description of the Field.
+    * Description of the Field.
     */
    public double sizeFirst_ = 2;
    //
@@ -59,11 +60,10 @@ public class AutofocusTB extends AutofocusBase implements AutofocusPlugin, SciJa
    public String channel2_ = "DAPI";
 
    private final double indx = 0;
-   //snapshot show new window iff indx = 1
+   // snapshot show new window iff indx = 1
 
    private boolean verbose_ = true;
    // displaying debug info or not
-
 
    private String channelGroup_;
 
@@ -76,6 +76,9 @@ public class AutofocusTB extends AutofocusBase implements AutofocusPlugin, SciJa
    private long tPrev;
    private long tcur;
 
+   // channel group for autofocus
+   private static final String KEY_COLOR_CHANNEL = "Color Channel";
+   private String selectedColorChannel_ = "Red (BGRA)"; // Default
 
    /**
     * Constructor for the Autofocus_ object.
@@ -89,8 +92,8 @@ public class AutofocusTB extends AutofocusBase implements AutofocusPlugin, SciJa
       super.createProperty(KEY_CROP_SIZE, Double.toString(cropSize_));
       super.createProperty(KEY_CHANNEL1, channel1_);
       super.createProperty(KEY_CHANNEL2, channel2_);
+      super.createProperty(KEY_COLOR_CHANNEL, selectedColorChannel_);
    }
-
 
    @Override
    public void applySettings() {
@@ -104,6 +107,8 @@ public class AutofocusTB extends AutofocusBase implements AutofocusPlugin, SciJa
          channel1_ = getPropertyValue(KEY_CHANNEL1);
          channel2_ = getPropertyValue(KEY_CHANNEL2);
 
+         selectedColorChannel_ = getPropertyValue(KEY_COLOR_CHANNEL);
+
       } catch (Exception e) {
          // TODO Auto-generated catch block
          e.printStackTrace();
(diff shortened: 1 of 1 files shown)


autofocus/src/main/java/org/micromanager/autofocus/AutofocusTB.java
```diff
@@ -246,37 +249,95 @@ public void run(String arg) {
...
    }
 
-
-
-   //take a snapshot and save pixel values in ipCurrent_
+   // take a snapshot and save pixel values in ipCurrent_
    private boolean snapSingleImage() {
       try {
          core_.snapImage();
+         Thread.sleep(50);
```
marktsuchida, line 265: Did you find this sleep necessary? In theory `snapImage` should wait for the exposure to finish, and `getImage` should wait for the last-snapped image. If that's not working, we perhaps should check for bugs in the camera device adapter.

autofocus/src/main/java/org/micromanager/autofocus/AutofocusTB.java
```diff
@@ -246,37 +249,95 @@ public void run(String arg) {
...
          Object img = core_.getImage();
-         ImagePlus implus = newWindow();
-         // this step will create a new window iff indx = 1
-         implus.getProcessor().setPixels(img);
-         ipCurrent_ = implus.getProcessor();
+
+         int width = (int) core_.getImageWidth();
+         int height = (int) core_.getImageHeight();
+
+         ImagePlus implus;
```
marktsuchida: It looks like you replaced the call to `newWindow()` with the code below, but now `newWindow()` (which you also updated) is never called. Could you maybe put the code below into `newWindow()` and go back to calling it?

marktsuchida (changes requested):
Thank you, @lokeshvelayudham and Madhusudhana. This is much needed functionality.

I just had one question and one suggestion.

autofocus/src/main/java/org/micromanager/autofocus/AutofocusTB.java
lokeshvelayudham: I initially added the Thread.sleep(50) after noticing occasional issues where getImage() would return incomplete  data immediately after snapImage() when using a Basler color camera via the Pylon device adapter. It was inconsistent but reproducible enough during my autofocus testing to warrant a delay.

That said, I agree this shouldn’t be necessary if the adapter is compliant. I’m happy to remove the sleep .

autofocus/src/main/java/org/micromanager/autofocus/AutofocusTB.java
lokeshvelayudham: You’re right — thanks for catching that! I’ve moved the image initialization logic back into newWindow() and updated snapSingleImage() to call it again.

autofocus/src/main/java/org/micromanager/autofocus/AutofocusTB.java
marktsuchida: Looking at the Basler device adapter code, it looks like the image is copied to an internal buffer during the call to `snapImage()` (`SnapImage()` in the device adapter), and `getImage()` (`GetImageBuffer()` in the device) merely returns that buffer. So the behavior you see is surprising, though I am unable to test this because I don't have a Basler camera.

It would be great if you could see if there is a robust way to reproduce the issue, but for the purpose of this autofocus code, I think it will be good to remove the sleep.

PR #2143, opened 2025-06-18
Deskew plugins manages test acquisitions like the MDA window does

nicost:


nicost:
Tried to explain this better.  The concept of a Test Acquisition is very much growing on me, especially for the Light Sheet, but also in other modalities (like testing Z Stack settings).  Having this concept here makes it easier for the UI to be accommodating user's workflows.

Issue #1791, opened 2023-09-13
New Micro-manager versioning scheme


nicost:
So far, we have equated the version of MMStudio with the version of the Micro-Manager release.  @marktsuchida proposed to uncouple the two and switch to a simplified date for the release, such as 2023-1, possibly coupled with a yearly or twice a year release cycle.  This would make it possible to change the major version of MMStudio according to semver, whenever we feel it is needed without building an official release.  

I like this idea very much, and believe we can switch without too much effort.  Couple of obvious questions:
- Do we want to stick to a once or twice a year release schedule?  Should there be a fixed date, by which we freeze new features, test for a few weeks, then release?
- Should the name be 2023-1, 2023-2, or 2023A, 2023B, or something completely different?

marktsuchida:
Discussed in 2023-10-23 core dev meeting and approved. Details to be worked out.

marktsuchida:
Turns out I had already written this up as #1301. [CalVer](https://calver.org/) provides some case studies on the naming scheme. I think I like 2023.11, 2024.5 (using the month), but it might be good to have a plan for when we need two in the same month -- maybe 2023.11.0, 2023.11.1, with anything other than .0 being rare.

I'd say we don't need to commit to a particular release interval (always can later if we think it's a good idea). Also probably easier to avoid feature freezes and just release a version that we deem stable (which need not be the tip of main at the time of release). Alternatively, we could simply promote a chosen nightly build to "release" status by overwriting its version info.

Rough tasks needed once we decide on naming specifics:
- Change MMStudio to use its own version number rather than getting it from version.txt (plus build scripts) during the build (eventually this will be in a pom.xml, but for now it can be in MMVersion.java)
- Add a package-wide version file (probably JSON), generated during the build and installed by the installer
- Make MMStudio display the package-wide version info (when available) where it shows the version (title bar; About dialog)
- Make sure MMStudio APIs that return the version do the right thing (probably should check where it is used currently)

marktsuchida:
Also note: the 2.x version number can then be repurposed for the MMStudio API version (SemVer).

PR #2144, opened 2025-06-18
Retire version.txt and simplify version setting

marktsuchida:
Part of #1791/#2140 (without the new naming patterns).

In the direction of switching to using CalVer for Micro-Manager versioning, remove the 2.0.x version number everywhere (we may later resurrect the 2.x version numbering for MMStudio API versions).

For builds using the "nightly" scripts but not explicitly setting the version, the version now defaults to `adhoc` plus the local date (appropriate for local test builds). Previously the date in US Pacific Standard Time (appropriate for nightly builds) was used in this case; it is now the invoker's responsibility to set such a date (with or without other version information).

We also remove the distinction, in the scripts, between nightly and release builds; for all official automated builds, the invoker must set the whole version string (including the date, if applicable).

For plain builds not using the "nightly" scripts, no version is set, the `MMVersion` class is not overwritten, and (on Windows) the installer name does not contain a version (all as before). In this case, `MMVersion` now defaults to `adhoc-build` rather than `2.0.x`.


marktsuchida:
- [ ] `buildscripts/installer_Windows.iss` contains `2.0` in several places. Decide what to do with those.
    - We could inject the actual version number into the installer script, but that may cause each version to be considered a different application (though maybe the AppID prevents that) or be installed (by default) into different directories. Maybe remove the version number in some places.

Issue #2113, opened 2025-05-17
Upgrade to JDK 11


marktsuchida:
Our Windows installer now ships with JRE 11 (#2101), but the Windows and macOS builds still use JDK 8. Upgrade these, as well as instructions for Ubuntu.

bls337:
@marktsuchida @nicost Are there any issues preventing the project from upgrading the build system to JDK11?

marktsuchida:
Mostly needing to rebuild the machine images we use for building. Sorry for the wait....

bls337:
@marktsuchida No worries, thanks for the update!

PR #2156, opened 2025-06-24
ASIRingTirf: NDTiff support, add support for Photometrics Prime 95B, new PLogic program,

bls337:
Add support for Photometrics Prime 95B camera
Add NDTiff support to plugin acquisition
Add new PLogic program to control output BNC timing (requires firmware >= 3.51)
Use BNC1 or BNC8 depending on pre-init "PLogicMode" property
"Number of Circles > 1" now only emits 1 TTL pulse with the new PLogic program
Add support for the DemoCam camera
Use Galvo API instead of direct serial commands for scanner (supports axes other than H/I)
Detect axis names for the Scanner and use them in UI
Only restart Fast Circles if enabled (so fast circles isn't silently enabled)
Add dialog box if the camera is not supported
Prompt user with dialog box before opening the web browser
Make it easier to add support for new cameras with improved code structure/enums
Use double instead of float, "FloatProperty" underlying type is double


marktsuchida:
Thanks @bls337!

Just once compile error. This might be because we're still using Java 8 to build, even though we use Java 11 at run-time (#2113).

```text
 [mm-javac] Compiling 20 source files to C:\Users\mark\devel\micro-manager\build\intermediates\Classes\plugins\ASIRingTIRF
 [mm-javac] warning: No processor claimed any of these annotations: java.lang.FunctionalInterface,org.scijava.plugin.Plugin,com.google.common.eventbus.Subscribe,com.google.gson.annotations.Expose
 [mm-javac] C:\Users\mark\devel\micro-manager\plugins\ASIRingTIRF\src\main\java\com\asiimaging\tirf\model\TIRFControlModel.java:276: error: cannot infer type arguments for SwingWorker<T,V>
 [mm-javac]       final SwingWorker<Void, Void> worker = new SwingWorker<>() {
 [mm-javac]                                                             ^
 [mm-javac]   reason: cannot use '<>' with anonymous inner classes
 [mm-javac]   where T,V are type-variables:
 [mm-javac]     T extends Object declared in class SwingWorker
 [mm-javac]     V extends Object declared in class SwingWorker
 [mm-javac] C:\Users\mark\devel\micro-manager\plugins\ASIRingTIRF\src\main\java\com\asiimaging\tirf\model\TIRFControlModel.java:356: error: cannot infer type arguments for SwingWorker<T,V>
 [mm-javac]       final SwingWorker<Void, Void> worker = new SwingWorker<>() {
 [mm-javac]                                                             ^
 [mm-javac]   reason: cannot use '<>' with anonymous inner classes
 [mm-javac]   where T,V are type-variables:
 [mm-javac]     T extends Object declared in class SwingWorker
 [mm-javac]     V extends Object declared in class SwingWorker
 [mm-javac] 2 errors
 [mm-javac] 1 warning
```

bls337:
@marktsuchida I changed those two lines back to `new SwingWorker<Void, Void>()` which should work in Java 8. Thanks for letting me know, I will have to keep track of #2113.

PR #2159, opened 2025-06-27
DisplaySettings API change

nicost:
This PR updates the `DisplaySettings `so that they behave a bit more predictable and logical, so that different code/plugins can more easily maintain their own `DisplaySettings`.  Specifically, any change in DisplaySettings will cause the DisplaySettings to be written to the profile under the profileKey provided through the DisplaySettings Builder.  DisplaySettings can be restored from the profile using a static function that takes the profile and profileKey as input.

 The behavior of `ChannelDisplaySettings `has not been changed (yet), in part to avoid making this PR too big, and in part since I am not yet completely convinced what that should look like (rather experience the current behavior for a bit).

The DisplaySettings API has the following additions:
- static Keys for Window positioning 
- static Key to store `DisplaySettings` in profile
- Builder function to set `WindowPositioningKey`
- Builder function to set profile and key to sync these `DisplaySettings `with
- Access functions for WindowPositionKey, profile, and profileKey
- static function `Builder fromPropertyMap(PropertyMap)`
- `static PropertyMap toPropertyMap(DisplaySettings displaySettings)`
- `static DisplaySettings.Builder restoreFromProfile(UserProfile profile, String key)`

### DIFF
--- a/mmstudio/src/main/java/org/micromanager/display/DisplayManager.java
+++ b/mmstudio/src/main/java/org/micromanager/display/DisplayManager.java
@@ -24,7 +24,6 @@ import java.io.IOException;
 import java.util.List;
 import org.micromanager.EventPublisher;
 import org.micromanager.PropertyMap;
-import org.micromanager.UserProfile;
 import org.micromanager.data.DataProvider;
 import org.micromanager.data.Datastore;
 import org.micromanager.data.Image;
@@ -57,14 +56,14 @@ public interface DisplayManager extends EventPublisher {
    DisplaySettings getStandardDisplaySettings();
 
    /**
-    * Restore DisplaySettings.Builder from the UserProfile using the
+    * Restore DisplaySettings from the UserProfile using the
     * profile key. Returns null if the profile does not contain the requested DisplaySettings.
     *
     * @param profileKey The key in the UserProfile to use for the DisplaySettings.
-    * @return A DisplaySettings.Builder with the values from the UserProfile or null if
+    * @return DisplaySettings with the values from the UserProfile or null if
     *         the UserProfile did not contain DisplaySettings under the provided key.
     */
-   DisplaySettings.Builder displaySettingsBuilderFromProfile(String profileKey);
+   DisplaySettings displaySettingsFromProfile(String profileKey);
 
    /**
     * Generate a "blank" DisplaySettings.Builder with all null values.
--- a/plugins/Deskew/src/main/java/org/micromanager/deskew/DeskewAcqManager.java
+++ b/plugins/Deskew/src/main/java/org/micromanager/deskew/DeskewAcqManager.java
@@ -164,14 +164,17 @@ public class DeskewAcqManager {
       }
 
       final String displayKey = projectionTypeDisplayKeys.get(projectionType);
-      DisplaySettings.Builder displaySettingsBuilder =
-               studio_.displays().displaySettingsBuilderFromProfile(displayKey);
-      if (displaySettingsBuilder == null) {
-         displaySettingsBuilder = studio_.displays().displaySettingsBuilderFromProfile(
+      DisplaySettings displaySettings =
+               studio_.displays().displaySettingsFromProfile(displayKey);
+      DisplaySettings.Builder displaySettingsBuilder = null;
+      if (displaySettings == null) {
+         displaySettings = studio_.displays().displaySettingsFromProfile(
                   PropertyKey.ACQUISITION_DISPLAY_SETTINGS.key());
-         if (displaySettingsBuilder == null) {
-            displaySettingsBuilder = studio_.displays().displaySettingsBuilder();
-         }
+      }
+      if (displaySettings == null) {
+         displaySettingsBuilder = studio_.displays().displaySettingsBuilder();
+      } else {
+         displaySettingsBuilder = displaySettings.copyBuilder();
       }
       if ((settings.containsKey(DeskewFrame.SHOW) && settings.getBoolean(DeskewFrame.SHOW, false))
                || (settings.containsKey(DeskewFrame.OUTPUT_OPTION)
--- a/mmstudio/src/main/java/org/micromanager/acquisition/internal/MMAcquisition.java
+++ b/mmstudio/src/main/java/org/micromanager/acquisition/internal/MMAcquisition.java
@@ -191,8 +191,8 @@ public final class MMAcquisition extends DataViewerListener {
 
          // Use settings of last closed acquisition viewer
          DisplaySettings.Builder displaySettingsBuilder =
-                  studio_.displays().displaySettingsBuilderFromProfile(
-                           PropertyKey.ACQUISITION_DISPLAY_SETTINGS.key());
+                  studio_.displays().displaySettingsFromProfile(
+                           PropertyKey.ACQUISITION_DISPLAY_SETTINGS.key()).copyBuilder();
 
          if (displaySettingsBuilder == null) {
             displaySettingsBuilder = DefaultDisplaySettings.builder();
--- a/mmstudio/src/main/java/org/micromanager/acquisition/internal/testacquisition/TestAcqAdapter.java
+++ b/mmstudio/src/main/java/org/micromanager/acquisition/internal/testacquisition/TestAcqAdapter.java
@@ -241,10 +241,12 @@ public class TestAcqAdapter extends DataViewerListener implements
          displayWindow_ = studio_.displays().createDisplay(curStore_, null);
          displayWindow_.setCustomTitle(TITLE);
 
-         // Use settings of last closed acquisition viewer
-         DisplaySettings.Builder displaySettingsBuilder =
-                  studio_.displays().displaySettingsBuilderFromProfile(
-                           TEST_ACQUISITION_DISPLAY_SETTINGS);
+         DisplaySettings displaySettings = studio_.displays().displaySettingsFromProfile(
+                     TEST_ACQUISITION_DISPLAY_SETTINGS);
+         DisplaySettings.Builder displaySettingsBuilder = null;
+         if (displaySettings != null) {
+            displaySettingsBuilder = displaySettings.copyBuilder();
+         }
          final int nrChannels = summaryMetadata.getChannelNameList().size();
          if (displaySettingsBuilder == null) {
             displaySettingsBuilder = DefaultDisplaySettings.builder();
--- a/mmstudio/src/main/java/org/micromanager/display/internal/DefaultDisplayManager.java
+++ b/mmstudio/src/main/java/org/micromanager/display/internal/DefaultDisplayManager.java
@@ -170,7 +170,7 @@ public final class DefaultDisplayManager extends DataViewerListener implements D
    }
 
    @Override
-   public DisplaySettings.Builder displaySettingsBuilderFromProfile(String profileKey) {
+   public DisplaySettings displaySettingsFromProfile(String profileKey) {
       return DefaultDisplaySettings.restoreFromProfile(profileKey);
    }
 
--- a/mmstudio/src/main/java/org/micromanager/display/internal/DefaultDisplaySettings.java
+++ b/mmstudio/src/main/java/org/micromanager/display/internal/DefaultDisplaySettings.java
@@ -470,13 +470,13 @@ public final class DefaultDisplaySettings implements DisplaySettings {
     *                Will be pre-pended with PROFILEKEY.
     * @return Stored DisplaySettings or null if none found.
     */
-   public static DisplaySettings.Builder restoreFromProfile(String key) {
+   public static DisplaySettings restoreFromProfile(String key) {
       UserProfile profile = MMStudio.getInstance().profile();
       MutablePropertyMapView mpmv = profile.getSettings(DefaultDisplaySettings.class);
       final String finalKey = PROFILEKEY + "-" + key;
       if (mpmv.containsPropertyMap(finalKey)) {
          PropertyMap propertyMap = mpmv.getPropertyMap(finalKey, null);
-         return fromPropertyMap(propertyMap);
+         return fromPropertyMap(propertyMap).build();
       }
       return null;
    }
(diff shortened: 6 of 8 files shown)


mmstudio/src/main/java/org/micromanager/display/DisplaySettings.java
```diff
@@ -273,7 +305,153 @@ Builder copyBuilderWithComponentSettings(
...
+               .putEnumAsString(PropertyKey.COLOR_MODE.key(), displaySettings.getColorMode())
+               .putBoolean(PropertyKey.UNIFORM_CHANNEL_SCALING.key(), displaySettings.isUniformChannelScalingEnabled())
+               .putBoolean(PropertyKey.AUTOSTRETCH.key(), displaySettings.isAutostretchEnabled())
+               .putBoolean(PropertyKey.HISTOGRAM_IS_LOGARITHMIC.key(), displaySettings.isHistogramLogarithmic())
+               .putBoolean(PropertyKey.ROI_AUTOSCALE.key(), displaySettings.isROIAutoscaleEnabled())
+               .putBoolean(PropertyKey.IGNORE_ZEROS_AUTOSCALE.key(), displaySettings.isAutoscaleIgnoringZeros())
+               .putDouble(PropertyKey.AUTOSCALE_IGNORED_QUANTILE.key(), displaySettings.getAutoscaleIgnoredQuantile())
+               .putPropertyMapList(PropertyKey.CHANNEL_SETTINGS.key(), channelSettings)
+               .putString(PropertyKey.WINDOW_POSITION_KEY.key(), displaySettings.getWindowPositionKey())
+               .putString(PropertyKey.PROFILE_KEY.key(), displaySettings.getWindowPositionKey())
```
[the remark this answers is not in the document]
nicost: Applied

plugins/Deskew/src/main/java/org/micromanager/deskew/DeskewAcqManager.java
```diff
@@ -35,6 +38,7 @@ public enum ProjectionType {
       ORTHOGONAL_VIEWS,
       FULL_VOLUME
    }
+   public EnumMap<ProjectionType, String> projectionTypeDisplayKeys = new EnumMap<>(ProjectionType.class);
```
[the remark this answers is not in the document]
nicost: Applied

mmstudio/src/main/java/org/micromanager/display/internal/RememberedDisplaySettings.java
```diff
@@ -167,7 +167,16 @@ public static DisplaySettings fixMissingInfo(
...
    }
+
+   private static String generateChannelKey(String key, String channelGroup, String channelName) {
+      if (key == null || key.isEmpty()) {
+         return generateChannelKey(channelGroup, channelName);
+      }
+      return key + ":" + channelGroup + ":" + channelName;
+   }
+
+
```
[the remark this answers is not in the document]
nicost: Leaving for now, as this will be useful in the work on ChannelDisplaySettings

nicost:
Thought: maybe the static functions I added to DisplaySettings would make more sense in DisplayManager?

nicost:
Moved the static functions either to displayManager (`DisplaySettings.Builder displaySettingsBuilderFromProfile(UserProfile profile, String profileKey);`) or DefaultDisplaySettings.

mmstudio/src/main/java/org/micromanager/display/AbstractDataViewer.java
```diff
@@ -20,6 +20,7 @@
 import org.micromanager.data.Coords;
 import org.micromanager.data.DataProvider;
 import org.micromanager.data.Datastore;
+import org.micromanager.display.internal.DefaultDisplaySettings;
```
marktsuchida: API should not depend on internal class (included by accident?).

mmstudio/src/main/java/org/micromanager/display/DisplayManager.java
```diff
@@ -47,12 +48,26 @@ public interface DisplayManager extends EventPublisher {
    /**
     * Retrieve a DisplaySettings holding the values the user has saved as their
     * default values.
+    * Nico 20250627: It is unclear what this method is supposed to do. I can
+    * not find a "Set as default" button, or a DisplayWindow Settings tab.
     *
     * @return The DisplaySettings as of the last time the user clicked the
     *     "Set as default" button in the Settings tab of a DisplayWindow.
```
marktsuchida: But what does it actually return before these changes? Was it always broken? In any case, if there is no longer a meaningful "user-set default for all displays", then probably the best thing to do is to deprecate this method and make it return a plain `DisplaySettings` (i.e., the fixed default), and document as such.

mmstudio/src/main/java/org/micromanager/display/DisplayManager.java
```diff
@@ -47,12 +48,26 @@ public interface DisplayManager extends EventPublisher {
...
+    * Restore DisplaySettings.Builder from the provided UserProfile and
+    * profile key. Returns null if the profile does not contain the requested DisplaySettings.
+    *
+    * @param profile    The UserProfile to use for the DisplaySettings.
+    * @param profileKey The key in the UserProfile to use for the DisplaySettings.
+    * @return A DisplaySettings.Builder with the values from the UserProfile or null if
+    *         the UserProfile did not contain DisplaySettings under the provided key.
+    */
+   DisplaySettings.Builder displaySettingsBuilderFromProfile(UserProfile profile,
+                                                             String profileKey);
```
marktsuchida: Letting the user pass in the profile explicitly is a huge break from how the API works everywhere else. Is that needed?

marktsuchida: Likewise, returning a builder rather than a plain `DisplaySettings` goes against the general design of the API that emphasizes immutable objects. Is there any reason why we can't just have `DisplaySettings loadDisplaySettingsForKey(String)`?

mmstudio/src/main/java/org/micromanager/display/DisplaySettings.java
```diff
@@ -40,9 +51,16 @@ public interface DisplaySettings {
...
+   String ALBUM_DISPLAY = "ALBUM_DISPLAY";
+   String PREVIEW_DISPLAY = "PREVIEW_DISPLAY";
+   String MDA_DISPLAY = "MDA_DISPLAY";
+
+   /**
+    * Key used to store DisplaySettings in the UserProfile.
+    * This key is prepended to the key used to store the DisplaySettings
+    * in the UserProfile.
+    */
+   String PROFILEKEY = "Default_Display_Settings";
```
marktsuchida: This seems like an implementation detail of `DefaultDisplaySettings` and should not be in the API interface.

mmstudio/src/main/java/org/micromanager/display/DisplaySettings.java
```diff
@@ -144,6 +162,16 @@ interface Builder {
...
+      /**
+       * Sets the key under which these DisplaySettings will be stored in the profile.
+       * This key allows the DisplaySettings to keep an up-to-date copy in the profile.
+       *
+       * @param profile UserProfile to which these DisplaySettings will be saved.
+       * @param key Used to identify these DisplaySettings in the profile.
+       * @return builder instance to enable chaining commands
+       */
+      Builder profileKey(UserProfile profile, String key);
+
```
marktsuchida: Does this really need to be inside the `DisplaySettings` object? Can't we just have a `DisplayManager` method (`saveDisplaySettingsForKey(String, DisplaySettings)`) to save a given `DisplaySettings` under a given key?

Putting this inside the `DisplaySettings` means that the object needs to be cleansed of the profile key when being repurposed for another type of display; that seems error-prone. And logically the type of display is not part of the display settings.

mmstudio/src/main/java/org/micromanager/display/AbstractDataViewer.java
nicost: Yes, that import was not used.  Note, however, that AbstractDateViewer still depends on four internal classes:
```
import org.micromanager.display.internal.event.DefaultDisplayPositionChangedEvent;
import org.micromanager.display.internal.event.DefaultDisplaySettingsChangedEvent;
import org.micromanager.internal.utils.EventBusExceptionLogger;
import org.micromanager.internal.utils.ThreadFactoryFactory;
```
These predate this PR, and will be difficult to remove.  Would it make most sense to make AbstractDataViewer an internal class?

mmstudio/src/main/java/org/micromanager/display/DisplaySettings.java
nicost: The problem here is that the DisplaySettings need to auto-save themselves to the profile whenever the settings change.  We respond to a DisplaySettingsChanged event, but then how do we find out which type of display those settings belonged to.  I don't really see another way than storing the key itself in the DisplaySettings, anything else become very convoluted.  If you have ideas, let me know.

mmstudio/src/main/java/org/micromanager/display/DisplayManager.java
nicost: Passing the profile is removed in e1e926978.  It re-introduces dependence on a bunch of uses of static instance of MMStudio, which is something I do not like very much.

nicost: As for returning Builder rather than the object itself, I found myself invariably getting a builder out of the object since there is always something that needs to be changed.  It saves quite a few CPU cycles to return a builder rather than a build builder that needs to get copied back into a builder then build again.  The UI felt noticeably smoother with this approach.  If you really want objects rather than builder returned, I can do that but it introduces a lot of (imho) unnecessary overhead.

mmstudio/src/main/java/org/micromanager/display/DisplaySettings.java
nicost: Done in a335c7a00

mmstudio/src/main/java/org/micromanager/display/AbstractDataViewer.java
marktsuchida: No, that was my mistake -- this is a base implementation class, so it's not wrong to depend on internal classes as long as they are not exposed.

mmstudio/src/main/java/org/micromanager/display/DisplaySettings.java
marktsuchida: I think it is worth figuring out another way, even if it takes a bit of work, because embedding the key also breaks existing code (previously it was correct to take the settings from one viewer and apply to another; not it will cause settings to be randomly saved to the wrong key or not saved at all; that breaks the whole idea of saving settings per viewer category).

To me it looks like the DisplayWindow (or its subcomponent DisplayUIController) just needs to store which key to use (default = null = don't save), for which an API can be added. Then the saving code in `applyDisplaySettings` can just save to that key -- unless I'm missing something.

I probably should have thought of this also for the window position key -- that probably should be a direct parameter of the DisplayWindow, not part of the DisplaySettings. If we can back out of that (not in this PR) I would be for it. But in that case the problem is smaller because the worst thing that can happen is that the window opens in a different position when DisplaySettings are transferred without amending.

mmstudio/src/main/java/org/micromanager/display/DisplayManager.java
marktsuchida: Re the use of static MMStudio instance -- I agree, but that needs to be fixed globally (though not sure how) rather than introducing local inconsistencies. Currently, API users are not expected to care about the current profile, and they shouldn't need to do so just to save display settings.

Re the return type: I think the overhead does serve a good purpose, but I do care about how much is "a lot".

Suppose we, in the future, add a way to cache the "saved" settings (per key, of course). Then, if we can return a `DisplaySettings` we simply return the stored object without copying. That will more than likely be faster than loading from the profile, even if most callers subsequently need to create a builder from it (though that's an assumption that would need to be tested). But if our API specifies that we must return a bulider, we would need to create a new builder on every call, even if the caller would otherwise not need to modify it.

So in general I don't think it's a good idea to distort the logical structure of the API based on incidental effects of the current implementation. But that is not to say that API design should completely ignore performance issues. How bad is it? Isn't this only called when a new display is opened (= not in a tight loop) anyway?

mmstudio/src/main/java/org/micromanager/display/DisplaySettings.java
nicost: Good call.  Turns out this was not too hard.  DisplaySetting saving is now managed by DisplayWindow, which makes more sense.  Changes are in eab05ff01. It should be straight forward to do the same with the WIndowPositionKey.  Since we introduced that only very recently, I think it will be safe to make that change now.  Do it right after this PR is merged?

mmstudio/src/main/java/org/micromanager/display/DisplayManager.java
nicost: Yes, this only happens when the DisplayWindow is opened.  To do a builder > build cycle, at least 10 variables will be copied twice, going through the various builder functions.  If the UI code did not need to immediately create another builder, I would be totally fine with it.  Currently, most of the code is a variation on the following (this is from TestAcqAdapter):


```
         DisplaySettings.Builder displaySettingsBuilder =
                  studio_.displays().displaySettingsBuilderFromProfile(
                           TEST_ACQUISITION_DISPLAY_SETTINGS);
         final int nrChannels = summaryMetadata.getChannelNameList().size();
         if (displaySettingsBuilder == null) {
            displaySettingsBuilder = DefaultDisplaySettings.builder();
            if (nrChannels == 1) {
               displaySettingsBuilder.colorModeGrayscale();
            } else {
               displaySettingsBuilder.colorModeComposite();
            }
         }
         if (nrChannels > 0) {
            // the do-while loop is a way to set display settings in a thread
            // safe way.  See docs to compareAndSetDisplaySettings.
            for (int channelIndex = 0; channelIndex < nrChannels
                     && channelIndex < acquisitionSettings.channels().size(); channelIndex++) {
               displaySettingsBuilder.channel(channelIndex,
                        RememberedDisplaySettings.loadChannel(studio_,
                                 summaryMetadata.getChannelGroup(),
                                 summaryMetadata.getChannelNameList().get(channelIndex),
                                 acquisitionSettings.channels().get(channelIndex).color()));
            }
         }

         displayWindow_.setDisplaySettings(displaySettingsBuilder.build());
```

marktsuchida:
@nicost Thanks for the changes, looks good to merge!

> It should be straight forward to do the same with the WIndowPositionKey. Since we introduced that only very recently, I think it will be safe to make that change now. Do it right after this PR is merged?

Yes, I think that's fine since it's only been days.

PR #2172, opened 2025-07-18
mm1.4 handle XYStage offset and mirror operations in sequences and scanning

jondaniels:
We have discussed this on the forum at https://forum.image.sc/t/device-adapter-for-xy-stage-handling-mirroring-and-offsets.

We can treat this commit to the mm1.4 branch (used only by ASI I think) as a trial run.

The API in MMDevice.h creates an optional offset and mirroring between hardware units ("steps") and units displayed by MM.  Notably these are used in the HCS plugin.  As long as you only are using positions via the following 3 API calls everything is fine.

```
int SetPositionUm(double x, double y);
int SetRelativePositionUm(double dx, double dy);
int GetPositionUm(double& x, double& y);
```

However the ASITiger device adapter uses positions it at least 2 other places, and these didn't previously take into account the disconnect between hardware units and the displayed units. 
- sequence positions
- scan positions (set with properties in units of millimeters)

So we need a method to do the same calculations as the the get/set position methods do so that other places in the device adapter can send correct information to the controller.

My solution is to add a new method to MMDevice.h under XY stage that is code extracted from `SetPositionUm(...)` and exposed separately (and now `SetPositionUm(...)' calls this method internally).

`int CalculatePositionSteps(long& xSteps, long& ySteps, double x, double y);`

I use this method in my device adapter so that sequence positions and scan positions calculate the correct hardware positions before sending anything to the controller.

### DIFF
--- a/DeviceAdapters/ASITiger/ASIXYStage.cpp
+++ b/DeviceAdapters/ASITiger/ASIXYStage.cpp
@@ -723,7 +723,7 @@ int CXYStage::SendXYStageSequence()
    for (unsigned i=0; i< sequenceX_.size(); i++)  // send new points
    {
       command.str("");
-      command << "LD " << axisLetterX_ << "=" << sequenceX_[i]*unitMultX_ << " " << axisLetterY_ << "=" << sequenceY_[i]*unitMultY_;
+      command << "LD " << axisLetterX_ << "=" << sequenceX_[i] << " " << axisLetterY_ << "=" << sequenceY_[i];
       RETURN_ON_MM_ERROR ( hub_->QueryCommandVerify(command.str(),":A") );
    }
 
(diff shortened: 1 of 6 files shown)


jondaniels:
OK so it looks like I never finished my previous PR with the ASIdiSPIM plugin for Hamamastu light sheet mode.  I don't immediately see how to disentangle the two sets of commits but they are all mine and probably don't affect anyone else...

First two commits relate to diSPIM plugin and only affect light sheet mode on Hamamatsu cameras.  It is to work around what seems to be a bug in HamamatsuHam.

Last commit is in the ASITiger device adapter and touches MMDevice.h.

marktsuchida:
Pushed one last fix -- remembering that MM1.4 was pre-C++11.

jondaniels:
Sorry for the lag, yes you can go ahead and commit.

marktsuchida:
Thanks, @jondaniels!

PR #2185, opened 2025-08-13
Studio and plugins: Attempt to code coming from events on the EDT whenever appropriate.

nicost:


nicost:
It turns out that the callbacks are already running on the EDT, so there is no need for this.

PR #2263, opened 2025-12-26
Scripts: Added script that stops all running sequences.

nicost:


nicost:
Also added a script that automates power measurements with the Thorlabs power meter.  I am not totally sure if these scripts should go in the repo or somewhere else (if so, where?), but I need them in our code, and they may very well be useful to others.  @marktsuchida please advise about best course of action.

PR #2264, opened 2025-12-31
Fix automated calibration binning with Tucsen cameras

rhysDevans97:
Previously, as some Tucsen cameras use different "Binning" & "Binning Sum" parameters, AutomaticCalibrationThread scaled by the number of pixels instead of the binning factor, leading to incorrect results in image acquisition.

This change ensures that 'Binning Sum' is used when values are erroneous, so pixels scale correctly.
No other device adapters are affected.


nicost:
I do not think this is the right approach.  The property "Binning" is supposed to return numbers like 1, 2, or optionally 1x1, 2x2, etc...  The Tucsen device adapter - for reasons that are unclear to me - is not following that approach, but returns the image with and height in pixels instead.  Note that the Micro-Manager API has other functions to get exactly those numbers.  Apparently, the Tucsen device adapter has another - non-standard - property "Binning-Sum" that returns what the "Binning" property should return.  

It really does not make sense to try to work around all kinds of idiosyncrasies and  mis-understandings originating in device adapter in the user interface code.  Rather, please fix this this in the device adapter itself.  It is simple enough in this case, simply use the rename the properties that are currently "Binning" and "Binning-Sum" (swapping them would be a start, although you may want to rename Binning-Sum afterwards).

PR #2277, opened 2026-01-22
Create Docker build environment Linux/WSL

eirinnm:
This PR introduces a Dockerfile which can be used to build MM into an image which can be run on Linux or WSL2.

Features:

* Multi-stage build to reduce filesize
* Included docker-compose script sets up device mounting.
* Included run script sets up Xauthority pass-through to enable the GUI to operate from the container, on Linux or WSL2 (tested on Ubuntu on both hosts). 
* Supports optional third-party drivers such as PVCam.

### Building
Option 1:
```
cd docker
docker compose build
```
Option 2:
```
docker build -t micro-manager:latest -f docker/Dockerfile .
```
The result is the same. Note: building takes about 20 minutes.

### Running
`./docker/run-linux.sh`

### Third-party drivers
If you need extra drivers installed, copy them to the docker/drivers/ path, and edit `setup-drivers.sh` to unpack them or run the install script. The included `setup-drivers.sh` is provided to unpack PVCAM-Linux-3-10-0-3.zip if present.

For more information, see docker/README.md.

### DIFF
--- a/docker/docker-compose.yaml
+++ b/docker/docker-compose.yaml
@@ -10,7 +10,6 @@ services:
     environment:
       - DISPLAY=${DISPLAY:-:0}
       - XAUTHORITY=/tmp/.Xauthority/Xauthority
-      - TZ=${TZ:-Europe/London}
     volumes:
       - ${X11_UNIX_SOCKET:-/tmp/.X11-unix}:/tmp/.X11-unix
       - ${XAUTH_PATH:-../.Xauthority}:/tmp/.Xauthority:ro


docker/Dockerfile
```diff
@@ -0,0 +1,114 @@
+################################################################################
+# STAGE: Builder
+################################################################################
+FROM ubuntu:22.04 AS builder
+
+ENV TZ=Europe/London
```
marktsuchida: Would be nice to avoid hard-coding the timezone. LLM tells me you can put the following in docker-compose.yml:

```yaml
services:
  your-service:
    image: your-image
    volumes:
      - /etc/localtime:/etc/localtime:ro
      - /etc/timezone:/etc/timezone:ro
```

Or at least `TZ` could be set there so as not to bake into the image.

docker/Dockerfile
```diff
@@ -0,0 +1,114 @@
...
+RUN wget https://wsr.imagej.net/distros/linux/ij154-linux64-java8.zip \
+-O /tmp/ij154-linux64-java8.zip && unzip /tmp/ij154-linux64-java8.zip \
+    -d /root && rm -rf $IMAGEJ_DIR/jre && \
+    chmod a+Xr $IMAGEJ_DIR
+
+### Setup Micromanager from local source ###  
+ARG MM_SRC_DIR=/root/mm-src
+RUN mkdir -p $MM_SRC_DIR && \
+    mkdir -p $MM_SRC_DIR/3rdpartypublic/classext && \
+    wget --no-check-certificate https://svn.micro-manager.org/3rdpartypublic/classext/iconloader.jar -P $MM_SRC_DIR/3rdpartypublic/classext/
```
marktsuchida: Did you need `--no-check-certificate`?
`svn.micro-manager.org` currently uses LetsEncrypt certs and should pass the normal checks.

(Let me know if you saw it in the docs somewhere; we should remove it.)

docker/Dockerfile
```diff
@@ -0,0 +1,114 @@
...
+
+# Copy helper scripts
+COPY ./docker/micromanager.sh /root/ImageJ/micromanager.sh
+
+################################################################################
+# STAGE: Final Image
+################################################################################
+FROM ubuntu:22.04
+
+ENV TZ=Europe/London
```
marktsuchida: Ditto with above.

docker/README.md
```diff
@@ -0,0 +1,57 @@
+# Micro-Manager Docker Environment
+
+This Docker setup provides a containerized environment to build and run Micro-Manager.
+
+## Getting Started
+
+The recommended way to run Micro-Manager is using Docker Compose, which handles the necessary configuration for hardware access (USB/Serial) and GUI display.
```
marktsuchida: "The recommended way to run Micro-Manager" -> "The recommended way to run Micro-Manager in Docker"

docker/README.md
```diff
@@ -0,0 +1,57 @@
...
+
+### Prerequisites
+
+- Docker and Docker Compose installed.
+- (Linux) X11 server running (local display).
```
marktsuchida: Does this mean Ubuntu users need to switch from Wayland (the default) to X11? (I don't know too much in this area.)

docker/setup-drivers.sh
```diff
@@ -0,0 +1,53 @@
+#!/bin/bash
+set -e
+
+# This script is called by the Dockerfile during the builder stage.
+# It optionally installs the PVCAM SDK if the files are present.
+
+ZIP_FILE="./SecretDeviceAdapters/PVCAM-Linux-3-10-0-3.zip"
```
marktsuchida: `SecretDeviceAdapters` (actually `mmCoreAndDevices/SecretDeviceAdapters`) is where we place an internal git repo containing closed-source device _adapters_, not _drivers_ from vendors. So it would help us maintainers if another directory could be used for this. Perhaps `docker/drivers` (and add a `docker/drivers/.gitignore`).

docker/README.md
eirinnm: No, I think it runs fine on both.

.gitmodules
```diff
@@ -1,3 +1,4 @@
 [submodule "mmCoreAndDevices"]
 	path = mmCoreAndDevices
 	url = https://github.com/micro-manager/mmCoreAndDevices.git
+	branch = main
```
marktsuchida: Missed this earlier. Was this addition intentional/necessary?

I don't think it really makes a difference as long as mmCoreAndDevices' default branch remains `main` (and only affects `git submodule update --remote`), but it might be good to leave at the default (no `branch = `), since our intent is that the pinned commit of mmCoreAndDevices should be used, not necessarily the latest (the pinned commit is currently bumped whenever mmCoreAndDevices changes, so things should generally be in sync).

eirinnm: Good catch. TBH this is my first time working with git submodules. I've removed this line now.

docker/docker-compose.yaml
```diff
@@ -0,0 +1,23 @@
...
+    build:
+      context: ..
+      dockerfile: docker/Dockerfile
+    container_name: micromanager-container
+    network_mode: host # Required for many hardware device adapters (IP cameras, etc)
+    privileged: true   # Required for USB/Serial device access
+    environment:
+      - DISPLAY=${DISPLAY:-:0}
+      - XAUTHORITY=/tmp/.Xauthority/Xauthority
+      - TZ=${TZ:-Europe/London}
```
marktsuchida: By deleting this line I was able to get local time (US Central for me) inside the container, which I think is usually what is desired. Probably good to have it be either local time or UTC, not London time (which has summer time).

eirinnm: Great! I've removed that line so the timezone should be taken from the host, even on WSL2.

Also added the profiles folder to gitignore. This folder is created on the host by the docker compose script, but we don't want it checked into git.

PR #2292, opened 2026-02-14
restore stage scanning properties in ASIdiSPIM (MM1.4 only)

jondaniels:
settings for stage scan were accidentally short-circuited when ensuring that angle is always present


jondaniels:
for the ASIdiSPIM plugin in MM1.4 (no longer being built automatically), no impact on anything else

PR #2293, opened 2026-02-17
Deskew Explorer and NDV2: A variant of NDViewer that attaches to the Image Inspector

nicost:
Adds Explore mode to the Deskew plugin, and adds NDV2, a modified version of NDViewer but that attaches to the standard Image INspector Window..  It all is a bit of a hack. I'll see how I can clean this up and if it is worth it.  It looks quite nice though.


marktsuchida:
If the only reason is shared use by multiple plugins, then it can be moved (later, doesn't have to be in this PR) to a library (i.e., separate JAR that multiple plugins can depend on). But if it cannot be implemented without reaching into the internals of MMStudio, then I wonder if there is any better way.

Maybe we need some new APIs that would allow integration with the histogram computation without reaching for internal classes?

Anyway, I'd be okay with merging this first and later refactoring into a separate library (with supporting APIs in MMStudio). But putting everything permanently into MMStudio just because it's convenient to use MMStudio's internals feels a bit like giving up on modularity before even trying.

PR #2355, opened 2026-05-08
Fix bug introduced in February that caused the Z stage to first move to the start, then the desired position for all Z slices.

### LINKED ISSUE #2352: Z stage issue running MDA with AcqEngJ
Hi all,

This week we noticed an issue when running an MDA with a z stack in relative mode on a recent nightly build. It appears AcqEngJ is telling the z stage to go to the starting position and then telling it to go the next Z position in the acquisition, creating a "bouncing" effect.

Summarizing the behavior with an example:

Starting point: 100 microns
Relative stack of 5 microns at 1 micron will yield:

1. Core call to move to 98  microns
2. Capture images
3. Core call to move to 98 microns
4. Core call to move to 99 microns
5. Capture images
6. Core call to move to 98 microns
7. Core call to move to 100 microns
8. etc...

This behavior does not happen with the Clojure engine.

Problem report attached showing the behavior on our actual hardware.

[20260429_Problem Report.txt](https://github.com/user-attachments/files/27248424/20260429_Problem.Report.txt)

nicost:
Closes #2352.

Bug was introduced in #2301.  If there are side effect to this change, please open a new Issue.

### DIFF
--- a/mmstudio/src/main/java/org/micromanager/acquisition/internal/acqengjcompat/MDAAcqEventModules.java
+++ b/mmstudio/src/main/java/org/micromanager/acquisition/internal/acqengjcompat/MDAAcqEventModules.java
@@ -94,9 +94,6 @@ public class MDAAcqEventModules {
                      throw new RuntimeException(e);
                   }
                }
-               if (positionList == null) {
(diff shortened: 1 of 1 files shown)


nicost:
@aandreev0 could you check that everything still works as you intended?

PR #2371, opened 2026-05-20
added checkbox to change timing calculations to static sheet instead …

jondaniels:
…of hardcoding to only when the ASIdiSPIM.SCOPE flag is true.  allows static sheet generators to be used in e.g. diSPIM


jondaniels:
To be clear, this is about the ASIdiSPIM plugin on the 1.4 branch for which there is no automated build.  It should not affect the main branch with 2.0.

PR #2384, opened 2026-06-02
fix erroneous missing break causing deskew to be incorrect for SCOPE …

jondaniels:
…galvo scan


jondaniels:
only affects ASIdiSPIM plugin on mm1.4 branch, and only one special use case of that plugin

PR #2385, opened 2026-06-02
ClearVolume: Bumped up dependencies, mainly triggered by JOGL moving to 2.5.0.

nicost:
This makes a couple of upgrades in cleargl, clearvolume and coremem needed, which are not yet in the upstream repos.  Nevertheless, this code has been tested (on windows only), so when those dependencies have been updated this PR can be merged. Only one minor change was needed in our own code.

### DIFF
--- a/buildscripts/ivy.xml
+++ b/buildscripts/ivy.xml
@@ -96,33 +96,33 @@
 
 
       <!-- ClearVolume dependencies -->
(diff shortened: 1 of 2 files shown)


nicost:
Note, the expected ClearVolume and ClearGL jars work with JOGL 2.6.0, so that will need to be adjusted here once the jars are available.

PR #2399, opened 2026-06-20
ASIdiSPIM: fixes for static sheet setting recently added (mm1.4)

jondaniels:
fix issue with SetupPanel line 1078
added isStaticSheet field to AcquisitionSettings object
narrowed short-circuiting of static sheet settings to original intention


jondaniels:
ASIdiSPIM plugin edits on mm1.4 branch only

PR #2410, opened 2026-07-10
EMU refresh on property change

zacsimile:
This enables the EMU GUI to listen for property changes and update the GUI when they occur. It also addresses the GUI refresh issue described at https://forum.image.sc/t/a-way-to-refresh-gui-of-emu-simple-ui-by-a-script/119684.


nicost:
Looks great.  Thanks!

PR #2413, opened 2026-07-14
ASIdiSPIM: Replace hacky way to have plugin with different prefs location

jondaniels:
Now if the firmware timestamp on TigerCommHub is midnight then we use alternate pref location.


jondaniels:
affects mm1.4 branch only, nothing on main

PR #2416, opened 2026-07-17
ASIdiSPIM added lag compensation setting

jondaniels:
Add offset so user can calibrate out piezo inertial effects during acquisitions.  Expect that value will change with slice period.


jondaniels:
should only impact mm1.4 branch which is no longer built automatically

PR #2417, opened 2026-07-19
Allows building and running Micro-Mamager with JDK25.  Independently better display on Linux

nicost:
Changes the LAF on LInux.  Way better looking than before.

### DIFF
--- a/bindist/MacOSX/ImageJ.app/Contents/Info.plist
+++ b/bindist/MacOSX/ImageJ.app/Contents/Info.plist
@@ -61,8 +61,6 @@
   <string>-Xms256m -Xmx3000m</string>
   <key>JVMOptions</key>
   <array>
-    <string>--add-opens=java.desktop/java.awt=ALL-UNNAMED</string>
-    <string>--add-opens=java.desktop/java.awt.color=ALL-UNNAMED</string>
     <string>--add-opens=java.desktop/sun.awt=ALL-UNNAMED</string>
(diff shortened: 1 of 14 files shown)


nicost:
@marktsuchida  is it OK to merge this?  I believe it is safe while still using JVM11, but enables using JDK25 (and everything in between).  Sorry for missing in the Linux UI improvements (which are substantial), I thought these issues were related, but they were not.

marktsuchida:
Hard to review and test in this form. Can you split into 2 or more PRs, each on a single topic? It's easy to do this with agents.

Also, are we sure that no more `--add-opens` flags will be needed, including for plugins? (This could wait since we aren't actually shipping with JDK 17+ yet, but I think it's actually better to try to be as complete as possible at once rather than making false promises.) What strategy did you use to find all/most of the packages that need to be opened (this really ought to be in the commit message)?

Are the Gson changes strictly behavior-preserving (how do we know)?

And even if everything is perfect, I'd like to wait on merging the `--add-opens` flags -- as already discussed -- because it breaks my dev workflow in some edge cases (it's hard to build with Java 8 and launch with Java 11 using the GNU build). Only until we are able to drop the Java 8 build.

nicost:
Split out into 3 PRs: one for the Linux LAF (#2418), one to exclude a dependency on miglayout 3.4.2 (#2419, brought in the NDTiff) and one for the JDK25 upgrade (#2420).  I will copy your comments to #2420 and address them there.

PR #2420, opened 2026-07-21
Enable using JDK25

nicost:
Claude assisted analysis of issues with transitioning to JDK25.  Main change is that ChannelSpec and SequenceSettings now serialize Color via a dedicated ColorGsonAdapter instead of relying on Gson reflecting into Color's private fields (which the module system blocks on JDK 17+).  Tests for interoperability of old and new style serialization are included and show no issues (but we should be aware of possible caveats related to float numbers).  

Add documentation of possible issues in plugins, Clojure and Beanshell (nothing appears to be a problem).  This is ready to go one we are ready to ship another JVM with Micro-Manager.


nicost:
> Also, are we sure that no more --add-opens flags will be needed, including for plugins? (This could wait since we aren't actually shipping with JDK 17+ yet, but I think it's actually better to try to be as complete as possible at once rather than making false promises.) What strategy did you use to find all/most of the packages that need to be opened (this really ought to be in the commit message)?

I had Claude go through everything and analyze whether issues would arise on jdk25.  Summary of analysis is in the doc directory.  It looks like we are godo to go, including with all plugins.

> Are the Gson changes strictly behavior-preserving (how do we know)?

Good question!  I added tests to compare old and new style color serialization.  They are completely interoperable with the exception of using floats (which does not happen in our code base).  Test is added.

> And even if everything is perfect, I'd like to wait on merging the --add-opens flags -- as already discussed -- because it breaks my dev workflow in some edge cases (it's hard to build with Java 8 and launch with Java 11 using the GNU build). Only until we are able to drop the Java 8 build.

Sure.  I am very curious whether we will get performance improvements as promised for JVM25.  The good thing is that the transition seems much less painful than I anticipated.

marktsuchida:
> I had Claude go through everything and analyze whether issues would arise on jdk25. Summary of analysis is in the doc directory. It looks like we are godo to go, including with all plugins.

Great, this is exactly what I wanted to see (and some!).

I'll try to get the build onto JDK 11 soon so we can merge this!

PR #2452, opened 2026-09-12
Studio: Preserve queued TIFF and profile writes during shutdown

### LINKED ISSUE #2262: Multipage writer uses daemon thread to write files
The WriteExecutor in the MultipageTiffWriter is created as follows (StorageMultipageTiff.java line 427):

  ```
 writingExecutor_ = new ThreadPoolExecutor(1, 1, 0,
               TimeUnit.NANOSECONDS,
               new LinkedBlockingQueue<>(),
               ThreadFactoryFactory.createThreadFactory("StorageMultiPageTiff"));
```

Threadfactory creates daemon threads (which will not stop the JVM from exiting and will simply be killed if they are still running when the JVM quits).  That does not seem right fro a writing task.  

One simple solution would be to make a second createThreadFactory static function that allows for creation of a non static thread and use that in file writing code. @marktsuchida ?

seanwessmith:
Fixes #2262.

Multipage TIFF and profile saves use daemon threads, so JVM termination can abandon queued writes. This follows the requested explicit daemon/non-daemon factory split, uses non-daemon workers for saving, and keeps the existing factory as a deprecated daemon alias for plugin compatibility.

TIFF close now waits for queued writes before releasing file channels, including metadata-error and interrupted-wait paths. Profile shutdown serializes an in-flight autosave with the final save and drains the executor before clearing the profile.

**Review guide:** 11 production files only select the explicit daemon factory for existing background work. The six behavior-bearing files are `ThreadFactoryFactory`, `StorageMultipageTiff`, `ProfileSaver`, `UserProfileAdmin`, `MMStudio`, and the new `FileSavingExecutor`. Five test files account for 650 added lines.

**Shutdown policy:** non-daemon threads alone do not survive `System.exit`. The new fallback drains already-submitted file writes without Swing, using one shared 60-second deadline. This does not finalize an unclosed dataset or bound normal synchronous disk flushing. If the deadline expires, it reports that saved data may be incomplete; forced process termination cannot be protected. Idle TIFF workers time out so an unused pool does not keep the JVM alive.

**Validation:** all 15 focused tests pass on Java 11, including queued TIFF frames closed/reopened with exact pixel comparison, real profile JSON persistence, explicit exit from the EDT, and executor-wrapper garbage collection. All MMStudio Java sources and 34 test source files compile. The broader run has 189 passes and four failures reproduced with original code. No interactive GUI or microscope test was performed.

<details>
<summary>Build environment, regression evidence, and known baseline failures</summary>

Validation on macOS with Temurin 11 (`--release 8`), using the official 20260828 nightly dependency jars and excluding its prebuilt MMStudio jar:

- Compiled every MMStudio production source and all 34 test source files.
- All 15 focused shutdown/storage/profile tests pass. Coverage includes real queued TIFF frames followed by close/reopen and exact-pixel comparison, real profile-event-to-JSON persistence, metadata failure, interrupted drains, ordinary JVM termination, explicit exit from the EDT, and executor-wrapper lifetime.
- Six TIFF/profile regression cases fail against pre-change `8b607ac` and pass with the patch. The baseline ProfileSaver constructor was made package-visible only to allow the test harness to instantiate it; production behavior was unchanged.
- Broader headless run: 193 tests, 189 pass, four failures. The same four failures reproduce with original, byte-identical relevant production classes: SciFIO 8/16-bit roundtrips, `DataCoordsAnimationStateTest.testGetSetAdvance`, and `ComponentStatsTest.testSize2`.
- Reverting only the shutdown registry to weak references makes the new wrapper-GC regression fail; all four registry tests pass with strong registration.
- Changed-file Checkstyle and `git diff --check` pass.

No microscope hardware or interactive GUI session was exercised. The full native/device-adapter build and external-fixture/manual tests were not run.

</details>

AI disclosure: implementation, tests, and review were developed with OpenAI Codex assistance. The reported checks were executed locally.


nicost:
Closes issue #2262 (or so I assume)

